PR #262, opened 2015-04-03
add support for loading multiple traces in archive mode

bdoin:
Now the load button in archive mode does not erase the previous
trace. The new trace is drawn with another color to help differentiate
several traces. The clear button erases all the traces.

### DIFF
--- a/src/org/traccar/web/client/Application.java
+++ b/src/org/traccar/web/client/Application.java
@@ -63,7 +63,8 @@ public class Application {
         deviceController.getDeviceStore().addStoreHandlers(deviceStoreHandler);
         stateController = new StateController();
         mapController = new MapController(mapHandler);
-        archiveController = new ArchiveController(archiveHanlder, deviceController.getDeviceStore());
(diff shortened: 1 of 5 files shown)


tananaev:
Thanks for the pull request. It looks really good, but the thing is that I'm working on new web interface and I don't want to add any more features to the old one because otherwise I would have to re-implement them again in the new code.

You can send pull request to https://github.com/vitalidze/traccar-web.

bdoin:
I saw that a new version was under work in a post but without any pointers to it from the main traccar.org web site I did not got my hands on it. This is unfortunate as I would have worked on it instead. Can't you add a mention on traccar.org explaining what's going on, that this version is frozen and a pointer to the new version. That would have helped me.

tananaev:
You can find some initial work on the new version here:
https://github.com/tananaev/traccar/tree/master/web

I will add a note to README file to explain the state of this project.

The work on the new web interface is progressing very slowly, so this version would probably be around for quite some time.

tananaev:
I have added a note to the README file:
https://github.com/tananaev/traccar-web/blob/master/README.md

PR #287, opened 2016-09-23
Attributes aliases

Abyss777:
- Used two stores for state and for editor. Just not to mix and not overload state store for admins.
- Editor with device combobox
- Extended `attribute` model with `attribute` and `alias` fields for few reasons:
  - determine attribute/ordinary field
  - edit/add more easily (it is hard to lookup capitalized `name`)
  - use simple renderer to choose what show name or alias
- store `position` in view object to implement update grid
- update state grid after closing editor

image

Other:
- `jshint`/`jscs` fixes in `BaseMap.js`
- added `.gitignore` for Eclipse files

Again sorry for big pull request, it was hard to split on few working features


web/app/Application.js
```diff
@@ -60,6 +61,8 @@ Ext.define('Traccar.Application', {
         'Notifications',
         'AllNotifications',
         'GeofenceTypes',
+        'AttributesAliases',
```
tananaev: It should be `AttributeAliases`, no plural on the first word. Same for `All...`.

web/app/controller/Root.js
```diff
@@ -84,6 +84,7 @@ Ext.define('Traccar.controller.Root', {
                 this.asyncUpdate(true);
             }
         });
+        Ext.getStore('AllAttributesAliases').load();
```
tananaev: I think it's better to put this before loading devices.

web/app/model/Attribute.js
```diff
@@ -27,5 +27,11 @@ Ext.define('Traccar.model.Attribute', {
     }, {
         name: 'value',
         type: 'string'
+    }, {
```
tananaev: Do we need to change Attribute model?

web/app/store/AllAttributesAliases.js
```diff
@@ -0,0 +1,30 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+Ext.define('Traccar.store.AllAttributesAliases', {
+    extend: 'Ext.data.Store',
+    model: 'Traccar.model.AttributeAlias',
+
+    proxy: {
+        type: 'rest',
+        url: 'api/devices/aliases',
```
tananaev: I just thought maybe we should move aliases to its own top level path?

web/app/view/AttributesAliases.js
```diff
@@ -0,0 +1,62 @@
...
+            typeAhead: true,
+            listeners: {
+                change: 'onDeviceChange'
+            }
+        }]
+    },
+
+    listeners: {
+        selectionchange: 'onSelectionChange',
+        destroy: 'onClose'
```
tananaev: Do we need this? We can just subscribe to store updates directly.

web/app/view/BaseMap.js
```diff
@@ -63,7 +63,7 @@ Ext.define('Traccar.view.BaseMap', {
                     imagerySet: 'Aerial'
```
tananaev: Thanks for fixing code style issues here.

web/app/model/Attribute.js
Abyss777: The reasons is:
- Determine what row is selected: position field ( speed, altitude etc.) or attribute. `aliasEditButton` is enabled only for attributes. But I just thought that `priority` can be used for such check. OK.
- I consider that we should make it comfort and load selected row to `AttributeAliasDialog` on edit click. To load we need:
  - Know, if it is already alias, and load existed record from store to form. But we have to know original lower-camel-case attribute to lookup in store.
  - If it is not alias we have to fill `attribute` form field by original lower-camel-case attribute.

I can move alias applying from renderer co controller, then `alias` field can be removed from model.

Am I missing something?

web/app/store/AllAttributesAliases.js
Abyss777: I thought about it too. Should we mark it somehow that it is aliases for attributes? I mean can we have aliases for something else in feature?

web/app/view/AttributesAliases.js
Abyss777: You are right.

web/app/model/Attribute.js
tananaev: I guess we can add some fields, but why do we need 3 of them (name, attribute and alias). Don't we need just two (original attribute and alias)?

web/app/store/AllAttributesAliases.js
tananaev: Maybe "api/attributes/aliases"?

Abyss777:
Seems all done. I decided to leave `name` and `attribute` to not break attributes editor.

web/app/view/AttributesAliasesController.js
```diff
@@ -0,0 +1,94 @@
...
+        'Traccar.model.AttributeAlias'
+    ],
+
+    init: function () {
+        var admin = Traccar.app.getUser().get('admin');
+        this.lookupReference('deviceField').setStore(admin ? 'AllDevices' : 'Devices');
+        this.lookupReference('toolbarAddButton').setDisabled(true);
+        this.lookupReference('toolbarEditButton').setDisabled(true);
+        this.lookupReference('toolbarRemoveButton').setDisabled(true);
+        this.getView().getStore().loadData([], false);
```
tananaev: Is this a way to clear the store? There must be a better more clear way (maybe `removeAll`?).

web/app/view/AttributesAliasesController.js
```diff
@@ -0,0 +1,94 @@
...
+        this.lookupReference('toolbarRemoveButton').setDisabled(disabled);
+    },
+
+    onDeviceChange: function (combobox, newValue, oldValue) {
+        this.onSelectionChange('');
+        if (newValue !== null) {
+            this.getView().getStore().getProxy().setExtraParam('deviceId', newValue);
+            this.getView().getStore().load();
+        } else {
+            this.getView().getStore().loadData([], false);
```
tananaev: Same as above. Also, even if you use this method, you don't have to provide second parameter. It defaults to "false".

web/app/view/StateController.js
```diff
@@ -39,6 +43,14 @@ Ext.define('Traccar.view.StateController', {
                 },
                 '#Positions': {
                     clear: 'clearReport'
+                },
+                '#AllAttributeAliases': {
```
tananaev: At this point I thought... why do we need two stores? Maybe we can just use one which would have all aliases. And then we can filter (use chain store for it) it to display aliases per device. What do you think?

Abyss777:
You are right. Thank you for make me think more and check.
I've removed second store.
The logic became:
 - On application load aliases are loaded only for linked devices.
 - If user edit aliases, store are not additionally updated
 - If admin edit aliases, additional (not linked to him devices) are loaded dynamically only if he select such device.

Unfortunately there is some lag for devices that were not loaded before and spinner does not appear while loading.

PS: `loadData([], false);` was used to clear local store and not clear remote store. But it does not matter now.

web/l10n/en.json
```diff
@@ -29,6 +29,9 @@
     "sharedHourAbbreviation": "h",
     "sharedMinuteAbbreviation": "m",
     "sharedGetMapState": "Get Map State",
+    "sharedAttributeAlias": "Attribute Alias",
+    "sharedAttributesAliases": "Attributes Aliases",
```
tananaev: First word should be singular here as well.

web/app/view/SettingsMenu.js
```diff
@@ -58,6 +58,11 @@ Ext.define('Traccar.view.SettingsMenu', {
             handler: 'onNotificationsClick',
             reference: 'settingsNotificationsButton'
         }, {
+            hidden: true,
+            text: Strings.sharedAttributesAliases,
+            handler: 'onAttributesAliasesClick',
```
tananaev: Change attributes to singular. Double check other places as well.

web/app/store/AllAttributeAliases.js
```diff
@@ -0,0 +1,30 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+Ext.define('Traccar.store.AllAttributeAliases', {
```
tananaev: Can you rename it to just `AttributeAliases`?

Abyss777:
Sure, I'll rename.
Found some issue: if aliases editor are open and position are updated filter in editor will be reset.
I'll try to do the magic with chained stores.

Abyss777:
Two stores was really more simple.
I have to create 2 chained stores, reimplement `AttributeController` just to remove `store.sync` because `ChainedStores` does not have such function. Also find out that changing proxy extra param limiting whole store.
Can I back to two store scheme? It will be really more simple.

tananaev:
Wait, why do you need two chained stores? You just call `sync` on the original store. There must be a nice simple way to do it.

Abyss777:
If I filter original store it will impact on chained store. I need two different filters working at the same time for `State` and for `Editor`

tananaev:
You don't filter original store. That's what chained store is for.

For state you just use original store. Why do you need to filter it?

For editor you use chained store, but modify original store directly.

Abyss777:
For state I filter by `deviceId` and then use `findRecord` by attribute. I believe it is faster then use findBy function to search by two fields. Also because it is filtered I just call `findRecord` in `onAliasEditClick`.

OK, I will try again.

Abyss777:
Seems I managed do everything in right way.
Had to make dedicated `findAttribute` function and pass parameters via `this` because `jshint` complains "Do not make functions in the loop"

Sorry for spending your time.

tananaev:
Looks good now. Thank you.

PR #291, opened 2016-10-05
Tree view for devices panel

ninioe:
Hi Anton,
This treeview is based on your work.
I made some changes to the store and the filter.
Please try this...

Thanks
Erez


tananaev:
An issue I found straight away:

screen shot 2016-10-08 at 5 21 50 pm

Please do a thorough testing of every use case you can think of.

ninioe:
Hi Anton,
Sorry but I couldn't understand what is the issue from this picture, could you explain?

Thanks 
Erez

tananaev:
Status field is empty. It should be yellow if it's unknown.

ninioe:
What to write inside?

tananaev:
Last update time or nothing if there is no timestamp. Please compare with
current implementation. The behaviour should be exactly the same.

ninioe:
Hi Anton,
Before on the render event I spiked empty values, but now I only skip the groups entries.

var isGroup = (record.get('id').substr(0, 1) === 'g');
if (isGroup) {
    return;
}

Please try it now and tell me if this is OK.

Thanks
Erez

tananaev:
Filtering doesn't work properly. Issues:

1. Group filtering doesn't work and it should be removed because it doesn't make sense if we have tree.
2. Keyword filtering should work not only from the beginning of the work, but from in any part of the name/id. Logic should be the same as before.
3. There are issues with keyword filtering. Apply filter, add device, try to remove filter. Nothing happens.

ninioe:
Sorry for this, I will fix it...

ninioe:
Hi Anton,
I need your help.
I have issue when I add new device and a filter exist.
I don't know why is it acting different and I prefer to solve it without workaround.

Can you help me with this?

ninioe:
maybe we can clear the filter before we update the store and put it back again after?

tananaev:
That sounds like a workaround and I suspect that it won't look nice. I think you need something like chained store, but I don't think it works for trees. Possibly you need to develop something similar yourself.

ninioe:
sorry but I don't know about chained store.
can we just clear the filter & textbox when a new device is added?
or maybe you could solve this?

ninioe:
Hi Anton,
I solved the filter issue when adding new device.
Please check my solution...

Thanks
Erez

tananaev:
There is another problem. If you collapse any of the groups, then edit devices, all groups expand. It shouldn't happen.

ninioe:
Yes it was on the original function that update the store, I can fix this...

tananaev:
Yes please. Make sure everything else works correctly after the fix.

ninioe:
Hi Anton,
If I remove this line from storeUpdate() function:
this.getView().expandAll();

all the groups are collapsed after the store load:
this.getView().getStore().load();

Is the any way to remember the tree state?

ninioe:
Maybe something like this?
https://www.sencha.com/forum/showthread.php?24399-How-to-save-TreePanel-state-after-reload()/page2

ninioe:
what about this one?
https://github.com/AlexTiTanium/ExtJsStatefulTree

ninioe:
Hi Anton,
I solved this issue, and didn't find any more issues.
Please try this, and if you find any issue let me know.

Thanks
Erez

tananaev:
Another issue. "Last Updated" column doesn't get updated.

screen shot 2016-10-11 at 6 27 09 pm

ninioe:
Hi Anton,
Sorry I missed it.
I will check this on Thursday...

Thanks
Erez

ninioe:
Hi Anton,
I think it's ready now.
The last commit didn't solve the conflict but I have marge your last changes inside DevicesController.js.
I tested it and it works OK for me.

Please let me know if you find any issue I missed...

Thanks
Erez

tananaev:
Please fix the merge issue.

ninioe:
How can i do this?
The conflict is with your branch, only you can solve it.
All your last changes are inside this file version, If you have no issue with my work you can accept the changes and merge it.

ninioe:
Hello Anton,
How can we continue with this?

tananaev:
I have tested it again. Two problems:

1. By default everything should be expanded.
2. Every time any data is updated, selection is lost and if you have a long list, it always scrolls up to the top.

ninioe:
Hi Anton,
Thanks for testing this.
Do you want me to continue working on this, do you think we can use this if I restore the selection too?

tananaev:
I think the only way to do it properly is implement a "wrapper" store that won't be recreated every time, but rather handle any update in the original store and execute it in the tree store.

ninioe:
Can you do the wrapper?
I'm not so expert in Ext JS like you are.

tananaev:
I'm not a huge expert either, and it's not the highest priority on my list.

PR #292, opened 2016-10-07
Draw trip route on click

Abyss777:
also
- Add some object class checks
- Add missed imports

It was easier than I think. Can be some delays on slow connections or with many positions.


tananaev:
Looks good, thanks for pull request.

PR #295, opened 2016-10-11
Always show geofences on the Map

Abyss777:
Geofences are showed by default.

There is strange bug, if I set header in a view config - I get error `header.hide is not a function` If set it manually it is OK. In `State` panel it works OK. Weird.

Button looks alone, but when we move settings there, hope it won't looks so.
Just clear and show again on any geofences updates. Don't think we need spend time on precise synchronizing features and store.

image

### DIFF
--- a/web/app/view/MapController.js
+++ b/web/app/view/MapController.js
@@ -21,7 +21,8 @@ Ext.define('Traccar.view.MapController', {
 
     requires: [
         'Traccar.model.Position',
-        'Traccar.model.Device'
+        'Traccar.model.Device',
+        'Traccar.GeofenceConverter'
     ],
 
     config: {
@@ -45,6 +46,12 @@ Ext.define('Traccar.view.MapController', {
                 '#ReportRoute': {
                     load: 'loadReport',
                     clear: 'clearReport'
+                },
+                '#Geofences': {
+                    load: 'showGeofences',
+                    add: 'updateGeofences',
+                    update: 'updateGeofences',
+                    remove: 'updateGeofences'
                 }
             },
             component: {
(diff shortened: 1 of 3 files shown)


web/app/view/MapController.js
```diff
@@ -58,6 +65,20 @@ Ext.define('Traccar.view.MapController', {
     init: function () {
         this.latestMarkers = {};
         this.reportMarkers = {};
+        this.getView().header = {
+            xtype: 'header',
+            title: Strings.mapTitle,
+            items: [{
+                    xtype: 'button',
```
tananaev: Indentation is wrong here?

web/app/view/MapController.js
```diff
@@ -321,5 +342,55 @@ Ext.define('Traccar.view.MapController', {
...
+                    }),
+                    font : Traccar.Style.mapTextFont
+                })
+            });
+    },
+
+    showGeofences: function () {
+        if (this.lookupReference('showGeofencesButton').pressed) {
+            Ext.getStore('Geofences').each(function (geofence) {
+                var feature = new ol.Feature(Traccar.GeofenceConverter.wktToGeometry(this.getView().getMapView(), geofence.get('area')));
```
tananaev: This line looks pretty long. Are you sure `jscs` doesn't complain about it?

Abyss777:
`jscs` and `jshint` do not complain about line length but I added new line anyway.
Fixed indentation. Also labels look centered now
image

tananaev:
Perfect, thanks.

Abyss777:
Do not think you have ability to change committer name and mail in this commits , right?
It is lesson to me be more intent next time.

tananaev:
No, unfortunately it's not possible to change committer after it's been pushed to remote server. I mean it's technically possible, but it will effectively break history for everybody who's using this project.

Abyss777:
OK. I understand. Thanks.

PR #296, opened 2016-10-12
Automatically scroll to selected device or position

### LINKED ISSUE #294: Automatically scroll to position
It would be nice to automatically scroll to selected device when it's clicked on map. Would be useful for some reports as well.

Code to use:
```JavaScript
this.getView().getView().focusRow(record);
```

Abyss777:
fix #294 

Also fixed state clear on report clear

### DIFF
--- a/web/app/view/DevicesController.js
+++ b/web/app/view/DevicesController.js
@@ -143,6 +143,7 @@ Ext.define('Traccar.view.DevicesController', {
 
     selectDevice: function (device, center) {
         this.getView().getSelectionModel().select([device], false, true);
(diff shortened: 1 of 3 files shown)


tananaev:
What happens when you click on an item in the grid? Does it call focus as well? Does it scroll it in that case?

tananaev:
By the way, who's `kas`?

Abyss777:
You are right, It is double focused I'll think how to split source of events.
Oh, how I forgot change it, `kas` it is my company login.
Better recreate pull request then fix double focus and change commiter.

PR #297, opened 2016-10-12
Autoscroll to device and position selected on map

### LINKED ISSUE #294: Automatically scroll to position

Abyss777:
Second attempt
fix #294

- Limited `selectdevice` event to `map` in `DeviceController`
- Limited `selectreport` event to `map` in `ReportController`

It will allow not react on events fired by itself.

Also fixed clear State


tananaev:
Looks good, merged.

PR #300, opened 2016-10-17
Export reports to excel

Abyss777:
Web part for  https://github.com/tananaev/traccar/pull/2448

- Replaced CSV button to XLSX
- Adapted download function to binary files

Also
- Rewrite dom manipulation to extjs

### DIFF
--- a/web/app/view/ReportController.js
+++ b/web/app/view/ReportController.js
@@ -74,11 +74,11 @@ Ext.define('Traccar.view.ReportController', {
         time = this.fromDate && this.fromTime && this.toDate && this.toTime;
         disabled = !reportType || !devices || !time;
         this.lookupReference('showButton').setDisabled(disabled);
-        this.lookupReference('csvButton').setDisabled(disabled);
+        this.lookupReference('xlsxButton').setDisabled(disabled);
     },
 
     onReportClick: function (button) {
-        var reportType, from, to, store, url;
+        var reportType, from, to, store, url, accept;
 
         reportType = this.lookupReference('reportTypeField').getValue();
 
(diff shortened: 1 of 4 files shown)


web/l10n/en.json
```diff
@@ -160,7 +160,7 @@
     "reportSummary": "Summary",
     "reportConfigure": "Configure",
     "reportEventTypes": "Event Types",
-    "reportCsv": "CSV",
+    "reportXlsx": "XLSX",
```
tananaev: I think it's better to just have an "Export" button.

web/app/view/ReportController.js
```diff
@@ -102,9 +102,10 @@ Ext.define('Traccar.view.ReportController', {
                         to: to.toISOString()
                     }
                 });
-            } else if (button.reference === 'csvButton') {
+            } else if (button.reference === 'xlsxButton') {
                 url = this.getView().getStore().getProxy().url;
-                this.downloadCsv(url, {
+                accept = 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet';
+                this.downloadFile(url, accept, {
```
tananaev: Do we really need "accept" variable and parameter if we only have one option?

Abyss777: I was thinking to make universal function to download files, but you are right, will change it in future if it will be necessary.

Abyss777:
Done.

tananaev:
Looks good. Holding off on merging this until server side is complete.

Abyss777:
Synchronized `allEvents` constant with server part

tananaev:
Back-end is ready, so merging this one as well.

PR #305, opened 2016-10-23
Implement reset device distance dialog

Abyss777:
Added to settings menu
image

image

Also added label to device selector in AttributeAliases to make it more clear

image


tananaev:
Looks good, thanks.

PR #306, opened 2016-10-23
Corretion in tradution pt-BR

visued:


tananaev:
Please correct it on Transifex:

PR #308, opened 2016-10-27
Implement live routes

Abyss777:
- Live routes are drawn always, the button just toggles layer visibility
- Tails length can be configured via `web.liveRouteLength` server or user attribute.
- Added wrapper `getAttributePreference` to use it in future

image

### DIFF
--- a/web/app/Style.js
+++ b/web/app/Style.js
@@ -77,5 +77,7 @@ Ext.define('Traccar.Style', {
     coordinatePrecision: 6,
     numberPrecision: 2,
 
-    reportTagfieldWidth: 375
+    reportTagfieldWidth: 375,
+
+    headerButtonsMargin: '0 5'
 });
--- a/web/app/view/MapController.js
+++ b/web/app/view/MapController.js
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 Anton Tananaev (anton.tananaev@gmail.com)
+ * Copyright 2015 - 2016 Anton Tananaev (anton.tananaev@gmail.com)
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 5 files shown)


web/app/view/MapController.js
```diff
@@ -166,6 +182,43 @@ Ext.define('Traccar.view.MapController', {
                 if (marker === this.selectedMarker && this.followSelected()) {
                     this.getView().getMapView().setCenter(marker.getGeometry().getCoordinates());
                 }
+
+                if (deviceId in this.liveRoutes) {
+                    lastLiveCoordinates = this.liveRoutes[deviceId][this.liveRoutes[deviceId].length - 1].getGeometry().getCoordinates();
+                    if (lastLiveCoordinates[0] === geometry.getCoordinates()[0] &&
```
Abyss777: This `if` prevent tails cuts every 30 seconds in debug mode because of websocket re-connection. Also it is useful if device send positions (with same coordinates) while parking.
Only thing is `jscs` complains about `continue` statement.

What should I do?
- Remove workaround
- Refactor code to avoid `continue`
- Configure `jscs` to not complain about `continue`

web/app/Style.js
```diff
@@ -77,5 +77,7 @@ Ext.define('Traccar.Style', {
     coordinatePrecision: 6,
     numberPrecision: 2,
 
-    reportTagfieldWidth: 375
+    reportTagfieldWidth: 375,
+
+    headerButtonsMargin: '0 5'
```
tananaev: What is this for?

web/app/view/MapController.js
tananaev: I think we should get rid of "continue". Also, it would be nice to split this huge method into separate smaller methods for more readability.

web/app/view/MapController.js
```diff
@@ -166,6 +182,43 @@ Ext.define('Traccar.view.MapController', {
...
+                            lastLiveCoordinates,
+                            ol.proj.fromLonLat([
+                                position.get('longitude'),
+                                position.get('latitude')
+                            ])
+                        ])
+                    });
+                    liveLine.setStyle(this.getRouteStyle(position.get('deviceId')));
+                    this.liveRoutes[deviceId].push(liveLine);
+                    this.getView().getLiveRouteSource().addFeature(liveLine);
```
tananaev: Do we add line for every position item? Can't we extend existing line instead?

web/app/Style.js
Abyss777: Five is left margin, without this it look like:
image

Abyss777:
Split `updateLatest` function
Used one `lineString` 
I was thinking, If we do not react on marker click it can confuse user, I just removed markers. Looks good:

image

Also `continue` become `return` after splitting.

tananaev:
Looks really good, thanks.

PR #311, opened 2016-11-04
Adjust textfield width to Triton theme

Abyss777:
Buttons look OK now

image


tananaev:
I want to avoid changing any theme parameters, so I was thinking to change save and cancel buttons to be icon-only. You can easily find good icons for those.

tananaev:
Updated all dialogs to use icons for save and cancel buttons: 166237d0c15ea228105afa65958d8a2889aab05b.

Abyss777:
Close it then.

PR #312, opened 2016-11-06
Fix close geofence geometry edit window after cancel click.

Abyss777:
Geofence geometry edit window do not close on Cancel click.


tananaev:
Thanks.

PR #313, opened 2016-11-07
Draft of device images

Abyss777:
Here is draft, actually fully working.

Added 3 images (car, bus and truck), and arrow as svg. Set CC BY-AS license for images (It is most similar to GPL)

Removed width and height from store items, we need only scale and id of elements to fill and rotate.
Images must be squared and centered to not crop it while rotation, also directed vertically up (0 degree).

Leaved comments for now. Tested in Chrome, FF and IE (it was most capricious as usual)

Known issue is Label offset. Also have not implemented yet selection in device dialog.

image

Mobile Chrome render it much better (on phone, screenshot looks worse)
image


web/app/DeviceImages.js
```diff
@@ -0,0 +1,74 @@
...
+        svg.getElementById(device.get('rotateId')).setAttribute('transform', transform);
+
+        // Set dimension attributes
+        svg.documentElement.setAttribute('width', width);
+        svg.documentElement.setAttribute('height', height);
+        svg.documentElement.setAttribute('viewBox', '0 0 ' + width + ' ' + height);
+   
+        image =  new ol.style.Icon({
+            imgSize: [width, height], // Workaround for IE
+            src: 'data:image/svg+xml;charset=utf-8,' + encodeURIComponent((new XMLSerializer()).serializeToString(svg.documentElement))
```
tananaev: You don't need parentheses around new operator.

web/app/store/DeviceImages.js
```diff
@@ -0,0 +1,69 @@
...
+ */
+Ext.define('Traccar.store.DeviceImages', {
+    extend: 'Ext.data.Store',
+    fields: ['key', 'name', 'url', 'svg', 'fillId', 'rotateId', 'scale'],
+
+    data: [{
+        key: 'default',
+        name: Strings.categoryDefault,
+        svg: (new DOMParser())
+                .parseFromString('<svg version="1.1" xmlns="http://www.w3.org/2000/svg" width="26" height="26">'
```
tananaev: I think it's better to put it into a file.

web/app/store/DeviceImages.js
```diff
@@ -0,0 +1,69 @@
...
+        fillId: 'path4718',
+        rotateId: 'layer2',
+        scale: 0.1,
+    }],
+    
+    constructor: function() {
+        this.callParent(arguments);
+        this.config.data.forEach(function (device) {
+            if (device.url) {
+                Ext.Ajax.request({
```
tananaev: Any potential issues with async download of the images?

Abyss777: I have not come across, but it is possible.
Do not have another idea for now. Will try to search something.

Abyss777:
About pictures...
Buying is easiest way, but we have to check carefully license or conditions if we can use them with GPL project.

You can mail me example how it should looks like, I will try to redraw it.

Want to note, we do not need high detailed pictures, its final size about 30-40 pixels (45-60 zoomed)

tananaev:
I don't really know how they should look like. I think the style should be the same a the arrow. It seems like there is some border on other icons, but it's very thin. Are windows transparent? They should be solid, probably white or grey color.

I'll try to search for some icon sets. If you come across some nice images that we can use, we can buy them. That's not a problem.

Abyss777:
It is caused by scaling. One pixel width stroke become 1/10 pixel, and rendered as light grey line.
Make it bolder a bit:
image

If I make its width same as arrow it will blend.

tananaev:
You can make arrow outline a bit thinner and icons outline a bit thicker. I just think they should be consistent.

Abyss777:
image

Can not find whole set of transportation icons, they are mostly divided by types

Here is main types http://www.shutterstock.com/pic-244859614/stock-vector-set-of-transportation-flat-icons-vector-illustration.html

Here is some city vehicles https://ru.fotolia.com/id/113610190
Also this author has more icons https://ru.fotolia.com/search?k=top+view&filters%5Bcontent_type%3Avector%5D=1&nca=1013&p=204220148&search-submit=%D0%98%D1%81%D0%BA%D0%B0%D1%82%D1%8C

May be better contact with some of this designers and buy whole set.

tananaev:
What do you think about using images like this one:

icon

It's basically font awesome icon inside a circle. Circle is to show direction.

Easier to find icons. Also, icons are much more understandable. It means we can have a person there for example. It's really hard to find good top-view icons.

Abyss777:
May be you are right about top view. Top view of dog and cow will look strange :) 
Any way we prepared infrastructure to use what ever users want.

I'll do few icons and we will see how it will look.

About async download...
My suggestion is to put them in html (as objects) and access by Id from store. Just an idea, have not tested.

tananaev:
I think we can include them in the HTML. That might solve any potential issues.

Abyss777:
Implemented loading svg to `document`. I like how it turned out.

About new pictures...
as I understand we will rotate only "circle-with-arrow" and internal icon will be steady, then we need split elements that rotate and that scale.

tananaev:
Yes, we want to rotate circle only. I thought you can already specify element to rotate.

Abyss777:
Yes, and scale applied to the same element. But now we need rotate one of sub-elements and scale whole group of elements. It does not matter. It is small addition.

tananaev:
Do you we really need to scale? Can't we just normalize images and save them in correct scaling?

Abyss777:
And what about scale on select?

tananaev:
Good point, I completely forgot about that. Anyway, it would be nice to normalize images so that scale is 1 for normal size.

Abyss777:
Sure, for such icon style it will be logical.

Abyss777:
image

They looks better on phone
image

What do you think?

tananaev:
I think looks good. I think we can adjust following things:

- Make circle thinner
- Make icon inside a bit bigger
- I think space between inner icon and circle should be transparent

Otherwise I think it looks really good. Also we probably have all the icons we need in the FontAwesome people and pets. What do you think?

tananaev:
One thing... maybe for reports and trail we should consider something simpler otherwise it's gonna be too much clutter.

Abyss777:
Will try your suggestions.
There is no pets in FontAwsome, but I going to add male and female.

Sure, now reports show as arrows as was.

tananaev:
Just one more idea. If we keep arrows, it would be nice to make arrow on the circle same angle as arrow icon, so it would look consistent.

Abyss777:
It is for latest devises. But I think we should leave bare arrow for reports or may be rhombus.

Abyss777:
Bus is selected
image

Route
image

Abyss777:
On the phone
screenshot_2016-11-08-17-21-37

tananaev:
I think it looks good. I don't think we need arrow in the circle though. Maybe we should just default to car icon? Or just show a simple arrow if nothing is selected.

Abyss777:
Let's be consistent, device is not always a car :)

Here is without arrow
image

Or may be something neutral?
image

tananaev:
Maybe we can have just a circle inside? That would be more neutral.

Abyss777:
Sure
image

Abyss777:
Agree

I will add rest a few icons
- Ship http://fontawesome.io/icon/ship/
- Plane http://fontawesome.io/icon/plane/
- Motorcycle http://fontawesome.io/icon/motorcycle/
- Bicycle http://fontawesome.io/icon/bicycle/


- Male http://fontawesome.io/icon/male/
- Female http://fontawesome.io/icon/female/


- Mobile http://fontawesome.io/icon/mobile/

I've found one pet :) http://fontawesome.io/icon/linux/

Then I will try to implement combobox to select.

tananaev:
I was thinking for person just use one:

http://fontawesome.io/icon/user/

I don't think we need mobile icon. Don't need Linux either.

We can add more later.

Abyss777:
What do you think about internal icon without stroke? (like in original font) For some icons (like bicycle) lines too thin and stroke fill almost all lines.

tananaev:
Maybe we can exclude bike for now.

Abyss777:
It is not a way out. Ship has the same problem with waves.
screenshot_2016-11-09-13-46-53

Here is without stroke
screenshot_2016-11-09-13-49-18

It looks pale without stroke.

What do you think?

tananaev:
Just thought of another option. What if we make circle solid and then just put back icon inside? This way we can even reduce the size of the circle a little bit.

Abyss777:
Like this?
bicycle

tananaev:
Yes, something like that. Maybe we need to make background color a bit lighter to have better contrast.

Abyss777:
Made offline color a bit lighter, green and yellow are OK.
image

Abyss777:
Some other icons, Test2 is default
image

tananaev:
It just doesn't feel right for some reason. Maybe make icon a bit smaller inside because it's too close to the border. I'm also thinking that maybe instead of black we should use some dark shade of grey. Let's try this last change and then finish with experimenting for now... we can always change icons in future.

Abyss777:
Make icons a bit smaller

Black
image

80% Grey
image

tananaev:
OK, let's stop on the black version for now. Maybe I'll experiment with icons a bit more later.

Abyss777:
Implemented combobox with pictures
image
image

Selected value is without picture
image

I leaved possibility to select bare arrow to user. May be somebody does not like default icon.

Frankly, I'm not guru of html and combobox items formatted as `<table>`.

The rest thing to do is update device icon on map after editing the device object.

Abyss777:
Looks like all done.
I can squash all commits here, or make new pull-request.

web/app/DeviceImages.js
```diff
@@ -0,0 +1,88 @@
...
+Ext.define('Traccar.DeviceImages', {
+    singleton: true,
+
+    getImageSvg: function (color, zoom, angle, category) {
+        var i, device, svg, width, height, rotateTransform, scaleTransform, fill;
+
+        if (category) {
+            device = Ext.getStore('DeviceImages').findRecord('key', category, 0, false, false, true);
+        }
+        if (device === undefined || device === null) {
```
tananaev: This can be replaced with just `if (device) {`.

web/app/DeviceImages.js
```diff
@@ -0,0 +1,88 @@
...
+            fill = [fill];
+        }
+        for (i = 0; i < fill.length; i++) {
+            svg.getElementById(fill[i]).style.fill = color;
+        }
+
+        rotateTransform = 'rotate(' + angle + ' ' + (width / 2) + ' ' + (height / 2) + ')';
+        svg.getElementById(device.get('rotateId')).setAttribute('transform', rotateTransform);
+
+        width *= Traccar.Style.mapScaleNormal;
```
tananaev: Do we need this? Can't normal be 1?

web/app/view/DeviceDialog.js
```diff
@@ -57,6 +57,22 @@ Ext.define('Traccar.view.DeviceDialog', {
...
+            xtype: 'combobox',
+            name: 'category',
+            fieldLabel: Strings.deviceCategory,
+            store: 'DeviceImages',
+            queryMode: 'local',
+            displayField: 'name',
+            valueField: 'key',
+            listConfig: {
+                getInnerTpl: function () {
+                    return '<table><tr valign="middle" ><td><div align="center" style="width:40px;height:40px;" >' +
```
tananaev: I think it looks good enough for now... we can improve it in future if there are any issues.

Abyss777:
OK. Fixed `device` check.

tananaev:
Merged, thanks. This is a really important feature. Now we don't need to have arrow implementation for ol3 library.

PR #314, opened 2016-11-11
Try new style for images

tananaev:
@Abyss777, I have experimented a little bit with images and here is what I ended up with:

screen shot 2016-11-12 at 10 49 00 am

Arrow is smaller and outside of the inner-circle area, which solves the problem that icon inside looked misplaced.

Also, I've doubled border thickness. I think it looks better because icon inside is usually has think lines.

By the way, arrow has the same angle as "arrow" image, so it looks more consistent between two.

I also removed some rubbish added by Inkscape to reduce image size a little bit.

What do you think?


Abyss777:
Agree, old transition from circle to arrow sometimes looked curved a bit.
And border looks consistent to icon now.
  
It looks very careful now. I very like it.
Good work :+1:

tananaev:
OK, thanks for feedback. Merged.

PR #315, opened 2016-11-14
Fix icon blinking

Abyss777:
Hope I've fixed it. Please check.

Key changes:
- `image.load()` just in case, load image as soon as possible.
- Added checks to avoid blank/double style updates

I think significant change was moving `marker.setGeometry(geometry);` after `marker.setStyle()`

Surprising fact that it did not blink in IE.


tananaev:
Unfortunately it didn't help. Same flickering on updates.

PR #316, opened 2016-11-14
Remove marker style recreation on marker rotate and color change

Abyss777:
Here is another try of fixing flickering, hope it is last.

Here we do not create new style on every update, just updating `src` of inner `` object. After `setGeometry` browser draws new image.

Also do the same then updating device, but we have to call `changed()` to re-render image, otherwise it will be updated in a few seconds.

Moved this functions to `DeviceImages` because it is better place and to unload `MapController` a bit


tananaev:
The issue seems to be fixed, thanks.

PR #318, opened 2016-11-15
Workaround for kitkat

Abyss777:
Added workaround

It is working OK now on 4.4 default browser in emulator.
screenshot_1479191350

### DIFF
--- a/web/app/DeviceImages.js
+++ b/web/app/DeviceImages.js
@@ -70,10 +70,10 @@ Ext.define('Traccar.DeviceImages', {
                 encodeURIComponent(new XMLSerializer().serializeToString(svg.documentElement));
     },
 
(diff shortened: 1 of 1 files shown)


web/app/DeviceImages.js
```diff
@@ -67,6 +70,14 @@ Ext.define('Traccar.DeviceImages', {
                 encodeURIComponent(new XMLSerializer().serializeToString(svg.documentElement));
     },
 
+    cloneDocument: function (document) {
```
tananaev: I would rename "document" to something else because "document" is a global object in java.

Abyss777:
You are right. Sorry. Renamed.

tananaev:
Merged, thanks.

PR #320, opened 2016-11-16
Fix update marker label on device name change

Abyss777:
If we are not recreating style, we have to watch text consistence too.


tananaev:
Merged, thanks.

PR #321, opened 2016-11-16
Move Follow button to Map toolbar

Abyss777:
I think it is more intuitive place, especially in mobile version.

Also we removed one `getCmp`


tananaev:
Totally makes sense. Merged.

PR #322, opened 2016-11-17
Add user expiration and device limit

tananaev:
@Abyss777, can you please review the changes.


Abyss777:
Looks good for me.
I liked idea that user can see his expiration date and device limit.

One wish is say to user somehow that he can be expired/disabled. May be make `Strings.loginFailed` something like this:
"Incorrect email address/password or user disabled/expired"

tananaev:
Fixed this one as well. Expired and disabled accounts are now general errors.

Abyss777:
Here is new problem:
It happen after user disabling and page refresh.

image

I think login screen won't appear until cookies will be cleared.

tananaev:
I noticed it as well. You can't even logout, but I think it's a very rare case when you get locked in the middle of the session, so I don't think we need to handle this case.

tananaev:
What do you think?

PR #323, opened 2016-11-18
Implement users tokens

Abyss777:
Here is web part

If token in uri it is passed to `api/session` request

Also added generator just to comfort

image

### DIFF
--- a/web/app/controller/Root.js
+++ b/web/app/controller/Root.js
@@ -37,11 +37,14 @@ Ext.define('Traccar.controller.Root', {
         var tokenParameter;
         if (success) {
             Traccar.app.setServer(Ext.decode(response.responseText));
-            tokenParameter = window.location.search.match('/token=([a-zA-Z0-9]+)/');
-            tokenParameter = tokenParameter !== null ? '?' + tokenParameter[0] : '';
+            tokenParameter = window.location.search.match('token=([a-zA-Z0-9]+)');
             Ext.Ajax.request({
                 scope: this,
-                url: 'api/session' + tokenParameter,
+                url: 'api/session',
+                method: 'GET',
+                params: tokenParameter !== null ? {
+                    token: tokenParameter[1]
+                } : {},
                 callback: this.onSessionReturn
             });
         } else {
--- a/web/app/controller/Root.js
+++ b/web/app/controller/Root.js
@@ -34,11 +34,14 @@ Ext.define('Traccar.controller.Root', {
     },
 
     onServerReturn: function (options, success, response) {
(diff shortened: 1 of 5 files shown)


web/app/controller/Root.js
```diff
@@ -34,11 +34,14 @@ Ext.define('Traccar.controller.Root', {
...
     onServerReturn: function (options, success, response) {
+        var tokenParameter;
         if (success) {
             Traccar.app.setServer(Ext.decode(response.responseText));
+            tokenParameter = window.location.search.match('/token=([a-zA-Z0-9]+)/');
+            tokenParameter = tokenParameter !== null ? '?' + tokenParameter[0] : '';
             Ext.Ajax.request({
                 scope: this,
-                url: 'api/session',
+                url: 'api/session' + tokenParameter,
```
tananaev: I believe there is a standard way to pass query parameters into the ExtJs Ajax request.

web/app/view/UserDialogController.js
```diff
@@ -26,9 +26,23 @@ Ext.define('Traccar.view.UserDialogController', {
...
             this.lookupReference('deviceLimitField').setDisabled(false);
+            this.lookupReference('tokenField').setDisabled(false);
         }
     },
 
+    symbols: 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789',
+
+    generateToken: function () {
+        var newToken, i;
+        newToken = '';
```
tananaev: I think this can be moved to declaration. Just make `newToken` last variable.

web/app/controller/Root.js
```diff
@@ -34,11 +34,17 @@ Ext.define('Traccar.controller.Root', {
...
         if (success) {
             Traccar.app.setServer(Ext.decode(response.responseText));
+            tokenParameter = window.location.search.match('token=([a-zA-Z0-9]+)');
             Ext.Ajax.request({
                 scope: this,
                 url: 'api/session',
+                method: 'GET',
+                params: tokenParameter !== null ? {
+                    token: tokenParameter[1]
+                } : {},
```
Abyss777: Is formatting OK? Ternary operator is just to not get `api/session?token=` requests

tananaev: I think you can do something like this instead:
```javascript
var token, parameters = {};
...
token = ...;
if (token) {
    parameters.token = token;
}
```

Also, I think there is a function to parse query parameters in ExtJS. Cleaner than using regex.

Abyss777:
Sorry. It looks really cleaner now.

PR #324, opened 2016-11-18
Add Device Name column to Route report

Abyss777:
Reused function as AttributeFormatter
fix https://github.com/tananaev/traccar/issues/2554

### DIFF
--- a/web/app/AttributeFormatter.js
+++ b/web/app/AttributeFormatter.js
@@ -47,6 +47,10 @@ Ext.define('Traccar.AttributeFormatter', {
         return (hours + ' ' + Strings.sharedHourAbbreviation + ' ' + minutes + ' ' + Strings.sharedMinuteAbbreviation);
     },
 
+    deviceNameFormatter: function (value) {
+        return Ext.getStore('Devices').findRecord('id', value).get('name');
+    },
+
     defaultFormatter: function (value) {
         if (typeof value === 'number') {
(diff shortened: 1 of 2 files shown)


web/app/AttributeFormatter.js
```diff
@@ -47,6 +47,10 @@ Ext.define('Traccar.AttributeFormatter', {
         return (hours + ' ' + Strings.sharedHourAbbreviation + ' ' + minutes + ' ' + Strings.sharedMinuteAbbreviation);
     },
 
+    deviceNameFormatter: function (value) {
+        return Ext.getStore('Devices').findRecord('id', value).get('name');
```
tananaev: There is a special method `getById` or something like that. Potentially it's more efficient.

web/app/AttributeFormatter.js
```diff
@@ -78,6 +82,8 @@ Ext.define('Traccar.AttributeFormatter', {
             return this.hoursFormatter;
         } else if (key === 'duration') {
             return this.durationFormatter;
+        } else if (key === 'deviceName') {
```
tananaev: Key should be `deviceId`. That's the naming convention to have the same name as the model key.

Abyss777:
Sorry. Fixed.

tananaev:
Merged, thanks. I guess this one goes into the next release.

PR #325, opened 2016-11-18
Add th lang

CVSz:
user expiration
device limit
device image
Category
Attribute
Device Distance


tananaev:
All localization should be done via Transifex platform:

PR #326, opened 2016-11-19
Add disabled and expirationTime columns

Abyss777:
Added "Disabled" column and hidden by default "Expiration" column to users dialog

image

image

### DIFF
--- a/web/app/view/Users.js
+++ b/web/app/view/Users.js
@@ -88,13 +88,7 @@ Ext.define('Traccar.view.Users', {
             text: Strings.userExpirationTime,
             dataIndex: 'expirationTime',
             hidden: true,
-            renderer: function (value) {
-                if (Traccar.app.getPreference('twelveHourFormat', false)) {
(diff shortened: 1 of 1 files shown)


web/app/view/Users.js
```diff
@@ -81,6 +81,20 @@ Ext.define('Traccar.view.Users', {
...
             dataIndex: 'admin'
+        }, {
+            text: Strings.userDisabled,
+            dataIndex: 'disabled'
+        }, {
+            text: Strings.userExpirationTime,
+            dataIndex: 'expirationTime',
+            hidden: true,
+            renderer: function (value) {
+                if (Traccar.app.getPreference('twelveHourFormat', false)) {
```
tananaev: Would it be possible to share this logic somewhere. For example in AttributeFormatter.

Abyss777:
Sorry, It is already there.

tananaev:
Merged, thanks.

PR #330, opened 2016-11-20
Decrease chances of not unique tokens

Abyss777:
- Increase token length
- Make not editable
- Add clear trigger

image


tananaev:
I don't like how it looks with two buttons. Maybe we should keep it editable. Just make the generated length higher. I think power of two would be nicer (32?).

tananaev:
Thanks. We should probably change it on the back-end side as well.

Abyss777:
Sure.

PR #333, opened 2016-11-23
Pass client timezone to excel export

Abyss777:
Web part of https://github.com/tananaev/traccar/pull/2592


tananaev:
Do we have similar code anywhere else? We should probably change other places as well. Statistics table comes to mind. Do we use `toISOString` anywhere else?

Abyss777:
Only `StatisticsController` and `ReportController.
Then we have to change server side as well. Format is different a bit.

tananaev:
OK, let's leave it as it is for now for statistics.

PR #334, opened 2016-11-24
Add string for maintenance event

Abyss777:
subj


tananaev:
Decided to rename maintenance events a little bit. See f1038fa65fe2befb2a1828c46cab99de6b560f67.

PR #337, opened 2016-11-26
Totally update device marker instead of updating only image src

### LINKED ISSUE #335: Arrow is not centered
I have noticed that the arrow icon is not centered:

screen shot 2016-11-25 at 1 14 03 pm

Abyss777:
fix #335 

The main problem that `ol.style.Icon` does not have necessary setters to correctly set `src` and `size`.
Therefore changing all other icon to arrow or arrow to all other will cause visual bug like not centered or what is worse
image

To fix it, I have to adjust some private fields of `ol.style.Icon` that is prohibited by our style requirements.
Because of that I reverted total style recreation on device update.

PS: may be we have to spend some time to contribute necessary setters to openlayers, it is welcomed https://github.com/openlayers/ol3/issues/3137#issuecomment-241994282

### DIFF
--- a/web/app/DeviceImages.js
+++ b/web/app/DeviceImages.js
@@ -98,15 +98,8 @@ Ext.define('Traccar.DeviceImages', {
     },
 
     rotateImageIcon: function (image, angle) {
-        var svg = Traccar.DeviceImages.getImageSvg(image.fill, image.zoom, angle, image.category);
+        var svg = this.getImageSvg(image.fill, image.zoom, angle, image.category);
         image.getImage().src = this.formatSrc(svg);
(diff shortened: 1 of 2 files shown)


tananaev:
Does it affect performance? Any flickering?

Abyss777:
Not, I've tested. 
Firstly, It happens only if something really changed. (color or category) not every socket update as before.
Secondly, actually flickering was fixed here https://github.com/tananaev/traccar-web/blob/master/web/app/DeviceImages.js#L100 it is not changed.
And it works correctly because size of icon is not changing while rotation.

tananaev:
Thanks.

PR #340, opened 2016-11-28
Split MapController

Abyss777:
Split `MapController` as you suggested some time ago. It became really huge.

Made it in way of inheritance. Moved all functions connected to markers to `MapMarkerController` (devices, routes and reports)

Buttons handlers and geofences leaved in `MapController`

Theoretically bare `MapMarkerController` can be used with something like mini-site

### DIFF
--- a/web/app/view/MapController.js
+++ b/web/app/view/MapController.js
@@ -16,114 +16,38 @@
  */
 
 Ext.define('Traccar.view.MapController', {
-    extend: 'Ext.app.ViewController',
+    extend: 'Traccar.view.MapMarkerController',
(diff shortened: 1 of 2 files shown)


tananaev:
Are there any changes in the functions themselves? If not, can you please redo it this way:

1. First commit just copy the file.
2. Second commit do the changes.

This way we can see the difference.

Abyss777:
I've "repushed" branch
`this.callParent()` in `init` and license header is only changes, other things just split.

tananaev:
Merged, thanks.

PR #341, opened 2016-11-28
Fix warning about IMG alt

Abyss777:
There was warning in console
````
[W] For WAI-ARIA compliance, IMG elements SHOULD have an alt attribute.
````

### DIFF
--- a/web/l10n/en.json
+++ b/web/l10n/en.json
@@ -58,7 +58,7 @@
     "loginFailed": "Incorrect email address or password",
     "loginCreated": "New user has been registered",
     "loginLogout": "Logout",
-    "loginLogo":"Logo",
+    "loginLogo": "Logo",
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by tananaev** [comment]
I think it's better to rename it to "Logo" or something. Also, it should probably be in the strings file.

**Comment by Abyss777** [comment]
Done.

File: web/l10n/en.json [hunk 1]
```diff
@@ -58,6 +58,7 @@
     "loginFailed": "Incorrect email address or password",
     "loginCreated": "New user has been registered",
     "loginLogout": "Logout",
+    "loginLogo":"Logo",
```
tananaev: Space is missing :-)

Abyss777:
Sorry, why I'm so inattentive sometime...

tananaev:
Thanks for pull request, merged.

PR #342, opened 2016-11-29
Improve Notifications dialog and controller

Abyss777:
There are two reasons:

1. Redo it in simple and more correct way
2. Make it work on ExtJS 6.2.0

### DIFF
--- a/web/app/model/Notification.js
+++ b/web/app/model/Notification.js
@@ -30,5 +30,13 @@ Ext.define('Traccar.model.Notification', {
         type: 'int'
     }, {
         name: 'attributes'
+    }, {
+        name: 'attributes.web',
+        type: 'bool',
+        mapping: 'attributes.web'
+    }, {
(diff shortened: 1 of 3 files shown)


tananaev:
Is it that easy? Looks great, thanks.

Abyss777:
Yes. I wrote it then was not experienced in ExtJS. It was too complicated.
`mapping` simplified everything and allowed to remove confusing 6.2.0 code.

tananaev:
Are there any other places that we can simplify using mapping?

Abyss777:
Not in that way. We use "known" attributes nowhere at the moment except notifications. Also want to note that mapping is one direction. Changing mapping 'attributes.web` does not change 'attributes[web]'. 

But I know one place where we can use it:
Add columns to route report grid with some common attributes.

tananaev:
For example distance. Anyway, thanks for pull request again.

PR #345, opened 2016-11-30
Show events on map

Abyss777:
Key changes:

- Request positions to "Positions" store and add them to "ReportRoute" then. I think it is more natural to request positions from `api/positions` not from `api/report`
- Split `loadReport` on two functions. One draws route line, other draws markers. The only thing we iterate `data` twice in case routes.
- Handle bidirectional selecting. Select marker on map highlight event in grid and backward.

- in `loadEvents` function I used object instead of array to remove duplicate ids. It is not very frequent case, but there can be few events connected to one position. Just not to request them twice.

image

### DIFF
--- a/web/app/view/MapMarkerController.js
+++ b/web/app/view/MapMarkerController.js
@@ -202,24 +202,26 @@ Ext.define('Traccar.view.MapMarkerController', {
 
         this.addReportMarkers(store, data);
 
-        this.reportRoute = [];
-        for (i = 0; i < data.length; i++) {
-            position = data[i];
-            point = ol.proj.fromLonLat([
-                position.get('longitude'),
-                position.get('latitude')
-            ]);
-            if (i === 0 || data[i].get('deviceId') !== data[i - 1].get('deviceId')) {
-                this.reportRoute.push(new ol.Feature({
-                    geometry: new ol.geom.LineString([])
-                }));
-                this.reportRoute[this.reportRoute.length - 1].setStyle(this.getRouteStyle(data[i].get('deviceId')));
-                this.getView().getRouteSource().addFeature(this.reportRoute[this.reportRoute.length - 1]);
+        if (data.length > 0) {
+            this.reportRoute = [];
+            for (i = 0; i < data.length; i++) {
+                position = data[i];
+                point = ol.proj.fromLonLat([
+                    position.get('longitude'),
+                    position.get('latitude')
+                ]);
+                if (i === 0 || data[i].get('deviceId') !== data[i - 1].get('deviceId')) {
+                    this.reportRoute.push(new ol.Feature({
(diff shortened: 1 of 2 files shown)


web/app/view/ReportController.js
```diff
@@ -170,6 +184,41 @@ Ext.define('Traccar.view.ReportController', {
...
+            if (position) {
+                this.fireEvent('selectreport', position, true);
+            }
+        }
+    },
+
+    loadEvents: function (store, data) {
+        var i, eventObject, positionIds = {};
+        Ext.getStore('ReportRoute').removeAll();
+        if (data.length > 0) {
```
tananaev: I don't think check is required.

web/app/view/ReportController.js
```diff
@@ -170,6 +184,41 @@ Ext.define('Traccar.view.ReportController', {
...
+            for (i = 0; i < data.length; i++) {
+                eventObject = data[i];
+                if (eventObject.get('positionId')) {
+                    positionIds[eventObject.get('positionId')] = true;
```
tananaev: Can't we just use array instead? I don't expect there to be too many duplicates, so we can just de-dupe on the server side if really needed.

Abyss777:
Removed check and used array.

web/app/view/MapMarkerController.js
```diff
@@ -197,47 +198,52 @@ Ext.define('Traccar.view.MapMarkerController', {
...
+                this.reportRoute.push(new ol.Feature({
+                    geometry: new ol.geom.LineString([])
+                }));
+                this.reportRoute[this.reportRoute.length - 1].setStyle(this.getRouteStyle(data[i].get('deviceId')));
+                this.getView().getRouteSource().addFeature(this.reportRoute[this.reportRoute.length - 1]);
             }
+            this.reportRoute[this.reportRoute.length - 1].getGeometry().appendCoordinate(point);
+        }
+
+        this.getView().getMapView().fit(this.reportRoute[0].getGeometry(), this.getView().getMap().getSize());
```
tananaev: We actually need check here. I remember we had problems with empty list. I guess just for this line.

web/app/view/ReportController.js
```diff
@@ -170,6 +184,41 @@ Ext.define('Traccar.view.ReportController', {
...
+            if (position) {
+                this.fireEvent('selectreport', position, true);
+            }
+        }
+    },
+
+    loadEvents: function (store, data) {
+        var i, eventObject, positionIds = [];
+        Ext.getStore('ReportRoute').removeAll();
+        if (data.length > 0) {
```
tananaev: Another place where we don't need check.

Abyss777:
You are right. It is needed there. Reverted it and removed another.

tananaev:
Looks good now, thanks.

PR #346, opened 2016-11-30
Add Test Mail button to Account dialog.

Abyss777:
Added button to Account dialog only.
It may confuse admins if enable for all users dialog.
They may think it sends test email to other user.

image


web/app/view/SettingsMenuController.js
```diff
@@ -53,6 +53,7 @@ Ext.define('Traccar.view.SettingsMenuController', {
     onUserClick: function () {
         var dialog = Ext.create('Traccar.view.UserDialog');
         dialog.down('form').loadRecord(Traccar.app.getUser());
+        dialog.lookupReference('testMailButton').setHidden(false);
```
Abyss777: If admin will edit other user and see this button, he may think that it sends mail to other user. But it sends mail only him. It may confuse.

tananaev:
Merged, thanks.

PR #347, opened 2016-12-01
Handle eventId url parameter

Abyss777:
Also fit map view to all markers not only first route.

The logic is not obvious because of asynchronous. I'll try to explain.

- `Application` add function to remove parameter from url and global variable.
- `ReportController.init` check if url has `eventId` parameter and initiate event load process.
Remove `eventId` from url
- `MapMarkerController.loadReport` removed map view fit
- `MapMarkerController.addReportMarkers` calculate `ol.Extent` to fit map view.
Fire 'selectreport' event in case we come here with `eventId` parameter

After this changes if we pass parameter, event will be loaded, position will be loaded and showed on map.

I was thinking, if event does not have linked position, user won't understand what's happen.

- `ReportController.loadEvents` open report panel and highlight event in case we do not have linked position.

### DIFF
--- a/web/app/controller/Root.js
+++ b/web/app/controller/Root.js
@@ -97,7 +97,7 @@ Ext.define('Traccar.controller.Root', {
         eventId = Ext.Object.fromQueryString(window.location.search).eventId;
         if (eventId) {
             this.fireEvent('showsingleevent', eventId);
-            //this.removeUrlParameter('eventId');
+            this.removeUrlParameter('eventId');
         }
     },
 
--- a/web/app/controller/Root.js
+++ b/web/app/controller/Root.js
@@ -96,7 +96,7 @@ Ext.define('Traccar.controller.Root', {
         }
         eventId = Ext.Object.fromQueryString(window.location.search).eventId;
         if (eventId) {
-            this.application.fireEvent('showSingleEvent', eventId);
+            this.fireEvent('showsingleevent', eventId);
             //this.removeUrlParameter('eventId');
         }
     },
--- a/web/app/controller/Root.js
+++ b/web/app/controller/Root.js
@@ -75,7 +75,7 @@ Ext.define('Traccar.controller.Root', {
     },
 
     loadApp: function () {
-        var attribution;
+        var attribution, eventId;
         Ext.getStore('Groups').load();
         Ext.getStore('Geofences').load();
         Ext.getStore('AttributeAliases').load();
@@ -94,6 +94,11 @@ Ext.define('Traccar.controller.Root', {
         } else {
             Ext.create('widget.main');
         }
+        eventId = Ext.Object.fromQueryString(window.location.search).eventId;
+        if (eventId) {
+            this.application.fireEvent('showSingleEvent', eventId);
+            //this.removeUrlParameter('eventId');
+        }
     },
 
     beep: function () {
@@ -106,6 +111,12 @@ Ext.define('Traccar.controller.Root', {
     mutePressed: function () {
         var muteButton = Ext.getCmp('muteButton');
         return muteButton && !muteButton.pressed;
+    },
+
+    removeUrlParameter: function (param) {
+        var params = Ext.Object.fromQueryString(window.location.search);
+        delete params[param];
+        window.history.pushState(null, null, window.location.pathname + '?' + Ext.Object.toQueryString(params));
     },
 
     asyncUpdate: function (first) {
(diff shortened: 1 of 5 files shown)


tananaev:
Is this intended for email links to show events?

Abyss777:
Yes. It is the main reason.

tananaev:
The logic is very weird. I don't like that we use global variable and then manipulate URL from report controller. There must be a better way.

Abyss777:
We can do it without manipulating url. It is just for user comfort, not load event again on page refresh.

tananaev:
I'm not against URL manipulation, but it should be in some more appropriate place. And we should pass parameters around instead of using global variables as much as possible.

Abyss777:
I agree with you but everything is asynchronous. Callbacks is everywhere.
Another problem is that everything is ready in different time. Some events are missed by some components on the load stage.

But I will try again.

tananaev:
Let me know if you need any help from me.

Abyss777:
is Root.js more appropriate place to handle url parameters and url manipulation?

tananaev:
Yes, I think so. Then you can pass the parameter down to map and/or report
somehow.

Abyss777:
Implemented it in other way.

fire `showsinlgleevent` in Root.js
catch it in `ReportController` and initiate load process
in callback select row in report and expand if needed

I need to temporary store variable in `ReportController` to show position in State panel

web/app/view/MapMarkerController.js
```diff
@@ -202,37 +202,37 @@ Ext.define('Traccar.view.MapMarkerController', {
...
-        var i, position, point, geometry, marker, style;
+        var i, position, point, geometry, marker, style, minx, miny, maxx, maxy;
         this.clearReport();
         for (i = 0; i < data.length; i++) {
             position = data[i];
             point = ol.proj.fromLonLat([
                 position.get('longitude'),
                 position.get('latitude')
             ]);
+            minx = minx ? Math.min(point[0], minx) : point[0];
```
tananaev: Why do we need this logic?

web/app/view/ReportController.js
```diff
@@ -208,15 +210,45 @@ Ext.define('Traccar.view.ReportController', {
                 params: {
                     id: positionIds
                 },
+                scope: this,
                 callback: function (records, operation, success) {
                     if (success) {
                         Ext.getStore('ReportRoute').add(records);
+                        if (this.singleEvent && records.length > 0) {
```
tananaev: I'm think maybe it's better to do it all reports anyway? Then we don't need `singleEvent` variable and everything is going to be simpler. What do you think?

web/app/controller/Root.js
```diff
@@ -94,6 +94,11 @@ Ext.define('Traccar.controller.Root', {
         } else {
             Ext.create('widget.main');
         }
+        eventId = Ext.Object.fromQueryString(window.location.search).eventId;
+        if (eventId) {
+            this.fireEvent('showsingleevent', eventId);
+            //this.removeUrlParameter('eventId');
```
tananaev: Why is this commented out?

web/app/view/MapMarkerController.js
Abyss777: The only reason is `Math.min(point[0], undefined)` gives `NaN`.
I will redo it with less compares.

web/app/view/ReportController.js
Abyss777: I was thinking about it too. But may be do it for reports with one position only?

tananaev: I think we can just select first item for any report. Even if it's now just one element. Kind of makes sense.

web/app/view/MapMarkerController.js
tananaev: I mean why can't we just rely on `this.getView().getMapView().fit`?

Abyss777:
Did the changes.

web/app/view/MapMarkerController.js
Abyss777: Fit accept geometry or rectangle. We have a lot of geometries and have to find such rectangle that contain all geometries.

web/app/view/ReportController.js
Abyss777: If we select first position for every event report, then we have to select event in report grid for consistent.
Also if we select everything (in three panels) it should repeat behavior of manual selection and move map to selected position. It is not good, it should fit all positions after  generating report.

tananaev: OK, then your option of auto-selecting if a single sounds better.

tananaev:
Merged, thanks.

PR #349, opened 2016-12-02
Fix notifications

Abyss777:
It was miracle coincidence, It looked like working, but shouldn't.

- Used `type` as unique id field
- ` loadData` does not  respect IDs, but `add` does. It just replace records with same id.
- Added `refresh` because it sometimes updates grid with delay or only after scrolling/resizing.

ExtJS does everything and now it works.

Sorry, that missed it before.

### DIFF
--- a/web/app/model/Notification.js
+++ b/web/app/model/Notification.js
@@ -17,7 +17,7 @@
 
 Ext.define('Traccar.model.Notification', {
     extend: 'Ext.data.Model',
-    identifier: 'negative',
+    idProperty: 'type',
 
     fields: [{
         name: 'id',
@@ -31,12 +31,10 @@ Ext.define('Traccar.model.Notification', {
     }, {
         name: 'attributes'
     }, {
-        name: 'attributes.web',
-        type: 'bool',
-        mapping: 'attributes.web'
+        name: 'web',
+        type: 'bool'
     }, {
-        name: 'attributes.mail',
-        type: 'bool',
-        mapping: 'attributes.mail'
+        name: 'mail',
+        type: 'bool'
     }]
 });
--- a/web/app/store/Notifications.js
+++ b/web/app/store/Notifications.js
@@ -22,5 +22,7 @@ Ext.define('Traccar.store.Notifications', {
     proxy: {
         type: 'rest',
         url: 'api/users/notifications'
-    }
+    },
+    sortOnLoad: true,
+    sorters: { property: 'type', direction : 'ASC' }
 });
--- a/web/app/view/Notifications.js
+++ b/web/app/view/Notifications.js
@@ -24,7 +24,7 @@ Ext.define('Traccar.view.Notifications', {
     ],
 
     controller: 'notificationsController',
-    store: 'AllNotifications',
+    store: 'Notifications',
 
     selModel: {
         selType: 'cellmodel'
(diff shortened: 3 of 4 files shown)


tananaev:
Not really related to the changed, but I just looked at the code and I couldn't understand why we need to load data in the store twice here:
```javascript
        this.getView().getStore().load({
            scope: this,
            callback: function (records, operation, success) {
                Ext.create('Traccar.store.Notifications').load({
                    params: {
                        userId: this.userId
                    },
                    scope: this,
                    callback: function (records, operation, success) {
                        if (success) {
                            this.getView().getStore().loadData(records);
                        }
                    }
                });
            }
        });
```

Second question is in the following code. Why don't we use values directly and why do we use "true" as a string and not actual boolean value?
```javascript
        if (record.get('attributes.web')) {
            attributes.web = 'true';
        }
        if (record.get('attributes.mail')) {
            attributes.mail = 'true';
        }
```

Also, why is `onCheckChange` so complicated? Can't we just save whole model now that we use attributes directly?

Abyss777:
The view store is `AllNotifications` it is all possible event types with empty attributes.
The second request is actual user notifications.

As I said before, the "mapping" work in one direction, moreover only on load stage. I mean:
Changing mappings (manually or programmatically) does not change attributes
Changing attributes via `set('attributes', a)` does not change mappings.

We can not just post whole objects because mappings and attributes not consistent in `onCheckChange` function.

About boolean... They comes from server in quotation marks and I decided send they back too.

tananaev:
I'm thinking maybe we should just move those values into the model from attributes. It will simplify the code.

As for 2 calls, maybe we should just change the call to return all notifications for a user including missing ones? Basically do the merge on the server. That way we can reduce the number of calls and we don't really use them separately anyway, right?

Abyss777:
Yes, we can merge them on server side. Or just use static store on web side.
Also there is no much benefits of generating all notifications list on server side, anyway we have to add Strings for new types.

If we do not have plans to support new notification transport type, moving it to model will simplify the code on web side, and will not complicate it on server side I think.

Abyss777:
The only thing, how convert existed objects?

tananaev:
If we want to add more, we can just add new columns.

As for migration, it is a problem, but I think we can add temporary code to extract data from attributes if they are present. Or we can do something in database changelogs. Second is probably a better.

Abyss777:
Not sure we will be able to do universal changlog using db json engines:
MySQL - `UPDATE notifications SET web = true WHERE attributes->'$.web' = 'true'`
PG - `UPDATE notifications SET web = true WHERE attributes->'web' = 'true'`
MSSQL - `UPDATE notifications SET web = true WHERE JSON_VALUE(attributes, '$.web') = 'true'`

And It is theoretically, not tested, may be columns must be json type.

Another option, more universal I think, to treat them as string
`UPDATE notifications SET web = true WHERE attributes LIKE  '%"web":"true"%'`

tananaev:
I think we know all the possible combinations, we can just use full text comparisons:
```
{web: true, email: true}
{email: true}
{web: true}
```
or something like that.

Abyss777:
You are right. Will we clean attributes after that to `{}` ?

tananaev:
Makes sense to clear them to not have duplicate information.

Abyss777:
Replaced changes here to support https://github.com/tananaev/traccar/pull/2638

Abyss777:
Hello,
Any comments for this changes?

tananaev:
Sorry, I completely forgot about this pull request.

PR #352, opened 2016-12-08
Fix parameter removing if it is single

Abyss777:
There is unnecessary `?` at the end if parameter is single


tananaev:
Merged, thanks.

PR #355, opened 2016-12-12
Align Toasts to bottom right corner

### LINKED ISSUE #354: Toasts in the middle of the screen
Looks like we missed ExtJS 6.2.0 incompatibility
image

Abyss777:
I think it is more "nice-looking" to have Toasts in bottom right corner than in the middle of the screen.
Also it is more usual.
fix #354


tananaev:
Merged, thanks.

PR #358, opened 2016-12-14
Calendars implementation

Abyss777:
Here is web part of https://github.com/tananaev/traccar/pull/2686

Nothing special, routine except loading file to `calendarData`

It is loaded as soon as user select file and converted to base64.
Jersey send it as base64, but accept if we pass it as byte array like `"calendarData":[66, 88, ...]`

image

image

image

image

### DIFF
--- a/web/app/view/CalendarDialog.js
+++ b/web/app/view/CalendarDialog.js
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2016 Anton Tananaev (anton@traccar.org)
+ * Copyright 2016 Andrey Kunitsyn (andrey@traccar.org)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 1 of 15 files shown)


web/app/view/CalendarDialog.js
```diff
@@ -0,0 +1,58 @@
...
+            fieldLabel: Strings.sharedName,
+            allowBlank: false
+        }, {
+            xtype: 'filefield',
+            name: 'file',
+            fieldLabel: Strings.calendarFile,
+            allowBlank: false,
+            buttonConfig: {
+                glyph: 'xf093@FontAwesome',
+                text: "",
```
Abyss777: It is needed, otherwise we get button with "Browse..." caption.

Abyss777: Sorry, sometimes hard to switch brain from java to js.

Abyss777:
Done changes.

It accepts any `.ics` files, most calendar applications have functionality to export to `.ics`. I used Mozilla Thunderbird to generate test calendars.

web/l10n/en.json
```diff
@@ -205,5 +207,7 @@
     "categoryPlane": "Plane",
     "categoryMotorcycle": "Motorcycle",
     "categoryBicycle": "Bicycle",
-    "categoryPerson": "Person"
+    "categoryPerson": "Person",
+    "calendarFile": "Calendar File",
```
tananaev: I'm thinking that maybe we should make these two string more common. Something like "File" and "Select File". It would reduce number of string if we decide to use it for something else in future.

Abyss777:
Made strings more common

Only shadow of "Select File" tooltip was captured...
image

tananaev:
Merged, thanks.

PR #362, opened 2016-12-20
UI for Feature/traccar 2702 watch ext

amenk:
https://github.com/tananaev/traccar/issues/2702


web/l10n/de.json
```diff
@@ -144,6 +144,8 @@
     "commandData": "Daten",
     "commandPhone": "Rufnummer",
     "commandMessage": "Nachricht",
+    "commandTimezoneOffset": "Zeitverschiebung zu UTC",
```
tananaev: You shouldn't change any of the localization files directly. All localization should be done via Transifex website.

amenk: okay ... also the english one? But in en.json it's fine, right? Shall I rebase or just revert the file?

tananaev: The way it works is that you only add/modify English strings, then Transifex automatically picks up the file directly from the repository and people can translate it. Then with every new release I update localization files using special script.

You can just revert it.

amenk: done

tananaev:
If we rename "hearts" to "indicator", we should rename it here as well.

amenk:
yep, done

web/l10n/en.json
```diff
@@ -144,6 +144,8 @@
     "commandData": "Data",
     "commandPhone": "Phone Number",
     "commandMessage": "Message",
+    "commandTimezoneOffset": "Timezone Offset to UTC",
```
tananaev: Command key should match the command key for consistency. Also, I think we can shorten the value to "Timezone Offset".

web/l10n/en.json
```diff
@@ -144,6 +144,8 @@
...
+    "commandSetIndicator": "Set Indicator",
```
tananaev: Please move this line next to other command names.

web/app/view/CommandDialog.js
```diff
@@ -90,6 +90,33 @@ Ext.define('Traccar.view.CommandDialog', {
...
+            xtype: 'fieldcontainer',
+            reference: 'paramSetTimezone',
+            name: 'attributes',
+            hidden: true,
+
+            items: [{
+                xtype: 'numberfield',
+                fieldLabel: Strings.commandTimezoneOffset,
+                name: 'timezone',
+                minValue: -12,
```
tananaev: I believe the value is in seconds. I guess you would have to change server side if it doesn't accept seconds.

amenk: in the watch protocol it is in hours ...

tananaev: API expects seconds, so it should be converted accordingly in the encoder.

web/l10n/en.json
amenk: ok

web/l10n/en.json
tananaev: The key in the command is "timezone", so it should be "commandTimezone".

tananaev:
This is minutes as well, not seconds.

tananaev:
Merged, thanks.

PR #363, opened 2016-12-21
locale by browser preferences

guterresrafael:
Hi,
I was having problems with "pt-BR" language in the Chrome and fix the definition of locale.
Added support to array "window.navigator.languages" to use browser preferences.


tananaev:
I think we can combine two ifs and remove some redundant cases. Otherwise looks good.

guterresrafael:
Done! Now I use the array for define language.

tananaev:
Merged, thanks.

PR #364, opened 2016-12-21
Migrate to openlayers 3.20.0

Abyss777:
Key changes
- Followed migration notes https://github.com/openlayers/ol3/releases/tag/v3.20.0
- Used style.setImage() new function
- Style fixes in `Root.js`

fix for https://www.traccar.org/forums/topic/device-course-wrong-on-map/

### DIFF
--- a/web/app/GeofenceConverter.js
+++ b/web/app/GeofenceConverter.js
@@ -43,7 +43,7 @@ Ext.define('Traccar.GeofenceConverter', {
                     projection = mapView.getProjection();
                     center = ol.proj.transform([Number(coordinates[1]), Number(coordinates[0])], 'EPSG:4326', projection);
                     resolutionAtEquator = mapView.getResolution();
-                    pointResolution = projection.getPointResolution(resolutionAtEquator, center);
+                    pointResolution = ol.proj.getPointResolution(projection, resolutionAtEquator, center);
                     resolutionFactor = resolutionAtEquator / pointResolution;
                     radius = (Number(coordinates[2]) / ol.proj.METERS_PER_UNIT.m) * resolutionFactor;
                     geometry = new ol.geom.Circle(center, radius);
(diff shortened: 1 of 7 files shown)


web/app/view/BaseMap.js
```diff
@@ -122,7 +122,7 @@ Ext.define('Traccar.view.BaseMap', {
         this.map.on('click', function (e) {
             this.map.forEachFeatureAtPixel(e.pixel, function (feature, layer) {
                 this.fireEvent('selectfeature', feature);
-            }, this);
+            }.bind(this));
```
tananaev: Is the old version no longer available?

web/load.js
```diff
@@ -111,7 +111,7 @@
 
     extjsVersion = '6.2.0';
     fontAwesomeVersion = '4.7.0';
-    olVersion = '3.19.1';
+    olVersion = '3.20.0';
```
tananaev: Please update it in the "simple" folder as well. By the way, I don't remember if I showed it to you. It's basically a simple web interface (just map) without any extra dependencies.

web/app/GeofenceConverter.js
```diff
@@ -43,7 +43,7 @@ Ext.define('Traccar.GeofenceConverter', {
                     projection = mapView.getProjection();
                     center = ol.proj.transform([Number(coordinates[1]), Number(coordinates[0])], 'EPSG:4326', projection);
                     resolutionAtEquator = mapView.getResolution();
-                    pointResolution = projection.getPointResolution(resolutionAtEquator, center);
+                    pointResolution = ol.proj.getPointResolution(projection, resolutionAtEquator, center);
```
tananaev: Is it another API change as well?

web/app/view/BaseMap.js
Abyss777: Unfortunately yes, it is in release upgrade notes
> ol.Map#forEachFeatureAtPixel and ol.Map#hasFeatureAtPixel parameters have changed

web/app/GeofenceConverter.js
Abyss777: Yes, also in upgrade notes
> Use ol.proj.getPointResolution() instead of projection.getPointResolution()

web/load.js
Abyss777: Updated.
I've seen it. So desirable feature and example for API. 
Need to share `DeviceImajes.js` library somehow between "simple" and full

tananaev:
Merged, thanks.

PR #365, opened 2016-12-21
Save toggles state in user/server attributes

### LINKED ISSUE #303: Store visual settings in cookies
Idea for improvement:
There are some settings that is not worth  to be stored in DB, but still good to be restored after browser reopen.
- State of "mute" button
- State of "Show geofences" button
- Set of devices columns (especially after extending device model)

Store them in cookies may be a good idea

### LINKED ISSUE #338: Save toggle button status
I remember this was discussed before somewhere, but I'll just add it anyway so we don't forget.

It would be nice to save toggle button statuses. Got a request for this feature from one client.

Not a huge priority, but we can consider it at some point.

I guess the best option would be to save it in attributes.

Abyss777:
fix #338 and fix #303 

I treat attributes as string because if we add them manually they will be string.
Also boolean attributes will cause wrong behavior of `getAttributePreference` function. 

Server dialog:
image

User dialog:
image

### DIFF
--- a/web/app/view/Map.js
+++ b/web/app/view/Map.js
@@ -48,7 +48,6 @@ Ext.define('Traccar.view.Map', {
             reference: 'showGeofencesButton',
             glyph: 'xf21d@FontAwesome',
             enableToggle: true,
-            pressed: true,
             tooltip: Strings.sharedGeofences
         }, {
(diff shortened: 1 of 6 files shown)


tananaev:
Merged, thanks.

PR #367, opened 2016-12-22
Add support for custom report colors

Abyss777:
Can't find where, but remember somebody asked it.
Used device attribute `web.reportColor`. Accept any html color without validation, if it is not valid, line will be black and arrows green. (black is default for openlayers, green is in our svg files)

image

Also fixed style issues in `load.js`


tananaev:
Looks good, merged.

PR #368, opened 2016-12-22
Add "Animal" device category

Abyss777:
Added new category as you suggested here https://www.traccar.org/forums/topic/additional-icons-and-current-location/

image


tananaev:
Looks great, merged.

I was wondering if we can combine all images in a single XML file. Maybe it will reduce the download time and increase speed? Do you think it makes sense?

Abyss777:
Not sure if it gives any benefits, they are mostly loaded in parallel:
image

But may be we should research it.

tananaev:
Maybe we can just keep it in mind if we have any performance issues at some point in future.

Abyss777:
Sure, it is not hard to combine it.

PR #371, opened 2016-12-26
Charts implementation

Abyss777:
Key changes:

- use two hidden panels: grid and chart, and show them then needed
- moved `getReportColor` to use from map and report
- have not found better solution for calculated fields than `convertValue` in `SpeedUnits` and `DistanceUnits`
- Distance chart is absolutely meaningless, added just for example and polish switching chart types.
Other attributes can be added just as mapping if they do not need conversion.

image

image

Selecting point on chart will cause selecting position on map.

Known issues:
- nothing works under chart in my mobile chrome. Hold-tap does not shows context menu, any swipes also do not work. Do not know who to blame.


web/app/view/Report.js
```diff
@@ -67,17 +67,50 @@ Ext.define('Traccar.view.Report', {
...
+        forceFit: true,
+        flex: 1,
+        columns: {
+            defaults: {
+                minWidth: Traccar.Style.columnWidthNormal
+            },
+            items: [
+            ]
+        },
+        style: 'borderTop: 1px solid lightgray'
```
Abyss777: Do not like it, but...
image

There is no border under grid header.

web/app/view/ReportConfigDialog.js
```diff
@@ -56,6 +56,17 @@ Ext.define('Traccar.view.ReportConfigDialog', {
         displayField: 'name',
         queryMode: 'local'
     }, {
+        fieldLabel: Strings.reportChartTypes,
+        xtype: 'combobox',
+        width: Traccar.Style.reportTagfieldWidth,
+        reference: 'chartTypeField',
+        store: 'ReportChartTypes',
+        hidden: true,
+        value: 'speedConverted',
```
Abyss777: Default chart type is "Speed"

web/app/view/ReportController.js
```diff
@@ -52,9 +55,15 @@ Ext.define('Traccar.view.ReportController', {
         Traccar.app.showReports(false);
     },
 
+    init: function () {
+        this.grid = this.getView().getComponent('grid');
+        this.chart = this.getView().getComponent('chart');
```
Abyss777: Added just to simpler reference

web/app/store/DistanceUnits.js
```diff
@@ -33,6 +33,15 @@ Ext.define('Traccar.store.DistanceUnits', {
...
+    convertValue: function (value, unit) {
+        var model;
+        if (!unit) {
+            unit = 'km';
+        }
+        model = this.findRecord('key', unit);
+        return value * model.get('factor');
+    },
+
     formatValue: function (value, unit) {
```
tananaev: I think we should re-use  `convertValue` logic in `formatValue`.

web/app/store/SpeedUnits.js
```diff
@@ -33,6 +33,15 @@ Ext.define('Traccar.store.SpeedUnits', {
...
+    convertValue: function (value, unit) {
+        var model;
+        if (!unit) {
+            unit = 'kn';
+        }
+        model = this.findRecord('key', unit);
+        return value * model.get('factor');
+    },
+
     formatValue: function (value, unit) {
```
tananaev: Same, re-use  `convertValue` logic in `formatValue`.

web/app/view/Report.js
```diff
@@ -67,17 +67,50 @@ Ext.define('Traccar.view.Report', {
         }]
     },
 
-    listeners: {
-        selectionchange: 'onSelectionChange'
+    layout: {
+        type: 'fit'
```
tananaev: I think `card` layout is more appropriate. Also, you don't need curly braces, you can just use string straight away.

web/app/view/Report.js
```diff
@@ -67,17 +67,50 @@ Ext.define('Traccar.view.Report', {
...
-            minWidth: Traccar.Style.columnWidthNormal
+    items: [{
+        xtype: 'grid',
+        itemId: 'grid',
+        listeners: {
+            selectionchange: 'onSelectionChange'
+        },
+        hidden: true,
+        forceFit: true,
+        flex: 1,
```
tananaev: Why do we need flex?

web/app/view/Report.js
tananaev: Check if card layout fixes this.

web/app/view/ReportController.js
```diff
@@ -52,9 +55,15 @@ Ext.define('Traccar.view.ReportController', {
         Traccar.app.showReports(false);
     },
 
+    init: function () {
+        this.grid = this.getView().getComponent('grid');
```
tananaev: Maybe we should just add two methods `getGrid` and `getChart`. I think it would look better than instance variables.

web/app/store/DistanceUnits.js
Abyss777: I was thinking about it too, but decided that `findRecord` costs a lot to call it twice.

web/app/view/Report.js
Abyss777: Unfortunately not. Line is there until select some type. 

image

Abyss777:
You are right. Card layout is more appropriate. Type switching is looks much much better now.

web/app/store/DistanceUnits.js
Abyss777: On the other hand there are only three items to lookup... not big overhead

Abyss777:
Reused `convertValue` logic in `formatValue`

Abyss777:
Forget to say, that I've updated openlayers to 3.20.1.
So soon release looks like some significant regression https://github.com/openlayers/ol3/releases/tag/v3.20.1

web/app/Style.js
```diff
@@ -78,5 +78,14 @@ Ext.define('Traccar.Style', {
     coordinatePrecision: 6,
     numberPrecision: 2,
 
-    reportTagfieldWidth: 375
+    reportTagfieldWidth: 375,
+
+    chartPadding: {
```
tananaev: Does it accept a string like '20 10 40'? Or only js object like this?

web/app/view/Report.js
tananaev: Let's at least move it to styles class.

web/app/Style.js
Abyss777: Unfortunately not. It accepts string http://docs.sencha.com/extjs/6.2.0/classic/Ext.chart.CartesianChart.html#cfg-insetPadding
But takes only first number for all dimensions (bug or feature)

Abyss777:
Moved grid style to `Style.js`

tananaev:
It uses `Ext.util.Format.parseBox`, so you should be able to specify a string.

Abyss777:
I'm very sorry, you are right. Looks like I missed order before

tananaev:
Merged, thanks.

By the way, very useful feature in ExtJS documentation is "view source" button where you can see what a function actually does and possible inputs.

Abyss777:
Thanks for the tip. I'll use it.

Abyss777:
Just one more picture, how chart looks with a device that sends positions frequently:

image

tananaev:
Excellent. One other thing I noticed. I think the type of the report should be called "Chart" (single) because it shows only one chart.

Abyss777:
I think it is reasonable.
image
I take "Event types" as example for "Chart types", but now thinking... we can select multiple event types and only one chart type, may be also rename it to single "Chart type" ?

tananaev:
Yes, it makes sense to rename that to single too... unless we allow multiple selection.

PR #372, opened 2016-12-27
Use singular for a few strings

Abyss777:
- "Chart" instead of "Charts"
- "Chart Type" instead of "Chart Types"


tananaev:
Merged, thanks.

PR #376, opened 2016-12-29
Add support for geofence colors

Abyss777:
Added "Attributes" button to geofence edit dialog

image

Take color value from geofence attributes (as usual any valid html color)

image

Result

image

Used simple trick for getting fill color, convert color to array and set alpha channel.

### DIFF
--- a/web/app/view/MapController.js
+++ b/web/app/view/MapController.js
@@ -86,13 +86,22 @@ Ext.define('Traccar.view.MapController', {
         this.fireEvent('togglestate', state);
     },
 
-    getGeofenceStyle: function (label) {
+    getGeofenceStyle: function (label, color) {
+        var fillColor, strokeColor;
+        if (color) {
+            fillColor = ol.color.asArray(color);
+            strokeColor = color;
+        } else {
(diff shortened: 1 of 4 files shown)


web/app/view/MapController.js
```diff
@@ -114,7 +123,8 @@ Ext.define('Traccar.view.MapController', {
             Ext.getStore('Geofences').each(function (geofence) {
                 var feature = new ol.Feature(Traccar.GeofenceConverter
                         .wktToGeometry(this.getView().getMapView(), geofence.get('area')));
-                feature.setStyle(this.getGeofenceStyle(geofence.get('name')));
+                feature.setStyle(this.getGeofenceStyle(geofence.get('name'),
+                        geofence.get('attributes') !== undefined ? geofence.get('attributes').color : undefined));
```
tananaev: You can remove `!== undefined` here. Also, it's not a good practice to use `undefined` as a value. You should be using `null`.

Abyss777:
Oh, yes of course. Done.

tananaev:
Merged, thanks.

PR #377, opened 2016-12-29
Bring markers to front and deselect on empty space click

### LINKED ISSUE #374: Bring selected icon to front
When we have many icons in one area, it would be nice to bring selected one to the front.

Another unrelated thing I was thinking about is to remove selection if you click outside of any icons on the map.

Abyss777:
fix #374 

- combine latest and report layers to one
- bring to front selected marker
- deselect marker (and do other related things) on click empty space

image

image

### DIFF
--- a/web/app/view/BaseMap.js
+++ b/web/app/view/BaseMap.js
@@ -120,9 +120,16 @@ Ext.define('Traccar.view.BaseMap', {
         });
 
         this.map.on('click', function (e) {
-            this.map.forEachFeatureAtPixel(e.pixel, function (feature, layer) {
-                this.fireEvent('selectfeature', feature);
-            }.bind(this));
+            if (this.map.hasFeatureAtPixel(e.pixel, {
+                layerFilter: function (layer) {
+                    return layer.get('name') !== 'geofencesLayer';
+                }})) {
+                this.map.forEachFeatureAtPixel(e.pixel, function (feature, layer) {
+                    this.fireEvent('selectfeature', feature);
+                }.bind(this));
+            } else {
+                this.fireEvent('deselectfeature');
+            }
         }, this);
     },
 
--- a/web/app/view/BaseMap.js
+++ b/web/app/view/BaseMap.js
@@ -123,7 +123,8 @@ Ext.define('Traccar.view.BaseMap', {
             if (this.map.hasFeatureAtPixel(e.pixel, {
                 layerFilter: function (layer) {
                     return layer.get('name') !== 'geofencesLayer';
-                }})) {
+                }
+            })) {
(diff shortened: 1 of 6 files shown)


web/app/view/BaseMap.js
```diff
@@ -120,9 +120,16 @@ Ext.define('Traccar.view.BaseMap', {
         });
 
         this.map.on('click', function (e) {
-            this.map.forEachFeatureAtPixel(e.pixel, function (feature, layer) {
-                this.fireEvent('selectfeature', feature);
-            }.bind(this));
+            if (this.map.hasFeatureAtPixel(e.pixel, {
+                layerFilter: function (layer) {
+                    return layer.get('name') !== 'geofencesLayer';
+                }})) {
```
Abyss777: Geofences can be rather big, consider them as empty space.
Also have some doubts in indentation and formatting in this place.

tananaev: I think you have some problems with formatting here.

web/app/view/MapMarkerController.js
```diff
@@ -346,11 +347,13 @@ Ext.define('Traccar.view.MapMarkerController', {
     selectMarker: function (marker, center) {
         if (this.selectedMarker) {
             this.resizeMarker(this.selectedMarker.getStyle(), false);
+            this.selectedMarker.getStyle().setZIndex(0);
             this.selectedMarker.changed();
         }
 
         if (marker) {
             this.resizeMarker(marker.getStyle(), true);
+            marker.getStyle().setZIndex(1);
```
tananaev: Why is it "1" here? I'm assuming lower index means higher layer, right?

Abyss777: I cant find the documentation explaining this, but if I use -1 marker become lower than others.

Abyss777:
I've changed formatting,  but previous was looking correct, at least `jscs` did not complain about it.

web/app/view/BaseMap.js
```diff
@@ -120,9 +120,15 @@ Ext.define('Traccar.view.BaseMap', {
         });
 
         this.map.on('click', function (e) {
-            this.map.forEachFeatureAtPixel(e.pixel, function (feature, layer) {
-                this.fireEvent('selectfeature', feature);
-            }.bind(this));
+            if (this.map.hasFeatureAtPixel(e.pixel, { layerFilter: function (layer) {
```
tananaev: It's still not correct. Everything after curly brace should be on a new line. Also there should be no 2 curly braces on the same line at the end. Split it into different lines.

Abyss777:
I've replaced last commit. It's pity that `jscs` can't analyze such complex formatting correctly.

tananaev:
Yes, unfortunately it can't. Thanks for pull request.

PR #378, opened 2016-12-30
Add an option to not show report markers

Abyss777:
Added option to report config (checked by default)

image

I decided to pass its state via store to map controller.

All report types are affected.

image

image

### DIFF
--- a/web/app/view/MapMarkerController.js
+++ b/web/app/view/MapMarkerController.js
@@ -240,15 +240,17 @@ Ext.define('Traccar.view.MapMarkerController', {
                 maxx = Math.max(point[0], maxx);
                 maxy = Math.max(point[1], maxy);
             }
-            geometry = new ol.geom.Point(point);
-            marker = new ol.Feature(geometry);
-            marker.set('record', position);
-            style = this.getReportMarker(position.get('deviceId'), position.get('course'));
-            /*style.getText().setText(
-                Ext.Date.format(position.get('fixTime'), Traccar.Style.dateTimeFormat24));*/
-            marker.setStyle(style);
-            this.reportMarkers[position.get('id')] = marker;
-            this.getView().getMarkersSource().addFeature(marker);
+            if (store.showMarkers !== false) {
+                geometry = new ol.geom.Point(point);
+                marker = new ol.Feature(geometry);
+                marker.set('record', position);
+                style = this.getReportMarker(position.get('deviceId'), position.get('course'));
+                /*style.getText().setText(
+                    Ext.Date.format(position.get('fixTime'), Traccar.Style.dateTimeFormat24));*/
+                marker.setStyle(style);
+                this.reportMarkers[position.get('id')] = marker;
+                this.getView().getMarkersSource().addFeature(marker);
+            }
         }
         if (minx !== maxx || miny !== maxy) {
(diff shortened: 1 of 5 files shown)


web/app/view/MapMarkerController.js
```diff
@@ -240,15 +240,17 @@ Ext.define('Traccar.view.MapMarkerController', {
...
-            geometry = new ol.geom.Point(point);
-            marker = new ol.Feature(geometry);
-            marker.set('record', position);
-            style = this.getReportMarker(position.get('deviceId'), position.get('course'));
-            /*style.getText().setText(
-                Ext.Date.format(position.get('fixTime'), Traccar.Style.dateTimeFormat24));*/
-            marker.setStyle(style);
-            this.reportMarkers[position.get('id')] = marker;
-            this.getView().getMarkersSource().addFeature(marker);
+            if (store.showMarkers !== false) {
```
tananaev: I think you can just check `if (store.showMarkers) {`.

web/app/view/MapMarkerController.js
```diff
@@ -240,15 +240,17 @@ Ext.define('Traccar.view.MapMarkerController', {
...
+                geometry = new ol.geom.Point(point);
+                marker = new ol.Feature(geometry);
+                marker.set('record', position);
+                style = this.getReportMarker(position.get('deviceId'), position.get('course'));
+                /*style.getText().setText(
```
tananaev: Please remove this comment. It's been here for a while and I don't think we would enable it any time soon.

web/app/view/ReportController.js
```diff
@@ -125,6 +128,7 @@ Ext.define('Traccar.view.ReportController', {
                 } else {
                     store = this.getGrid().getStore();
                 }
+                store.showMarkers = this.showMarkers;
```
tananaev: Not sure if it's the best option, but I guess it will do.

Abyss777: Agree with you. But I know only two ways of passing  values between controllers.
- `getCmp`
- events

Events do not fit, `getCmp` is very undesirable.

web/app/view/MapMarkerController.js
Abyss777: single event via parameter will not work in this case (`store.showMarkers` is not defined )

tananaev: I see. Agree then.

web/app/view/ReportController.js
tananaev: Let's leave it as it is for now.

Abyss777:
Removed comment

tananaev:
Merged, thanks. By the way, have you checked if it actually improves performance on large tracks?

Abyss777:
Yes, It is definitely improved the performance. 
Almost impossible move the map with markers with my 5 day route.
But without markers it moves and scrolls as empty without any lacks.

tananaev:
Great. Maybe we should make it a default option?

Abyss777:
Yes. I think we should make it, but ... we should do another thing before:
If markers is disabled, we should show at least one selected in grid/chart. Now only state is changed on select.

After that reports will look natural and understandable in both cases, with and without markers.

tananaev:
Agree.

PR #379, opened 2016-12-30
Highlight selected marker if they are disabled

Abyss777:
Highlight marker for selected position...

image

... and event

image

New logic allowed simplify check `if (store.showMarkers)`

Also do not show markers by default


tananaev:
Merged, thanks.

PR #383, opened 2017-01-10
Fix interface sliding on mobile

Abyss777:
The best way is `display:none` but in most browsers images are not loaded.

Just moved outside visible area.

### DIFF
--- a/web/load.js
+++ b/web/load.js
@@ -21,7 +21,7 @@
         svg.setAttribute('id', id);
         svg.setAttribute('data', file);
         svg.setAttribute('type', 'image/svg+xml');
-        svg.setAttribute('style', 'visibility:hidden');
+        svg.setAttribute('style', 'visibility:hidden;position:absolute;left:-100px;');
         document.body.appendChild(svg);
     }
 


web/load.js
```diff
@@ -21,7 +21,7 @@
         svg.setAttribute('id', id);
         svg.setAttribute('data', file);
         svg.setAttribute('type', 'image/svg+xml');
-        svg.setAttribute('style', 'visibility:hidden');
+        svg.setAttribute('style', 'visibility:hidden;position:absolute;left:-40px;');
```
tananaev: Is the `left:-40px` required, or `position:absolute` would be enough? I just don't like that we have some hardcoded number there. If it's required, what if we put `0px`?

tananaev: OK, let's use `-100px` for now.

Abyss777:
Replaced commit with `-100px`

tananaev:
Merged. Thanks for pull request.

Abyss777:
I guess "Merge" button just was not pressed :-)

tananaev:
Yes, looks like I forgot to click the button. Too many pull requests :)

PR #385, opened 2017-01-16
User management

Abyss777:
Web part for https://github.com/tananaev/traccar/pull/2801

New user field is added

image

Admin has new button (highlighted only for managers)
image

to link users
image

Manager see "Settings" -> "Users"
image

To see linked users and himself
image

Editable fields for other users
image

Editable fields for manager himself
image

Also fixed two issues:
- Variable check in `StateController`
- Simplified geofences update (they may duplicated sometimes)

### DIFF
--- a/web/app/model/User.js
+++ b/web/app/model/User.js
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 Anton Tananaev (anton@traccar.org)
+ * Copyright 2015 - 2017 Anton Tananaev (anton@traccar.org)
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -70,6 +70,9 @@ Ext.define('Traccar.model.User', {
         dateFormat: 'c'
     }, {
         name: 'deviceLimit',
+        type: 'int'
+    }, {
+        name: 'userLimit',
         type: 'int'
     }, {
         name: 'token',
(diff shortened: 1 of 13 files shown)


tananaev:
Why on the screenshot expiration and token fields are enabled for regular user, but disabled for manager?

Abyss777:
I would say disabled for manager himself. The logic is following:

1. Our initial decision was that token is rather sensitive and only admin be able to set it. Now manager can set other users tokens, but only admin can set tokens of managers.
2. Manager is ordinary computer user but with extended permissions. He should not be able to shut in his own leg, like disable himself. 
3. Also admin can set expiration time for manager (i.e. company account was paid till only that time) and manager himself should not be able change that.

PS: do not see other comments

tananaev:
1. A bit confusing. Maybe we should just enable all users to generate their tokens?

2. I think it doesn't make sense to add extra code to just handle this case. What about admin? He can disable himself, right?

3. But then manager can create a user account past his own expiration time, right?

By the way, are all the restrictions handled the same way on the server side, or is it just client-side?

Abyss777:
1. We should not forget about use case then user is public, such user should not be able change own token.
2. Yes, right. Admin can everything including disable himself.
3. Yes this case is not handled yet.

All restrictions are reflection of server side https://github.com/tananaev/traccar/blob/master/src/org/traccar/database/PermissionsManager.java#L225

tananaev:
Merged it. I feel like the permissions system became a bit too complicated.

PR #390, opened 2017-01-26
Improve permissions

### LINKED ISSUE #380: Problem with imports
Some "requires" definitions are incorrect. For example:
```JavaScript
Ext.define('Traccar.view.DeviceDialog', {
    extend: 'Traccar.view.BaseEditDialog',

    requires: [
        'Traccar.view.BaseEditDialog'
    ],

    controller: 'baseEditDialog',
```

In this case, instead of importing `BaseEditDialog` which is already imported via `extend`, we should import `BaseEditDialogController` which is required for `baseEditDialog` controller.

I guess we should go through the code and find if there are any other similar issues.

### LINKED ISSUE #387: Token is not fully visible
Because token field is disabled for a regular user, he can't read the full token value. We need to fix it somehow.

Abyss777:
Adjusted web interface to https://github.com/tananaev/traccar/issues/2810 https://github.com/tananaev/traccar/pull/2831 https://github.com/tananaev/traccar/pull/2835

User from admin account
image

New user from manager account
image

Ordinary user account (I've decided not hide "Device readonly" field, just disable)
image
 
Readonly view
image

Device readonly view
image

fix #387 

Also made up imports, fix #380

### DIFF
--- a/web/app/view/DevicesController.js
+++ b/web/app/view/DevicesController.js
@@ -48,10 +48,12 @@ Ext.define('Traccar.view.DevicesController', {
     },
 
     init: function () {
-        var readonly = Traccar.app.getPreference('readonly', false) && !Traccar.app.getUser().get('admin');
-        this.lookupReference('toolbarAddButton').setVisible(!readonly);
-        this.lookupReference('toolbarEditButton').setVisible(!readonly);
-        this.lookupReference('toolbarRemoveButton').setVisible(!readonly);
+        var readonly, deviceReadonly;
+        deviceReadonly = Traccar.app.getUser().get('deviceReadonly');
+        readonly = Traccar.app.getPreference('readonly', false) && !Traccar.app.getUser().get('admin');
+        this.lookupReference('toolbarAddButton').setVisible(!readonly && !deviceReadonly);
+        this.lookupReference('toolbarEditButton').setVisible(!readonly && !deviceReadonly);
+        this.lookupReference('toolbarRemoveButton').setVisible(!readonly && !deviceReadonly);
         this.lookupReference('toolbarGeofencesButton').setVisible(!readonly);
     },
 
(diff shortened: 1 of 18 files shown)


web/app/view/DevicesController.js
```diff
@@ -87,7 +89,12 @@ Ext.define('Traccar.view.DevicesController', {
                 if (btn === 'yes') {
                     store = Ext.getStore('Devices');
                     store.remove(device);
-                    store.sync();
+                    store.sync({
+                        failure: function (batch) {
```
Abyss777: Show errors if device deleting failed

web/app/view/UsersController.js
```diff
@@ -40,6 +40,12 @@ Ext.define('Traccar.view.UsersController', {
     onAddClick: function () {
         var user, dialog;
         user = Ext.create('Traccar.model.User');
+        if (Traccar.app.getUser().get('admin')) {
+            user.set('deviceLimit', -1);
```
Abyss777: Added just not break old experiences, new user created admin will have infinite device limit by default

web/app/view/UsersController.js
```diff
@@ -40,6 +40,12 @@ Ext.define('Traccar.view.UsersController', {
...
+        }
+        if (Traccar.app.getUser().get('expirationTime')) {
+            user.set('expirationTime', Traccar.app.getUser().get('expirationTime'));
```
Abyss777: Made for convenience, if manager has expiration time, new user will have same expiration time by default.

Abyss777:
Added small fix to determine manager correctly

web/app/view/DevicesController.js
tananaev: Should we handle it for other service calls as well? Then I guess it should be shared somewhere.

Abyss777: Sure. It's not handled in some other places. I'll we check what we can do.

djbo1982:
@Abyss777  @tananaev 
Hello man, im reading about one thing in other develop and i find this document, maybe is useful for user manager

copy as original 

I created a manager and give 10 device permission.
I then logged in with the manager and created a sub-manager called the manager and give 10 device access.
I went to admin and then tried to look at the user relations
The relation does not specify or show the which manager the sub-manager belongs

Problem.


[Relational.User.manager.pdf](https://github.com/tananaev/traccar-web/files/759712/Relational.User.manager.pdf)

Abyss777:
@djbo1982 
Looks like it is not about official traccar. 
In the current implementation manager can not create manager.

Issue #392, opened 2017-01-28
Better request error handling 


Abyss777:
There are some places without request error handling and `store.sync()` without failure handling.

Also we should reject changes on failure in `BasePermissionsController` somehow.

tananaev:
Not relevant for the modern app.

PR #394, opened 2017-01-30
Implement base controller for windows with EditToolbar

Abyss777:
While fixing #392 I've found something to optimize. Implemented `EditToolbarController` for windows that have "plus/pencil/cross" buttons.

Child must set fields:
- `objectModel`
- `objectDialog` - dialog called for edit
- `removeTitle` - title for remove confirmation window

Also view store must be defined.


Abyss777:
Rejecting changes in `BasePermissionsController` is rather simple, also added it here.

tananaev:
Merged, thanks.

PR #397, opened 2017-02-01
Correctly set object store for chained store

### LINKED ISSUE #395: Can't add new device after version e774ded
After this commit e774dedf3497f640af5a686723a0f5cf0ea42b93 the add device dialog don't close after clicking confirm. The device is added to the devices panel but not to the database.

Abyss777:
Chained store do not have `sync()` function, we have to set source store for new objects.

fix #395


tananaev:

PR #401, opened 2017-02-03
Better cleanup after device remove

### LINKED ISSUE #400: State stays after removing device
Happened after removing last device from the account. Not sure if it's relevant.

Abyss777:
fix #400 

Also noticed that accuracy and live route stayed after device removed

### DIFF
--- a/web/app/view/DevicesController.js
+++ b/web/app/view/DevicesController.js
@@ -32,11 +32,11 @@ Ext.define('Traccar.view.DevicesController', {
         listen: {
             controller: {
                 '*': {
-                    selectreport: 'selectReport',
-                    deselectfeature: 'deselectFeature'
+                    selectreport: 'selectReport'
                 },
                 'map': {
-                    selectdevice: 'selectDevice'
+                    selectdevice: 'selectDevice',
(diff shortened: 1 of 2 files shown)


web/app/view/StateController.js
```diff
@@ -49,6 +49,9 @@ Ext.define('Traccar.view.StateController', {
                 '#AttributeAliases': {
                     add: 'updateAliases',
                     update: 'updateAliases'
+                },
+                '#Devices': {
+                    remove: 'removeDevice'
```
tananaev: I'm just wondering, why it doesn't get de-selected? Do we really need to explicitly do this?

Abyss777: Hm, you are right, via de-select it will look more consequent.

Abyss777:
Replaced commit.

Devices grid correctly handles device removing and fires `onSelectionChange` event.
Just fire our `deselectfeature` event if nothing selected.

Also handle `deselectfeature` event only from `map`, to avoid double call `deselectFeature` function.

tananaev:
Looks good, thanks.

PR #402, opened 2017-02-07
Hide command button for readonly users

Abyss777:
Client side for https://github.com/tananaev/traccar/pull/2878


tananaev:
Merged, thanks.

PR #415, opened 2017-02-27
Add SMS to notifications and phone to user dialog

Abyss777:
Web part of https://github.com/tananaev/traccar/pull/2942

image

image

Fullfeatured account dialog showed without scroll only in fullscreen mode (1920x1080)


tananaev:
Merged, thanks. I wonder if we should move some parameters into a collapsable view. Not sure if something like that is available in ExtJS.

Abyss777:
I think it is possible http://examples.sencha.com/extjs/6.2.0/examples/kitchensink/#form-register
`fieldset` can be collapsible http://docs.sencha.com/extjs/6.2.0/classic/Ext.form.FieldSet.html#cfg-collapsible
I'll made some research.

tananaev:
That's what I was thinking about. If we add this, we should probably add it to all models. Also, it would be nice to split required parameters from optional. For example, in device model we only need name and id as required. Many people get confused if they need to fill everything. We can collapse all optional parameters by default.

Abyss777:
Good idea. I'll made new issue when start developing where we discuss how to group fields and other related questions.

PR #416, opened 2017-02-27
Add server "device readonly" flag

### LINKED ISSUE #410: Readonly user toolbar issue
@Abyss777, I think we should disable buttons on the toolbars for readonly users instead of hiding them. Otherwise the height of the toolbar shrinks and it doesn't look good:

screen shot 2017-02-21 at 11 43 01 am

I think we can have a general rule to never hide the buttons on the toolbar, just disable them.

### LINKED ISSUE #411: Device readonly user
@Abyss777, I have heard a few complaints that readonly user can't send commands anymore. I propose that we allow "device readonly" user to send commands. The idea is that he can't add or remove devices, but he can still manage his own devices, including sending commands.

Abyss777:
Web-part for https://github.com/tananaev/traccar/pull/2960

Also do not hide buttons in Devices toolbar, just disable.

fix #410 and fix #411

### DIFF
--- a/web/app/view/DevicesController.js
+++ b/web/app/view/DevicesController.js
@@ -53,13 +53,9 @@ Ext.define('Traccar.view.DevicesController', {
 
     init: function () {
         var readonly, deviceReadonly;
-        deviceReadonly = Traccar.app.getUser().get('deviceReadonly');
+        deviceReadonly = Traccar.app.getPreference('deviceReadonly', false) && !Traccar.app.getUser().get('admin');
         readonly = Traccar.app.getPreference('readonly', false) && !Traccar.app.getUser().get('admin');
-        this.lookupReference('toolbarAddButton').setVisible(!readonly && !deviceReadonly);
-        this.lookupReference('toolbarEditButton').setVisible(!readonly && !deviceReadonly);
-        this.lookupReference('toolbarRemoveButton').setVisible(!readonly && !deviceReadonly);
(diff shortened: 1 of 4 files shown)


tananaev:

web/app/view/DevicesController.js
```diff
@@ -53,13 +53,9 @@ Ext.define('Traccar.view.DevicesController', {
...
         var readonly, deviceReadonly;
-        deviceReadonly = Traccar.app.getUser().get('deviceReadonly');
+        deviceReadonly = Traccar.app.getPreference('deviceReadonly', false) && !Traccar.app.getUser().get('admin');
         readonly = Traccar.app.getPreference('readonly', false) && !Traccar.app.getUser().get('admin');
-        this.lookupReference('toolbarAddButton').setVisible(!readonly && !deviceReadonly);
-        this.lookupReference('toolbarEditButton').setVisible(!readonly && !deviceReadonly);
-        this.lookupReference('toolbarRemoveButton').setVisible(!readonly && !deviceReadonly);
-        this.lookupReference('deviceCommandButton').setVisible(!readonly && !deviceReadonly);
-        this.lookupReference('toolbarGeofencesButton').setVisible(!readonly);
+        this.lookupReference('toolbarAddButton').setDisabled(readonly || deviceReadonly);
```
Abyss777: Variables can be removed, but I think it is more visual

PR #417, opened 2017-02-27
Fix width of checkcolumns

### LINKED ISSUE #413: Permissions check column to wide
I suppose after this commit https://github.com/tananaev/traccar-web/commit/1db0f3fce321a6f22da47998a53f6a5bf7bc7b36
check column become too wide
image

May be It should be adjusted some how.

Abyss777:
Moved `flex` and `minWidth` from `defaults` to `columns`
fix #413 

image

Also adjusted a bit width of Notifications columns (types was cut). Set `flex: 2`

Before:
image

After:
image


tananaev:
Merged, thanks.

PR #421, opened 2017-03-03
Save cookie credentials in base64

### LINKED ISSUE #419: Remembered password saved incorrectly
Chars in password are replaced by html codes in cookie store, for example `pa$$word` by `pa%24%24word`. 

image

And then passed to server as is and login fails.

image

Abyss777:
fix #419 and web part for https://github.com/tananaev/traccar/pull/2973

### DIFF
--- a/web/app/view/LoginController.js
+++ b/web/app/view/LoginController.js
@@ -40,11 +40,14 @@ Ext.define('Traccar.view.LoginController', {
                 url: 'api/session',
                 params: form.getValues(),
                 callback: function (options, success, response) {
+                    var user, password;
                     Ext.get('spinner').setVisible(false);
                     if (success) {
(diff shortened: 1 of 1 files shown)


tananaev:
Merged.

By the way, what would happen if someone has old cookies from previous version?

Abyss777:
Authentication will fail (otherwise user exists in DB with email and password equal base64 of original) and user will be asked for user/password again.

tananaev:
That OK. I just wanted make sure that it won't keep trying to use invalid data over and over.

Abyss777:
Frankly it will try, but not every time, just if there is no session on the server (after server restart).
May be also add clear cookies on login fail?

tananaev:
I think it's fine as it is, as long as it still asks for regular login.

PR #422, opened 2017-03-07
Implement SMS commands

Abyss777:
Web-part for https://github.com/tananaev/traccar/pull/2971 and https://github.com/tananaev/traccar/pull/2982

added checkbox for transport
image

Allowed send commands to offline devices, and automatically lock to SMS
image

Also added exception handler in case something happen while loading supported command types.

### DIFF
--- a/web/app/store/CommandTypes.js
+++ b/web/app/store/CommandTypes.js
@@ -1,5 +1,6 @@
 /*
  * Copyright 2016 Gabor Somogyi (gabor.g.somogyi@gmail.com)
+ * Copyright 2017 Andrey Kunitsyn (andrey@traccar.org)
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -19,17 +20,9 @@ Ext.define('Traccar.store.CommandTypes', {
     extend: 'Ext.data.Store',
     fields: ['type', 'name'],
 
-    listeners: {
-        'beforeload' : function (store) {
-            var proxy;
-            proxy = store.getProxy();
-            proxy.setUrl('api/commandtypes?deviceId' + proxy.extraParams.deviceId);
-        }
-    },
-
     proxy: {
         type: 'rest',
-        url: '',
+        url: 'api/commandtypes',
         reader: {
             type: 'json',
             getData: function (data) {
@@ -45,6 +38,11 @@ Ext.define('Traccar.store.CommandTypes', {
                     }
                 });
                 return data;
+            }
+        },
+        listeners: {
+            'exception' : function (proxy, response) {
+                Traccar.app.showError(response);
             }
         }
     }
--- a/web/app/view/CommandDialogController.js
+++ b/web/app/view/CommandDialogController.js
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 Anton Tananaev (anton@traccar.org)
+ * Copyright 2015 - 2017 Anton Tananaev (anton@traccar.org)
  *
  * This program is free software: you can redistribute it and/or modify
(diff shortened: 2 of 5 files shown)


web/app/store/CommandTypes.js
```diff
@@ -19,17 +20,9 @@ Ext.define('Traccar.store.CommandTypes', {
     extend: 'Ext.data.Store',
     fields: ['type', 'name'],
 
-    listeners: {
-        'beforeload' : function (store) {
```
Abyss777: It was weird.

web/app/view/CommandDialogController.js
```diff
@@ -113,9 +113,18 @@ Ext.define('Traccar.view.CommandDialogController', {
...
+        typesStore.getProxy().setExtraParam('textChannel', newValue);
+        typesStore.reload();
+    },
+
     onSendResult: function (options, success, response) {
         if (success) {
-            Ext.toast(Strings.commandSent);
+            Ext.toast({
+                html: Strings.commandSent,
+                align: 'br'
```
Abyss777: Align toast to bottom right corner

web/app/view/DevicesController.js
```diff
@@ -74,28 +74,30 @@ Ext.define('Traccar.view.DevicesController', {
     },
 
     onCommandClick: function () {
```
tananaev: Please add some spacing to this method. It became too big and it's really hard to understand what it does.

web/app/view/CommandDialogController.js
tananaev: I thought we are aligning everything in the middle. Am I wrong?

web/app/store/CommandTypes.js
tananaev: How does it work now? Where does device id come from?

Abyss777: We are setting it in `DevicesController` as extra param, ExtJS everything does for us, we do not need add this parameters manually.

web/app/view/CommandDialogController.js
Abyss777: Default behavior was changed in ExjJS 6.20 and we decided to revert it back to bottom-right https://github.com/tananaev/traccar-web/pull/355

web/app/view/DevicesController.js
```diff
@@ -74,28 +74,30 @@ Ext.define('Traccar.view.DevicesController', {
...
+        var device, deviceId, command, dialog, comboStore, online;
         device = this.getView().getSelectionModel().getSelection()[0];
+        online = device.get('status') === 'online';
         deviceId = device.get('id');
         command = Ext.create('Traccar.model.Command');
         command.set('deviceId', deviceId);
+        command.set('textChannel', !online);
         dialog = Ext.create('Traccar.view.CommandDialog');
         comboStore = dialog.down('form').down('combobox').getStore();
         comboStore.getProxy().setExtraParam('deviceId', deviceId);
```
Abyss777: We are setting `deviceId` here as extra param

web/app/store/CommandTypes.js
tananaev: OK.

web/app/view/CommandDialogController.js
tananaev: OK, I forgot about it.

web/app/view/DevicesController.js
tananaev: I still think we should add some spaces to the method to make code more readable.

Abyss777:
Added spaces
Renamed variable to something more meaningful
User `lookupReference` instead of  `down().down()`

tananaev:
Merged, thanks.

PR #424, opened 2017-03-10
Rename "Test Email" to "Test Notification"

Abyss777:
We are testing both transports now, not only email.


tananaev:
Merged, thanks.

PR #425, opened 2017-03-10
Group user fields in fieldsets

Abyss777:
Grouped in fieldsets and reordered a bit

image

image

image


web/app/view/BaseDialog.js
```diff
@@ -22,6 +22,7 @@ Ext.define('Traccar.view.BaseDialog', {
     resizable: false,
     modal: true,
     autoScroll: true,
+    constrain: true,
```
Abyss777: Do not allow dialog windows be moved or expanded outside parent container (`document.body` by default)

tananaev:
Looks good. I think we should do this for all dialogs where it makes sense. For example, definitely for devices.

PR #426, opened 2017-03-10
Group device fields

Abyss777:
Only grouped fields, nothing reordered

image

image


tananaev:
I would rename "Preferences" to something like "Other" or "Optional" or "Extra" or something like that. Preferences make sense only for server and user account.

Abyss777:
I like "Extra"
image

tananaev:
Merged, thanks.

PR #429, opened 2017-03-11
Group fields in Server dialog

Abyss777:
Made Permissions collapsible and Preferences not

image

image


tananaev:
Merged, thanks.

PR #432, opened 2017-03-11
Use init instead of onShow in UserDialog

Abyss777:
Pass `selfEdit` parameter to `UserDialog`

Use disabling instead of hiding
image

image

Also grouped Geofence fields

image

image


tananaev:

PR #434, opened 2017-03-12
Remove selType

### LINKED ISSUE #428: Use selModel
Use `selModel` instead of `selType` (deprecated).

Abyss777:
Just removed `selType` for all views where it was, because selection model is `rowmodel` by default
http://docs.sencha.com/extjs/6.2.0/classic/Ext.grid.Panel.html#cfg-selModel

In other places we use actual `selModel`

fix #428


tananaev:
Merged, thanks.

PR #435, opened 2017-03-12
Group fields in Group and Calendar dialog

Abyss777:
image

image

image


tananaev:
Merged, thanks again.

PR #438, opened 2017-03-14
Implement Timezone field

Abyss777:
A was thinking about list and came to this:
- Just list of GMT+ and GMT- will not cover timezones with daylight saving
- If I took Microsoft list, I would prepare confrontation "Microsoft name" -> TZ ID for about 80 lines. It is not worth spent time.

Finally I just took full list from wikipedia without links. Looks fine to me on desktop and on mobile.

image

Also I had doubt what to choose `editable: false` or `typeAhead: true`. Stopped on the `editable: false` because `typeAhead` is useless on mobile.


tananaev:
Looks fine, merged. Thanks.

PR #440, opened 2017-03-15
Zoom to all devices on start

Abyss777:
Zoom to all devices on start if lat/lon/zoom preferences are not set.

We might add some button "Zoom to all" somewhere in future.

### DIFF
--- a/web/app/controller/Root.js
+++ b/web/app/controller/Root.js
@@ -167,6 +167,9 @@ Ext.define('Traccar.controller.Root', {
                 self.updateEvents(data.events);
             }
         };
+        if (first) {
+            this.first = true;
+        }
     },
 
     updateDevices: function (array) {
@@ -196,6 +199,10 @@ Ext.define('Traccar.controller.Root', {
                 store.add(Ext.create('Traccar.model.Position', array[i]));
             }
         }
+        if (this.first) {
+            this.zoomToAllDevices();
+            this.first = false;
+        }
     },
 
     updateEvents: function (array) {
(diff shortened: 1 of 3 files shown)


web/app/controller/Root.js
```diff
@@ -167,6 +167,9 @@ Ext.define('Traccar.controller.Root', {
                 self.updateEvents(data.events);
             }
         };
+        if (first) {
+            this.first = true;
```
Abyss777: I meant global for `Root.js`

web/app/view/MapController.js
```diff
@@ -98,5 +99,30 @@ Ext.define('Traccar.view.MapController', {
                 return true;
             }, this);
         }
+    },
+
+    zoomToAllDevices: function () {
+        var data, i, point, minx, miny, maxx, maxy;
```
tananaev: We have this login somewhere already, so maybe we can share it.

web/app/controller/Root.js
tananaev: It would be nice if we could use method parameter instead of instance variable.

Abyss777: The problem is that `updatePositions` is called from `socket.onmessage` callback and we need switch it off after first frame.

tananaev: We can set local variable to false.

Abyss777:
Something like that? (last commit)

web/app/controller/Root.js
```diff
@@ -124,7 +124,7 @@ Ext.define('Traccar.controller.Root', {
     },
 
     asyncUpdate: function (first) {
-        var self = this, protocol, pathname, socket;
+        var self = this, protocol, pathname, socket, firstPositions = first;
```
tananaev: I think you can just use `first` directly.

Abyss777:
Removed intermediate variable

tananaev:
Merged, thanks.

PR #442, opened 2017-03-18
Initialize stores on application load

### LINKED ISSUE #441: Inconsistency with store loading
I noticed that in `GeofencesController` we load `Geofences` and `Calendars` stored. Do we really need to do that? It means every time we open the dialog, we do two requests. Geofences should be loaded on startup. Maybe we should do same thing with calendars as well.

@Abyss777, what do you think?

Abyss777:
- Load `Geofences` and `Calendars` stores on application launch
- Initialize `ReportEventTypes` store on application launch
- Make two `Device` pickers not-editable 

fix #441


tananaev:
Merged, thanks.

PR #443, opened 2017-03-20
Strings for text message Event and some Alarms

Abyss777:
Probably some alarm strings should be corrected.


tananaev:
Merged, thanks.

PR #445, opened 2017-03-22
Implement extended device filtering

Abyss777:
There are three key changes:

1. Improve devices panel to use grid filters
- Removed bottom bar
- Added Status column
- Added filters for three columns: Name, Group, Status
- Increased panel width to not cut LastUpdate column by default
- Used `VisibleDevices` as chained store
2. Reflect device filtering on the map
- Set Id to all device related features to easily check feature showed or not
- Move taking decision of show or hide feature to `VisibleDevices` event handlers because only there we can make it
3. Save device panel state in cookies 
- There are a lot of possible convenient configurations of columns, I decided to store grid state to cookies, especially because it is really easy (just 5 additional lines)

Here is some screenshots
Default view
image

Name filter
image

Group filter
image

Status filter
image

Cookies in store
image

More comments in code


web/app/view/Devices.js
```diff
@@ -20,20 +20,19 @@ Ext.define('Traccar.view.Devices', {
...
-            groupField: 'groupId'
-        });
-        this.callParent();
-    },
+    plugins: 'gridfilters',
+
+    store: 'VisibleDevices',
+
+    stateful: true,
+    stateId: 'devices-grid',
```
Abyss777: Enable state autosave and set id

web/app/view/Devices.js
```diff
@@ -85,79 +84,30 @@ Ext.define('Traccar.view.Devices', {
...
-    }],
-
     listeners: {
         selectionchange: 'onSelectionChange'
     },
 
     columns: {
         defaults: {
-            flex: 1,
+            flex: 2,
```
Abyss777: All columns are likely more wide than Status

web/app/view/Devices.js
```diff
@@ -169,26 +119,46 @@ Ext.define('Traccar.view.Devices', {
...
+        }, {
+            text: Strings.deviceStatus,
+            dataIndex: 'status',
+            flex: 1,
+            filter: {
+                type: 'list',
+                labelField: 'name',
+                store: 'DeviceStatuses'
+            },
+            renderer: function (value, metaData) {
```
Abyss777: This renderer and filter relies on https://github.com/tananaev/traccar/pull/3034

web/app/view/MapMarkerController.js
```diff
@@ -198,11 +204,11 @@ Ext.define('Traccar.view.MapMarkerController', {
             } else {
                 feature = new ol.Feature(new ol.geom.Circle(center, radius));
                 feature.setStyle(this.getAreaStyle(null, Traccar.Style.mapAccuracyColor));
-                this.getView().getAccuracySource().addFeature(feature);
+                feature.setId(position.get('deviceId'));
```
Abyss777: Set feature id as device id (and for other device related features)

web/app/view/MapMarkerController.js
```diff
@@ -488,5 +494,56 @@ Ext.define('Traccar.view.MapMarkerController', {
...
+    },
+
+    checkDeviceVisibility: function (device) {
+        var deviceId, accuracy, liveLine, marker;
+        deviceId = device.get('id');
+        marker = this.latestMarkers[deviceId];
+        accuracy = this.accuracyCircles[deviceId];
+        liveLine = this.liveRoutes[deviceId];
+        if (Ext.getStore('VisibleDevices').contains(device)) {
+            if (marker && !this.getView().getMarkersSource().getFeatureById(marker.getId())) {
```
Abyss777: Add feature only if it is not already there

web/app/view/MapMarkerController.js
```diff
@@ -488,5 +494,56 @@ Ext.define('Traccar.view.MapMarkerController', {
...
+                this.getView().getMarkersSource().addFeature(marker);
+            }
+            if (accuracy && !this.getView().getAccuracySource().getFeatureById(accuracy.getId())) {
+                this.getView().getAccuracySource().addFeature(accuracy);
+            }
+            if (liveLine && !this.getView().getLiveRouteSource().getFeatureById(liveLine.getId())) {
+                this.getView().getLiveRouteSource().addFeature(liveLine);
+            }
+        } else {
+            if (marker && this.getView().getMarkersSource().getFeatureById(marker.getId())) {
```
Abyss777: Remove feature only if it is there

web/app/view/MapMarkerController.js
```diff
@@ -488,5 +494,56 @@ Ext.define('Traccar.view.MapMarkerController', {
...
+                this.getView().getMarkersSource().removeFeature(marker);
+            }
+            if (this.selectedMarker && marker && marker.getId() === this.selectedMarker.getId()) {
+                this.deselectFeature();
```
Abyss777: Deselect feature if need to hide it

web/app/view/MapMarkerController.js
```diff
@@ -488,5 +494,56 @@ Ext.define('Traccar.view.MapMarkerController', {
...
+            if (accuracy && this.getView().getAccuracySource().getFeatureById(accuracy.getId())) {
+                this.getView().getAccuracySource().removeFeature(accuracy);
+            }
+            if (liveLine && this.getView().getLiveRouteSource().getFeatureById(liveLine.getId())) {
+                this.getView().getLiveRouteSource().removeFeature(liveLine);
+            }
+        }
+    },
+
+    filterDevices: function (store) {
```
Abyss777: Handle filter changing here

web/app/view/Devices.js
```diff
@@ -169,24 +119,46 @@ Ext.define('Traccar.view.Devices', {
...
+                if (value) {
```
Abyss777: It's turned out that rendering happens twice after device was added. First time with empty string, second time with value from server response (offline).

Abyss777:
How it looks on mobile
image

web/app/controller/Root.js
```diff
@@ -25,6 +25,10 @@ Ext.define('Traccar.controller.Root', {
         'Traccar.model.Position'
     ],
 
+    init: function () {
+        Ext.state.Manager.setProvider(new Ext.state.CookieProvider());
```
tananaev: This is amazing. Can it save status of everything else as well (e.g. buttons on the map)? This way we won't need to save them in the user attributes.

web/app/store/DeviceStatuses.js
```diff
@@ -0,0 +1,35 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+Ext.define('Traccar.store.DeviceStatuses', {
+    extend: 'Ext.data.Store',
+    fields: ['id', 'name', 'tdCls'],
+
+    data: [{
+        id: 'online',
+        name: Strings.deviceStatusOnline,
+        tdCls: 'view-color-green'
```
tananaev: I think we can use a better name instead of `tdCls`. Maybe just `style` or something.

web/app/view/Devices.js
```diff
@@ -169,24 +119,46 @@ Ext.define('Traccar.view.Devices', {
...
+                    metaData.tdCls = Ext.getStore('DeviceStatuses').getById(value).get('tdCls');
```
tananaev: I think it would be better to call `getById` just once.

web/app/view/MapMarkerController.js
```diff
@@ -488,5 +494,56 @@ Ext.define('Traccar.view.MapMarkerController', {
         } else if (point) {
             this.getView().getMapView().fit(new ol.geom.Point(point));
         }
+    },
+
+    updateVisibleDevice: function (store, data) {
```
tananaev: Does this method ever gets array? I guess then it would make sense to rename it to plural "devices".

web/app/view/MapMarkerController.js
tananaev: What would happen if we call add if it's already there? Would it crash or OpenLayers would check it internally? If latter, it would be more readable to avoid this check.

web/app/view/MapMarkerController.js
tananaev: Wouldn't device grid do it automatically?

web/app/view/MapMarkerController.js
tananaev: Same question... can we remove feature if it's not there? Would it crash?

web/l10n/en.json
```diff
@@ -84,6 +84,11 @@
     "deviceCommand": "Command",
     "deviceFollow": "Follow",
     "deviceTotalDistance": "Total Distance",
+    "deviceStatus": "Status",
+    "deviceSearchFor": "Search for...",
```
tananaev: Do we really need this message? I think it should be pretty obvious even if the field is empty.

web/app/view/MapMarkerController.js
Abyss777: It throws this exception https://openlayers.org/en/v4.0.0/doc/errors/#30 if feature already there
and this then try to remove not existed feature
image

web/app/controller/Root.js
Abyss777: Looks like we can http://docs.sencha.com/extjs/6.2.0/classic/Ext.button.Button.html#cfg-stateful

web/l10n/en.json
Abyss777: Unfortunately it is not empty
image

Abyss777:
Fixed all your comments, except `get*Source().getFeatureById()` check, unfortunately we can't avoid it, but I hope "byId" does not cost a lot.

Also fixed one missed issue, show features on `updateLatest` as soon as possible and in more correct way, not waiting next `updateDevice` event.

Also added `allowNull: true` to string fields of Device model, If need I will add it to all other models in other pull request.

Abyss777:
Tomorrow is a new day.
I was thinking... We can implement some flag inside features itself to know if it is visible or not instead of checking `get*Source().getFeatureById()`. It might be faster. What do you think?

tananaev:
Merged the code, thanks.

I think we can leave the check there for now. If we find any performance issues, we can always implement optimization later.

PR #447, opened 2017-03-24
Use cookies to save buttons state instead of User/Server attributes

Abyss777:
Removed previous way of saving and optimized toolbar a bit with help of `defaults`

image

### DIFF
--- a/web/app/view/Map.js
+++ b/web/app/view/Map.js
@@ -71,23 +71,10 @@ Ext.define('Traccar.view.Map', {
             stateId: 'device-follow-button',
             toggleHandler: 'onFollowClick'
         }, {
-            id: 'muteButton',
-            glyph: 'xf1f7@FontAwesome',
-            tooltip: Strings.sharedMute,
-            pressed: true,
-            stateId: 'mute-button',
-            listeners: {
-                toggle: function (button, pressed) {
-                    if (pressed) {
-                        button.setGlyph('xf1f7@FontAwesome');
-                    } else {
-                        button.setGlyph('xf0a2@FontAwesome');
-                    }
-                }
-            },
-            applyState: function (state) {
-                this.toggle(state.pressed);
-            }
+            id: 'soundButton',
+            glyph: 'xf0a2@FontAwesome',
+            tooltip: Strings.sharedSound,
+            stateId: 'sound-button'
         }, {
             xtype: 'settingsMenu',
             enableToggle: false
--- a/web/app/view/Map.js
+++ b/web/app/view/Map.js
@@ -29,6 +29,15 @@ Ext.define('Traccar.view.Map', {
     title: Strings.mapTitle,
     tbar: {
         componentCls: 'toolbar-header-style',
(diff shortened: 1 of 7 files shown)


web/app/view/Map.js
```diff
@@ -29,60 +29,68 @@ Ext.define('Traccar.view.Map', {
...
                         button.setGlyph('xf1f7@FontAwesome');
                     } else {
                         button.setGlyph('xf0a2@FontAwesome');
                     }
-                },
-                scope: this
+                }
+            },
+            applyState: function (state) {
+                this.toggle(state.pressed);
```
Abyss777: Need it to correctly set glyph according to restoring state.

tananaev: Maybe we should remove the glyph switching? Less code and more consistency with other buttons.

Abyss777: OK. I think we should leave "not-crossed-bell" then and invert the logic: Pressed mean play sound.
And may be rename the button from "mute" to "sound" or something.

tananaev: Yes, agree. Pressed should mean play sound and rename button as well.

Abyss777:
Made changes
image

web/app/view/Map.js
```diff
@@ -29,60 +29,55 @@ Ext.define('Traccar.view.Map', {
...
-                    } else {
-                        button.setGlyph('xf0a2@FontAwesome');
-                    }
-                },
-                scope: this
-            }
+            id: 'soundButton',
+            glyph: 'xf0a2@FontAwesome',
+            tooltip: Strings.sharedSound,
+            stateId: 'sound-button'
```
tananaev: Do we need this id?

web/l10n/en.json
```diff
@@ -29,13 +29,12 @@
     "sharedAttributes": "Attributes",
     "sharedAttribute": "Attribute",
     "sharedArea": "Area",
-    "sharedMute": "Mute",
+    "sharedSound": "Play Sound",
```
tananaev: I would change it to something like "Alarm Sound" or "Sound Alarm", so it's clear what it's for.

Abyss777: May be "Notification Sound"? It is not only about alarm.

web/app/view/Map.js
Abyss777: We use `getCmp` in controller and need this `id`. I can replace the logic to inter-controller events, but it will be more complex than `getCmp`

web/l10n/en.json
tananaev: Agree.

web/app/view/Map.js
tananaev: Are you sure you checked line? I was talking about `stateId`.

Abyss777: Oh. Sorry.
`stateId` is name of cookie to save state
image

tananaev:
Thanks for explanation. Merging the code now.

PR #448, opened 2017-03-25
Save state of panels and add filter to Users dialog

### LINKED ISSUE #446: Save state of buttons, grids and panels
My suggestion is:

- Save state of buttons on the map tool bar
- Save state of viewport/panels width/height collapsed or not
- May be add more columns and save state of Users grid

Another question is what will we do with current way of save/restore buttons state? Remove it or combine?
For example: 
- Cookies have priority, if empty apply user or server attributes
- Or vise versa, cookies are applied only if users or server attributes are absent.

Abyss777:
- Save state of State panel - columns and collapsed/width in mobile view
- Save state of DevicesAndState panel in desktop view - width and collapsed
- Added filter to email column in Users dialog, added two columns

image

image

I think it is last part of fix #446

### DIFF
--- a/web/app/view/Main.js
+++ b/web/app/view/Main.js
@@ -41,6 +41,8 @@ Ext.define('Traccar.view.Main', {
         title: Strings.devicesAndState,
         titleCollapse: true,
         floatable: false,
+        stateful: true,
+        stateId: 'devices-and-state-panel',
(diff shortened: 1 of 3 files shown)


tananaev:
Merged, thanks.

PR #449, opened 2017-03-27
Fix two small deselecting issues

Abyss777:
- Clear report was causing clear State even if Device was selected
- If report was selected and "Show" button clicked second time, selection on map was reseted, but selection in report grid not. Just reset selection in report grid on every "Show" button click for consistence.


tananaev:
Merged, thanks.

PR #453, opened 2017-03-28
Move map related classes to subfolder

Abyss777:
We decided it some time ago because to much files already.

Also I think move all dialog to subfolder in next PR.


tananaev:
Merged, thanks.

PR #454, opened 2017-03-28
Move BaseDialog, subclasses and their controllers to dialog subfolder

Abyss777:
Moved Login, Register, Command, DeviceDistance and all object edit dialog. They all children or grandchildren BaseDialog.

I'm just thinking, may be we should remove `Dialog` from names because it is in path already, what do you think?


web/app/view/dialog/ReportConfigController.js
```diff
@@ -16,7 +16,7 @@
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
-Ext.define('Traccar.view.ReportConfigController', {
+Ext.define('Traccar.view.dialog.ReportConfigController', {
```
tananaev: I think this class should be renamed to `ReportConfigDialogController`. Also please check that all others match naming convention. Controller should have the same name as the view plus the "Controller" suffix.

tananaev:
Agree about the removal of "Dialog" from the names, if it's already in the path.

Abyss777:
Removed `Dialog` from names and aliases.
Two controller names became consistent to their dialog names: `DeviceDistanceController` and `ReportConfigController`

tananaev:
Merged, thanks. Everything looks much cleaner now.

Abyss777:
I have two more things to group and move in mind:
- `grid.Panel`s with `BasePermissionsController` used for permissions `User*` and `*Geofences`, move to subfolder "permissions" (or something shorter?)
- `grid.Panel`s with `EditTollbar` used for editing objects except `Devices`, move to some folder, but do not know how to call it.

What do you think?

tananaev:
- Agree with `permissions` subfolder. It sounds good to me.
- Maybe call it just `edit` or something like this.

PR #455, opened 2017-03-29
Move permissions windows to subfolder

Abyss777:
Another portion of optimization. Moved permissions windows to subfolder.
Moved all common things to `Traccar.view.permissions.Base` class.

Unfortunately git incorrectly recognized some files moving, if it is not OK, I will split commit to help git.


tananaev:
Looks good. Merged.

PR #456, opened 2017-03-29
Move edit windows to edit subdirectory

Abyss777:
Moved `grid.Panel`s where we editing objects to `edit` subfolder
`EditToolbar` renamed to `edit.Toolbar`, same for controller.

I was thinking move or not `Devices` and decided why not to move it if we already moved `Map`


tananaev:
Merged, thanks.

PR #457, opened 2017-03-29
Zoom to edited geofence or main Map

### LINKED ISSUE #373: [UX] When adding a geofence area, use the current map position
1. Add geofence
2. Click area
3. Map is zoomed to somewhere

Better:

3. Map is zoomed to the current main map position

Abyss777:
Fixed zoom to editing geofence or zoom to same position as on main Map.

fix #373

### DIFF
--- a/web/app/view/map/GeofenceMap.js
+++ b/web/app/view/map/GeofenceMap.js
@@ -65,7 +65,7 @@ Ext.define('Traccar.view.map.GeofenceMap', {
         map = this.map;
 
         this.features = new ol.Collection();
-        if (this.area !== '') {
+        if (this.area) {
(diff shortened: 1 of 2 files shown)


web/app/view/map/GeofenceMap.js
```diff
@@ -68,11 +68,9 @@ Ext.define('Traccar.view.map.GeofenceMap', {
         if (this.area !== '') {
```
tananaev: Unrelated to the change, but maybe we should change this to just `if (this.area) {`?

Abyss777:
Sure. Fixed.

tananaev:
Merged, thanks.

PR #463, opened 2017-04-02
Save panel states independently for desktop and mobile

### LINKED ISSUE #462: Orientation state issue
I noticed an issue on pad:
- Open Traccar Manager in vertical view (site will choose MainMobile)
- Leave both Devices and State panels collapsed
- Close application
- Open in horizontal view (site will choose Main)
- Both panels inside common panel "Devices and State" will be collapsed

I'll investigate how to handle it.

Also I'll check if it is possible change view on the fly with orientation change.

Abyss777:
Save grid states with different ids for mobile and desktop layout

fix #462


tananaev:
Merge, thanks. Also thanks for fixing style issue.

PR #466, opened 2017-04-04
Suppress events on failure to avoid infinite loop

### LINKED ISSUE #465: GUI error and blocked - Traccar 3.10
Scenario:
Admin makes an user  with User Limit =2  (non zero) ; the other fields in permission section not checked.
Admin assigns to this user some vehicles .
Admin make a group and assigns to this group the same (or some) vehicles. Also assigns this group  to the user just created.

**First case.**
The user just  created signs in ; then goes to menu in  users and see that group assigned by the admin.
If he tries to uncheck  that group assigned by administrator then the error bellow occurs and the only way to get rid of it is to close the browser window.
image
**Second case**
The user signs in ; then goes to menu in  users and see the vehicles assigned for him by the admin.
If he tries to uncheck  any of those vehicles assigned by administrator then the error bellow occurs and the only way to get rid of it is to close the browser window.
image

Abyss777:
fix #465

### DIFF
--- a/web/app/view/permissions/BaseController.js
+++ b/web/app/view/permissions/BaseController.js
@@ -57,7 +57,7 @@ Ext.define('Traccar.view.permissions.BaseController', {
             jsonData: Ext.util.JSON.encode(data),
             callback: function (options, success, response) {
                 if (!success) {
-                    selection.deselect(record);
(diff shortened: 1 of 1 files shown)


tananaev:
Merged, thanks.

PR #467, opened 2017-04-05
Known attributes

Abyss777:
Here is implementation of known attributes editing.
Key changes:
- Added 5 stores for all objects where we currently use some attributes
- Dinamically replace name `textfield` to `combobox` for known objects
- Dinamically replace value `textfield` according to attribute type
- Dinamically apply some limits to numbers if defined
- Implemented color picker

Number
image

Invalid number
image

Boolean
image

About color picker... firstly I decided to use `ux` component http://examples.sencha.com/extjs/6.2.0/examples/kitchensink/#form-color-picker
but it is buggy, huge and does not fit in screen in mobile view. I've replaced it with small simple selector.

image

More comments in code

**PS:** not sure if we should introduce it in current release, we may postpone it to next.


web/app/view/dialog/AttributeController.js
```diff
@@ -39,5 +39,49 @@ Ext.define('Traccar.view.dialog.AttributeController', {
...
+    defaultFieldConfig: {
+        name: 'value',
+        reference: 'valueField',
+        allowBlank: false,
+        fieldLabel: Strings.stateValue
+    },
+
+    onNameChange: function (combobox, newValue) {
+        var type, config, valueField = this.lookupReference('valueField'),
+                attribute = combobox.getStore().getById(newValue);
```
Abyss777: Hope you don't mind of this way of variable declaration and initialization

tananaev:
I haven't had a chance to look through the code yet, but here are some high level comments:

- As we plan to release soon, I think it's better to not include more big changes like this. We can create a separate branch for it for now and then merge it into master after 3.11 release.
- I think we should provide proper names for known attributes with localization.
- We also need to think about adding support for convesion. For example, speed limit value. We want user to enter value in their preferred units.

Abyss777:
- Sure. Too big changes. We can leave it as PR. I will improve it, you, when have a time, will review it.

- Yes, we can make it translatable, but I suggest show them as they are in Attributes grid view.

- Yes, I had it in plans, but totally forget about it.
Speed is easy, will we convert  meters <-> kilometers or miles for maintenance?

tananaev:
Why don't we translate it in the table as well? We can actually have both translation and actual id.

As for unit conversion, I think we can do it later in the next pull request.

Abyss777:
I'll check how it will look, but I'm assuming that there might be custom attributes for own purposes and combination might look not very good. Will see...

Abyss777:
Translated known attributes

Looks fine I think

image

image


Probably we should discuss how to name them in English

web/app/view/dialog/AttributeController.js
tananaev: I think it's better to split it into separate line. It would make it more readable.

web/l10n/en.json
```diff
@@ -54,6 +54,26 @@
     "sharedPermissions": "Permissions",
     "sharedExtra": "Extra",
     "sharedTimezone": "Timezone",
+    "sharedAttributeSpeedLimit": "Speed Limit",
```
tananaev: I think it's better to remove "shared" prefix. Otherwise the key is too long.

Abyss777:
Split variable declaration and initialization
Removed "shared" prefix

tananaev:
How does the dialog looks like now? Is user able to select by "translated" name?

Abyss777:
Sure, they are translated

Some long lines take two (automatically by ExtJS)
image

But others look good
image

tananaev:
Perfect, merged.

PR #470, opened 2017-04-19
Convert speed and distance for known attributes

Abyss777:
Implemented speed and distance conversion for known attributes
I decided to invent `CustomNumberField` to show units just after the filed
image
Unfortunately, if I put bare unit in `afterSubTpl` it is placed under input field. I made a trick with `inline-table`
Code looks not very clean, but result is good and work in all 3 browsers.
image

Also formatted attributes in grid. Color at the same time.

image

### DIFF
--- a/web/app/store/DeviceAttributes.js
+++ b/web/app/store/DeviceAttributes.js
@@ -23,7 +23,8 @@ Ext.define('Traccar.store.DeviceAttributes', {
     data: [{
         key: 'speedLimit',
         name: Strings.attributeSpeedLimit,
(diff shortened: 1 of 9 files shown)


tananaev:
I'm not sure if it's a good idea to highlight color. If it's dark, it makes it hard to read the value. What do you think?

Abyss777:
Do not think somebody wants to know value if he see actual color. But agree in general, one colored string looks so so.

Removed it

image

tananaev:
Merged, thanks.

PR #472, opened 2017-04-19
Check feature existence before delete while Device removing

### LINKED ISSUE #471: Removing device issue
Crash on 3.11 when removing some devices:
```
ol-debug.js:62905 Uncaught TypeError: Cannot read property 'forEach' of undefined
    at ol.source.Vector.removeFeatureInternal (ol-debug.js:62905)
    at ol.source.Vector.removeFeature (ol-debug.js:62893)
    at constructor.removeDevice (MapMarkerController.js:168)
    at constructor.fire (ext-all-debug.js:20731)
    at constructor.dispatch (ext-all-debug.js:57952)
    at constructor.prototype.doFireEvent (ext-all-debug.js:58109)
    at constructor.fireEventArgs (ext-all-debug.js:21553)
    at constructor.fireEvent (ext-all-debug.js:21512)
    at constructor.onCollectionRemove (ext-all-debug.js:92127)
    at constructor.notify (ext-all-debug.js:74624)
```

@Abyss777, can you please take a look. Not sure how to trigger it exactly, but it seems like in some cases removing markers doesn't work properly.

Here is the forum thread:

https://www.traccar.org/forums/topic/unable-to-delete-devices-from-web-ui/

Abyss777:
It is very very very unlikely then Device and Accuracy features are not on the layer on removing stage, but I also added checks for all three type of features.

fix #471

### DIFF
--- a/web/app/view/map/MapMarkerController.js
+++ b/web/app/view/map/MapMarkerController.js
@@ -157,15 +157,21 @@ Ext.define('Traccar.view.map.MapMarkerController', {
         for (i = 0; i < data.length; i++) {
             deviceId = data[i].get('id');
             if (this.latestMarkers[deviceId]) {
-                this.getView().getMarkersSource().removeFeature(this.latestMarkers[deviceId]);
+                if (this.getView().getMarkersSource().getFeatureById(this.latestMarkers[deviceId].getId())) {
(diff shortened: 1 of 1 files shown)


tananaev:
Merged, thanks.

PR #476, opened 2017-04-22
Adding some missing strings for alarms and position - Update en.json

lalamper:
Adding some missing strings for alarms and position


web/l10n/en.json
```diff
@@ -141,6 +141,8 @@
     "positionDistance": "Distance",
     "positionRpm": "RPM",
     "positionFuel": "Fuel",
+    "positionPower": "Power",
```
tananaev: Where do we use these two?

lalamper: I was not sure about these two, but may they can be needed later. I would like to add a charts for battery and power (like speed, distance, etc.) Tried to analyse your code to find out if we need them, ReportChartTypes refers strings like this. Can you point me to the right direction to add more charts?

tananaev:
Merged, thanks. Search for "chart" keyword in the code to find out how to add more chart options.

PR #486, opened 2017-05-16
Implement computed attributes

Abyss777:
Unfortunately its name intersects with just attributes. Named in the following way:
`[object]Attributes` named shortly, in other places they are `ComputedAttributes`

Chose http://fontawesome.io/icon/tasks/ glyph

image

image

Dialog now contain only text filelds
image

I'm going to implement "Known Position Attributes" store to use combobox for `attribute` field and automatically select appropriate `type` in this dialog.


web/app/view/edit/UsersController.js
```diff
@@ -155,13 +156,31 @@ Ext.define('Traccar.view.edit.UsersController', {
         }).show();
     },
 
+    onAttributesClick: function () {
+        var user = this.getView().getSelectionModel().getSelection()[0];
+        Ext.create('Traccar.view.BaseWindow', {
+            title: Strings.sharedCalendars,
```
tananaev: This seems to be a wrong string key.

web/app/view/permissions/DeviceAttributes.js
```diff
@@ -0,0 +1,31 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+Ext.define('Traccar.view.permissions.DeviceAttributes', {
+    extend: 'Traccar.view.permissions.Base',
+    xtype: 'deviceAttributesView',
+
+    columns: {
+        items: [{
+            text: Strings.sharedDescription,
```
tananaev: Maybe we should have an attribute key in the table as well?

tananaev: Same applies to the other similar tables.

Abyss777:
Sorry for missed string
Added `attribute` column, looks fine

image

Agree about huge amount of duplicate. We can gradually deduplicate it. I'll think over what can be done.

tananaev:
Merged, thanks. I agree that we can always remove duplication and improve things later.

PR #490, opened 2017-05-18
Implement known position attributes

Abyss777:
Huge PR again
Implemented known position attributes. Set `valueType` and `dataType` according to current server state. Some attributes using should be corrected on the server side. (Charge, Battery and some others)

Used new store for columns rendering

image

Lock `type` field if selected known attribute

image

Also, for consistence, changed other objects known attributes (`convert` -> `dataType`)

There was two options for strings naming:
- `attributeXxx` like for other known attributes
- `positionXxx` like for other position attributes already implemented

I chose the last one.

User/Server models might be extended in future to handle other units: liter/gallons, Celsius/Fahrenheit etc. And appropriate formatters.


web/app/view/edit/ComputedAttributes.js
```diff
@@ -46,13 +46,29 @@ Ext.define('Traccar.view.edit.ComputedAttributes', {
...
+                }
+            }
         }, {
             text: Strings.sharedExpression,
             dataIndex: 'expression'
         }, {
             text: Strings.sharedType,
-            dataIndex: 'type'
+            dataIndex: 'type',
+            renderer: function (value) {
```
tananaev: Similar code seems to be duplicated in many places. I suggest to create a method for it somewhere to avoid duplication.

Abyss777:
Thanks. You are right. Reused code in this way.
Also fixed forgotten Device Distance `dataType`

Abyss777:
Or you were talking about 'id->Name` comparison in general?

web/l10n/en.json
```diff
@@ -146,6 +152,45 @@
...
+    "positionVin": "VIN",
+    "positionApproximate": "Approximate",
+    "positionThrottle": "Throttle",
+    "positionMotion": "Motion",
+    "positionArmed": "Armed",
+    "positionDeviceTemp": "Device Temperature",
+    "positionOperator": "Operator",
+    "positionCommand": "Acceleration",
+    "positionBlocked": "Blocked",
+    "positionDtcs": "DTCS",
```
tananaev: This should be "DTCs". Basically plural for DTC.

Abyss777:
Got it. Fixed.

tananaev:
Merged, thanks.

PR #492, opened 2017-05-23
Revert distance as key for getFormatter

Abyss777:
Reverted it, because it is used in reports https://github.com/tananaev/traccar-web/blob/master/web/app/view/ReportController.js#L453


tananaev:
Merged.

PR #493, opened 2017-05-23
Implement testing computedAttribute

Abyss777:
Used question mark as glyph
image

Simple device selector dialog
image

Empty result example
image

Some result example
image

Error example
image

Changing `Application.showError` is rather quick. Need some improvement? `info` strings may be?


web/app/Application.js
```diff
@@ -164,15 +164,24 @@ Ext.define('Traccar.Application', {
         }
     },
 
-    showError: function (error) {
+    showError: function (error, success) {
```
tananaev: It doesn't make sense to use it for successful calls. I suggest to use `Ext.Msg.alert` directly.

web/l10n/en.json
```diff
@@ -87,6 +89,8 @@
     "errorGeneral": "Invalid parameters or constraints violation",
     "errorConnection": "Connection error",
     "errorSocket": "Web socket connection error",
+    "infoTitle": "Info",
```
tananaev: Probably makes sense to prefix these two with `shared`.

web/l10n/en.json
```diff
@@ -46,9 +46,11 @@
     "sharedAlias": "Alias",
     "sharedComputedAttribute": "Computed Attribute",
     "sharedComputedAttributes": "Computed Attributes",
+    "sharedCheckComputedAttribute": "Check Computed Attribute",
     "sharedExpression": "Expression",
     "sharedDeviceDistance": "Device Distance",
     "sharedDevice": "Device",
+    "sharedSelectDevice": "Select Device",
```
tananaev: Can we just reuse `sharedDevice` here? I want to avoid adding unnecessary new string for translation.

web/app/view/dialog/SelectDeviceController.js
```diff
@@ -0,0 +1,51 @@
...
+        Ext.Ajax.request({
+            url: 'api/attributes/computed?deviceId=' + deviceId,
+            method: 'POST',
+            jsonData: Ext.util.JSON.encode(record),
+            callback: function (options, success, response) {
+                if (success) {
+                    if (response.responseText) {
+                        Ext.Msg.alert(Strings.sharedInfoTitle, Strings.sharedInfoGeneral +
+                                '<br><br><textarea readonly rows="5" style="resize: none; width: 100%;">' +
+                                response.responseText + '</textarea>');
```
tananaev: Maybe we should just show `responseText` straight away? What do you think?

Abyss777:
Looks fine
image

image

web/app/view/dialog/SelectDeviceController.js
```diff
@@ -0,0 +1,49 @@
...
+        var deviceId, record;
+        deviceId = this.lookupReference('deviceField').getValue();
+        record = this.getView().record.data;
+        Ext.Ajax.request({
+            url: 'api/attributes/computed?deviceId=' + deviceId,
+            method: 'POST',
+            jsonData: Ext.util.JSON.encode(record),
+            callback: function (options, success, response) {
+                if (success) {
+                    if (response.responseText) {
```
tananaev: Sorry for being picky, but I think we can replace it with something like this:
```javascript
Ext.Msg.alert(Strings.sharedInfoTitle, response.responseText || response.statusText);
```

Abyss777:
You are absolutely right. The sky is the limit.

tananaev:
Perfect, thanks.

PR #498, opened 2017-05-31
Implement Events panel

Abyss777:
Implemented Events panel on the right side for desktop and in the same manner as reports for mobile.

- Moved Sound toggle to Events panel to save space in mobile view.
- Formatted time as for devices and make formatter ordinary.
- Make `Event.text` calculated field
- Added switch to always scroll on the last event (last is down)
- Made panel stateful

Default View
image

Events panel are showed
image

### DIFF
--- a/web/app/Application.js
+++ b/web/app/Application.js
@@ -110,6 +110,13 @@ Ext.define('Traccar.Application', {
         }
     },
 
+    showEvents: function (show) {
+        var rootPanel = Ext.getCmp('rootPanel');
+        if (rootPanel) {
+            rootPanel.setActiveItem(show ? 2 : 0);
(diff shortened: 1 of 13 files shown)


tananaev:
Looks good. Have you checked if it conflicts with anything else that uses "Events" store?

Abyss777:
Oh, missed one place, where it is used. But I'm not sure feature it is or bug. It is used when we pass `eventId` via URL parameter.
image

If it is not OK, I can implement `LastEvents` store for this.

tananaev:
I think it's fine. I'll merge the code then.

tananaev:
Thanks for pull request.

PR #501, opened 2017-06-06
Known position attributes for reports and charts

Abyss777:
- Split converting and formatting in `AttributeFormatter` (some converters do nothing, but leaved in case we implement volume/temperature and others user settings)

- Dynamically add known position attributes fields to `Position` model. Named it `attribute.xxx` to not intersect with `attributes` object.

- Dynamically add known Number attributes to `ChartTypes`

image

- Dynamically add known position attributes to Route report as hidden columns

image

- Fixed missed strings

Some comments in the code


web/app/controller/Root.js
```diff
@@ -26,7 +26,31 @@ Ext.define('Traccar.controller.Root', {
...
+                chartTypesStore.add({
+                    key: 'attribute.' + attribute.get('key'),
+                    name: attribute.get('name')
+                });
+            }
+        }
+    },
+
+    calculateAttribute: function (data) {
+        var value = data.attributes[this.attributeKey];
```
Abyss777: I could use here something like `this.name.split('.')[1]` instead of saving original `attributeKey` but think string manipulating is rather expensive

Abyss777:
Optimized `formatValue` functions a bit

web/app/AttributeFormatter.js
```diff
@@ -35,26 +39,51 @@ Ext.define('Traccar.AttributeFormatter', {
...
 
+    volumeConverter: function (value) {
+        return Number(value);
+    },
+
     consumptionFormatter: function (value) {
         return Number(value.toFixed(Traccar.Style.numberPrecision)) + ' ' + Strings.sharedLiterPerHourAbbreviation;
     },
 
+    consumptionConverter: function (value) {
```
tananaev: Do we really need a bunch of functions that return `Number(value)`?

web/app/view/ReportController.js
```diff
@@ -383,6 +387,20 @@ Ext.define('Traccar.view.ReportController', {
         this.updateButtons();
     },
 
+    addAttributeColumns: function () {
```
tananaev: Why can't we just do this in `init`?

Abyss777:
Fixed comments and also added checks for undefined in formatters to avoid `undefined.toFixed()` cases

Abyss777:
We can't. We usually pass them as renderers that accept only one parameter

tananaev:
How about this:
```javascript
     numberFormatterFactory: function (precision, suffix) {
         return function (value) {
             if (value !== undefined) {
                 return Number(value.toFixed(precision)) + ' ' + suffix;
             }
         }
     },

     ...

     } else if (dataType === 'consumption') {
         return this.numberFormatterFactory(Traccar.Style.numberPrecision, Strings.sharedLiterPerHourAbbreviation);
     }
```

Abyss777:
Thanks for the tip. Combined them.

tananaev:
Merged, thanks.

PR #504, opened 2017-06-15
Implement Stops report

Abyss777:
The changes are really simple. Used already implemented functionality. Selecting `Stop` shows related position.

image

Just renamed some things, because they handle not only `Event`s but also `Stop`s

### DIFF
--- a/web/app/view/ReportController.js
+++ b/web/app/view/ReportController.js
@@ -42,11 +42,14 @@ Ext.define('Traccar.view.ReportController', {
             },
             store: {
                 '#ReportEvents': {
-                    add: 'loadEvents',
-                    load: 'loadEvents'
+                    add: 'loadRelatedPositions',
+                    load: 'loadRelatedPositions'
                 },
                 '#ReportRoute': {
                     load: 'loadRoute'
+                },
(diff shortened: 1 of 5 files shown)


web/app/view/ReportController.js
```diff
@@ -525,5 +531,37 @@ Ext.define('Traccar.view.ReportController', {
...
+        xtype: 'datecolumn',
+        renderer: Traccar.AttributeFormatter.getFormatter('endTime')
+    }, {
+        text: Strings.reportDuration,
+        dataIndex: 'duration',
+        renderer: Traccar.AttributeFormatter.getFormatter('duration')
+    }, {
+        text: Strings.reportEngineHours,
+        dataIndex: 'engineHours',
+        renderer: Traccar.AttributeFormatter.getFormatter('duration')
```
Abyss777: Used `duration` formatter here instead of `hours` because it is frequently less than a hour.

tananaev:
Merged, thanks. By the way, we might want to use `duration` everywhere. People have been complaining about "hours".

Abyss777:
OK. Sure.

PR #507, opened 2017-06-19
Column filters to most windows

Abyss777:
- Added columns filters to most windows
- Added more columns to `UserDevices` and `UserGroups`
- Changed openlayers version to 4.2.0

Implemented `groupId` formatter, unfortunately we need both stores `Groups` and `AllGroups`, make it as second parameter.
Or I could replace it with other key, like `groupIdAll` for example.

### DIFF
--- a/web/app/AttributeFormatter.js
+++ b/web/app/AttributeFormatter.js
@@ -62,6 +62,18 @@ Ext.define('Traccar.AttributeFormatter', {
         return Ext.getStore('Devices').getById(value).get('name');
     },
 
+    groupIdFormatter: function (value) {
+        var group;
+        if (value !== 0) {
+            if (Ext.getStore('AllGroups').getTotalCount() === 0) {
+                group = Ext.getStore('Groups').getById(value);
+            } else {
+                group = Ext.getStore('AllGroups').getById(value);
+            }
+            return group ? group.get('name') : value;
+        }
+    },
+
     lastUpdateFormatter: function (value) {
         var seconds, interval;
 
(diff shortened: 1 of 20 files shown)


tananaev:
Why do we need both for `groupId` formatter? `AllGroups` should have everything, right?

Abyss777:
`AllGroups` request will fail for ordinary user https://github.com/tananaev/traccar/blob/master/src/org/traccar/api/resource/GroupResource.java#L48

tananaev:
Maybe we can just change it to return list of user groups instead of failing. What do you think?

Abyss777:
Then we should make the same for all other APIs for consistence ... Don't you think it looks like solving small inconvenience with changing some base things. ( I think API is some base ).

May be we can find the way on the client side only?

tananaev:
How about we check if `AllGroups` is empty first. If it is, use `Groups`.

Abyss777:
Good idea, or we can check if we are admin or manager.

tananaev:
Yes, I think either option will work.

Abyss777:
I have replaced commit with check `AllGroups` for empty.

Abyss777:
Sorry, added small optimization.

web/app/AttributeFormatter.js
```diff
@@ -62,6 +62,18 @@ Ext.define('Traccar.AttributeFormatter', {
         return Ext.getStore('Devices').getById(value).get('name');
     },
 
+    groupIdFormatter: function (value) {
+        var group;
+        if (value !== 0) {
+            if (Ext.getStore('AllGroups').getTotalCount() === 0) {
```
tananaev: I think it's better to call `getStore` once. Do something like

```js
store = Ext.getStore('AllGroups');
if (store.getTotalCount() === 0) {
    store = Ext.getStore('Groups');
}
group = store.getById(value)
return group ? group.get('name') : value;

tananaev:
Commented on the older commit, but I think it's still valid.

Abyss777:
Done

tananaev:
Merged, thanks.

PR #508, opened 2017-06-20
Implement zoom map on select

Abyss777:
- Implemented User/Server `web.selectZoom` attribute
- Zoom map view to desired value on marker (device/report) selection if it is more than current zoom.

If it is close enough leave as is, if it is too far zoom to desired value. Rather convenient I think without unneeded jumps.


tananaev:
Merged, thanks.

PR #513, opened 2017-06-27
Resource to copy (CTRL + C) the informations available in the STATUS table.

Turbovix:
The inserted code lines enable the copy (CTRL + C) of the information sent by the trackers and available in the STATUS table.

### DIFF
--- a/web/app/view/Report.js
+++ b/web/app/view/Report.js
@@ -72,6 +72,9 @@ Ext.define('Traccar.view.Report', {
     items: [{
         xtype: 'grid',
         itemId: 'grid',
+        viewConfig: {
+        enableTextSelection: true
+     },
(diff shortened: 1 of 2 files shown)


tananaev:
What does it look like? Are there any differences in the UI, or it just makes it possible to use copy/paste?

Turbovix:
Anton, there is no change in the UI. It only enables marking the information to copy via the right mouse button or via CTRL + C!

tananaev:
I think it makes sense to enable it for all grids. I suggest that we create a base class where we can configure it.

Turbovix:
screenshot_20170628_001225

tananaev:
Closing this as #603 is merged.

PR #524, opened 2017-07-13
Drivers implementation

Abyss777:
Web part for https://github.com/tananaev/traccar/pull/3349

image

Format Driver in such way: "driverUniqueId (driverName)" if name is present, otherwise uniqueId only.


web/l10n/en.json
```diff
@@ -269,6 +272,7 @@
     "eventIgnitionOff": "Ignition is OFF",
     "eventMaintenance": "Maintenance required",
     "eventTextMessage": "Text message received",
+    "eventDriverUnauthorized": "Driver Unauthorized",
```
tananaev: Do we need this event? I think instead we should have an event when a new driver has authorised. I believe in most cases this feature will be used for informational purposes only, not to detect unauthorised access.

Abyss777: May be you are right. It is very unlikely when somebody that has rfid mark take not his car and put the wrong mark on the reader.

tananaev: Exactly, but it would be nice to have event for driver change.

web/l10n/en.json
```diff
@@ -269,6 +272,7 @@
     "eventIgnitionOff": "Ignition is OFF",
     "eventMaintenance": "Maintenance required",
     "eventTextMessage": "Text message received",
+    "eventDriverChanged": "Driver Changed",
```
tananaev: I think text should be consistent with other events, which means only word should be capitalised.

Abyss777:
Replaced commit - make string consistence with other events

tananaev:
Merged, thanks.

PR #529, opened 2017-07-17
Implement Geofences column for Devices with filtering

Abyss777:
Sorry for deviation from plan, I just was interested by this topic https://www.traccar.org/forums/topic/track-only-inside-geofences/

Here is implementation of Geofences column with filtering:
image

The main problem was that `Ext.grid.filters.filter.List` used `in` operator to check if column value **in** the list. The simplest way was override `filterFn` function to check intersection of two arrays. The `ArrayListFilter` class implemented in universal way and might be used somewhere else.

Geofences are translated to names and rendered with comma.

### DIFF
--- a/web/app/AttributeFormatter.js
+++ b/web/app/AttributeFormatter.js
@@ -74,6 +74,18 @@ Ext.define('Traccar.AttributeFormatter', {
         }
     },
 
+    geofenceIdFormatter: function (value) {
+        var geofence, store;
+        if (value !== 0) {
+            store = Ext.getStore('AllGeofences');
+            if (store.getTotalCount() === 0) {
+                store = Ext.getStore('Geofences');
+            }
+            geofence = store.getById(value);
+            return geofence ? geofence.get('name') : value;
+        }
+    },
+
     driverUniqueIdFormatter: function (value) {
         var driver, store;
         if (value !== 0) {
@@ -139,6 +151,8 @@ Ext.define('Traccar.AttributeFormatter', {
             return this.deviceIdFormatter;
         } else if (key === 'groupId') {
             return this.groupIdFormatter;
+        } else if (key === 'geofenceId') {
+            return this.geofenceIdFormatter;
         } else if (key === 'lastUpdate') {
             return this.lastUpdateFormatter;
         } else if (key === 'spentFuel') {
--- a/web/app/view/ArrayListFilter.js
+++ b/web/app/view/ArrayListFilter.js
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2017 Anton Tananaev (anton@traccar.org)
+ * Copyright 2017 Andrey Kunitsyn (andrey@traccar.org)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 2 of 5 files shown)


web/app/view/ArrayListFilter.js
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+Ext.define('Traccar.view.ArrayListFilter', {
+    extend: 'Ext.grid.filters.filter.List',
+    alias: 'grid.filter.arraylist',
+
+    type: 'arraylist',
+
+    constructor: function (config) {
+        var me = this;
```
tananaev: Also, why do we need it in the first place?

web/app/view/ArrayListFilter.js
```diff
@@ -0,0 +1,46 @@
...
+        me.callParent([config]);
+        me.filter.setFilterFn(function (item) {
```
tananaev: Can you please explain how it works.

web/app/AttributeFormatter.js
```diff
@@ -74,6 +74,18 @@ Ext.define('Traccar.AttributeFormatter', {
         }
     },
 
+    geofenceIdFormatter: function (value) {
+        var geofence, store;
+        if (value !== 0) {
+            store = Ext.getStore('AllGeofences');
```
Abyss777: I've implemented it in consistence with `groupIdFormatter`

web/app/view/ArrayListFilter.js
Abyss777: `me` is used in ExtJS sources.
Hm, it did not worked without proxy variable, but now, I've removed and it is OK.

web/app/view/ArrayListFilter.js
Abyss777: Grid filters is UI above store filters. List filter perfectly fits to us, except he can not compare arrays.

Store filters can be set in two ways:
- `value` and `operator` http://docs.sencha.com/extjs/6.2.0/classic/Ext.util.Filter.html#cfg-value
http://docs.sencha.com/extjs/6.2.0/classic/Ext.util.Filter.html#cfg-operator 
Every store item (its field) compared with `value` with help of `operator`. List uses 'in' operator.
- custom `filterFn` function http://docs.sencha.com/extjs/6.2.0/classic/Ext.util.Filter.html#cfg-filterFn

`filterFn` has priority.

There is no default operator to intersect arrays, http://docs.sencha.com/extjs/6.2.0/classic/Ext.util.Filter.html#cfg-operator

because of that we should implement custom `filterFn` function that should return true if Device visible.
At least one `geofenceIds` array item contains in `value` array.

Now about function itself:

`this.getProperty()` return `dataIndex` value of the column. It can be replaced to `'geofenceIds'` directly, but component won't be universal.
`property` variable will be `device.geofenceIds` of every device
`value` is the value generated by `filter.List`  basically what user selected in dropdown menu.
Next we just look every item from one array in other.
Or if `property` not array just look in array.

Abyss777: Lets look on screenshot on device "Test 2"

"The Beginning" id = 1
"Soho" id = 5
"Broadway" id = 6

property = "Test 2".geofenceIds = [5, 6] - device in "Soho" and "Broadway"
value = filter.value = [1, 5]  - user selected "The Beginning" and "Soho"

If we used just "List" filter undercover check was array [5, 6] in array [1, 5] it is never true.
We need to check every item from first array in second.

web/app/AttributeFormatter.js
tananaev: Do we need it for groups?

Abyss777: Yes, we spent much time in optimizing it https://github.com/tananaev/traccar-web/pull/507

tananaev: I remember now, thanks.

tananaev:
Merged.

PR #531, opened 2017-07-23
Refactor managers

Abyss777:
Adjust API permissions path

Also fixed small issue, if device is added, there is small delay while server response and `geofenceIds` is undefined. Cause problem only if `Geofences` column is visible.


tananaev:
Merged, thanks.

PR #532, opened 2017-07-23
Minify.sh improvement

jon-stumpf:
In shell scripts, I avoid assuming the current working directory of where a script will be executed.  I have modified **tools/minify.sh** such that it does not need to be executed from within the **traccar-web/tools** directory.  I have modified it such that it takes one argument: the path to the **traccar-web/web** directory.  That said, the default behavior is preserved - i.e., if no argument is given, it works the same.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .project
 .settings
 .DS_Store
+/nbproject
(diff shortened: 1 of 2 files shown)


jon-stumpf:
Removed orthogonal commits.

PR #533, opened 2017-07-23
Minify improvement

jon-stumpf:
In shell scripts, I avoid assuming the current working directory of where a script will be executed.  I have modified **tools/minify.sh** such that it does not need to be executed from within the **traccar-web/tools** directory.  I have modified it such that it takes one argument: the path to the **traccar-web/web** directory.  That said, the default behavior is preserved - i.e., if no argument is given, it works the same.

### DIFF
--- a/tools/minify.sh
+++ b/tools/minify.sh
@@ -2,13 +2,20 @@
 
 cd $(dirname $0)/../web
 
-SDK="../../../ext-6.2.0"
+# Sencha Variables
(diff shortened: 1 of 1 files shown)


tananaev:
Original script does not assume any current working directory.

jon-stumpf:
I was mistaken.  It assumes that the **minify.sh** script is in a directory that is a subdirectory to the parent of the **web** directory.

PR #537, opened 2017-07-28
Do not show geofenceId if it is unknown

Abyss777:
I was thinking...
Our approach to geofences do not linked to user is simple, we do not send notifications and filter out events related to them.
But `deofencesIds` may contain such geofences. My opinion is that user should not even know about geofences he has no access. I do not think it is worth to implement filtering on server side, but we can hide them on web side.
I've changed renderer to do not show geofence id if it is unknown.

Also adjusted test API path related to https://github.com/tananaev/traccar/pull/3396
And fixed small style issue.

### DIFF
--- a/web/app/view/edit/Devices.js
+++ b/web/app/view/edit/Devices.js
@@ -164,7 +164,9 @@ Ext.define('Traccar.view.edit.Devices', {
                 if (Ext.isArray(value)) {
                     for (i = 0; i < value.length; i++) {
                         name = Traccar.AttributeFormatter.geofenceIdFormatter(value[i]);
-                        result += name ? name + ((i < value.length - 1) ? ', ' : '') : '';
+                        if (name) {
+                            result += name + ((i < value.length - 1) ? ', ' : '');
+                        }
                     }
                 }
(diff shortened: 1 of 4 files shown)


web/app/view/edit/Devices.js
```diff
@@ -160,11 +160,11 @@ Ext.define('Traccar.view.edit.Devices', {
...
             },
             renderer: function (value) {
-                var i, result = '';
+                var i, name, result = '';
                 if (Ext.isArray(value)) {
                     for (i = 0; i < value.length; i++) {
-                        result += Traccar.AttributeFormatter.geofenceIdFormatter(value[i]);
-                        result += (i < value.length - 1) ? ', ' : '';
+                        name = Traccar.AttributeFormatter.geofenceIdFormatter(value[i]);
+                        result += name ? name + ((i < value.length - 1) ? ', ' : '') : '';
```
tananaev: This is really hard to read. I suggest to use regular `if` statement.

Abyss777:
Done

tananaev:
Merged, thanks.

PR #539, opened 2017-08-01
Implement attribute to disable reports

Abyss777:
Lets start to implement features to hide some functionality in web-client.

Introduced attribute`web.disableReport` to hide report panel.

This is only feature I would make also respected by server. It will be additional protection for public use case. For example in taxi use case readonly user (customer) should not be able see history of car appointed to him. What do you think?

One comment in code...

### DIFF
--- a/web/app/controller/Root.js
+++ b/web/app/controller/Root.js
@@ -103,7 +103,7 @@ Ext.define('Traccar.controller.Root', {
     },
 
     loadApp: function () {
-        var attribution, eventId, main, i;
+        var attribution, eventId;
         Ext.getStore('Groups').load();
         Ext.getStore('Drivers').load();
         Ext.getStore('Geofences').load();
@@ -124,15 +124,7 @@ Ext.define('Traccar.controller.Root', {
         if (Traccar.app.isMobile()) {
             Ext.create('widget.mainMobile');
         } else {
-            main = Ext.create('widget.main');
-            if (Traccar.app.getAttributePreference('web.disableReport', false).toString() === 'true') {
-                for (i = 0; i < main.items.length; i++) {
-                    if (main.items.items[i].isXType('reportView')) {
-                        main.items.items[i].hide();
-                        break;
-                    }
-                }
-            }
+            Ext.create('widget.main');
         }
         eventId = Ext.Object.fromQueryString(window.location.search).eventId;
         if (eventId) {
--- a/web/app/controller/Root.js
+++ b/web/app/controller/Root.js
@@ -103,7 +103,7 @@ Ext.define('Traccar.controller.Root', {
     },
 
     loadApp: function () {
-        var attribution, eventId;
+        var attribution, eventId, main, i;
         Ext.getStore('Groups').load();
         Ext.getStore('Drivers').load();
         Ext.getStore('Geofences').load();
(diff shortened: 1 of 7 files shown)


web/app/controller/Root.js
```diff
@@ -124,7 +124,15 @@ Ext.define('Traccar.controller.Root', {
         if (Traccar.app.isMobile()) {
             Ext.create('widget.mainMobile');
         } else {
-            Ext.create('widget.main');
+            main = Ext.create('widget.main');
+            if (Traccar.app.getAttributePreference('web.disableReport', false)) {
```
Abyss777: The best way is `main.lookupReference('reportView')` but unfortunately it does not work (return `null`) here is similar problem without solution https://www.sencha.com/forum/showthread.php?330708-Event-in-child-controller-damages-lookupReference-called-in-parent-controller

We can use `main.items.items[5]` directly but I'm afraid we can forget this and change layout.
Also this loop might be reused to hide Events panel if we implement it in future.

Abyss777:
It is impossible to simply make `checkboxfield` return boolean value. Because of that unfortunately we always store boolean attributes as string. Added correctly check for both cases if value stored we as string or via API as boolean.

Abyss777:
Sorry for asking. Have you had a chance to review this PR?

web/app/controller/Root.js
```diff
@@ -124,7 +124,15 @@ Ext.define('Traccar.controller.Root', {
         if (Traccar.app.isMobile()) {
             Ext.create('widget.mainMobile');
         } else {
-            Ext.create('widget.main');
+            main = Ext.create('widget.main');
+            if (Traccar.app.getAttributePreference('web.disableReport', false).toString() === 'true') {
```
tananaev: I think this logic should be in the main layout itself.

web/app/store/ServerAttributes.js
```diff
@@ -47,5 +47,9 @@ Ext.define('Traccar.store.ServerAttributes', {
         allowDecimals: false,
         minValue: Traccar.Style.mapDefaultZoom,
         maxValue: Traccar.Style.mapMaxZoom
+    }, {
+        key: 'web.disableReport',
```
tananaev: I think we should use more generic parameter naming. Something like `ui`.

Abyss777:
You are right, logic in `Main` looks cleaner. Also renamed attribute.

web/app/view/Main.js
```diff
@@ -27,6 +27,19 @@ Ext.define('Traccar.view.Main', {
         'Traccar.view.map.Map'
     ],
 
+    initComponent: function () {
+        var i;
+        if (Traccar.app.getAttributePreference('ui.disableReport', false).toString() === 'true') {
+            for (i = 0; i < this.items.length; i++) {
+                if (this.items[i].xtype === 'reportView') {
```
tananaev: Can't we just get it by reference from here?

Abyss777: Unfortunately not, `lookupReference` also return `null` from here

tananaev: What about controller `init`? That's where we usually do things like this.

tananaev:
Merged, thanks.

Abyss777:
Sorry for did not check it myself. Lookup works.

Abyss777:
What do you think about also respect this attribute on the server side?

tananaev:
I think for now we just leave it on the client side. Don't want to add more complications on the server.

PR #543, opened 2017-08-09
Select device from deviceId url parameter

Abyss777:
Small feature for a few tasks:
- Possibility to give a link to specific device
- Always autoselect specific device (there was a couple requests)

Added alias to Root controller to distinguish its events.

I'm not sure should we or not delete url parameter like we do with `eventId`.

### DIFF
--- a/web/app/controller/Root.js
+++ b/web/app/controller/Root.js
@@ -242,7 +242,7 @@ Ext.define('Traccar.controller.Root', {
         if (first) {
             deviceId = Ext.Object.fromQueryString(window.location.search).deviceId;
             if (deviceId) {
-                device = Ext.getStore('Devices').getById(deviceId);
+                device = Ext.getStore('VisibleDevices').findRecord('id', deviceId, 0, false, true, true);
                 if (device) {
                     this.fireEvent('selectdevice', device, true);
                 }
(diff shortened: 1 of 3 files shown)


web/app/controller/Root.js
```diff
@@ -239,7 +240,16 @@ Ext.define('Traccar.controller.Root', {
             }
         }
         if (first) {
-            this.zoomToAllDevices();
+            deviceId = Ext.Object.fromQueryString(window.location.search).deviceId;
+            if (deviceId) {
+                device = Ext.getStore('VisibleDevices').findRecord('id', deviceId, 0, false, true, true);
```
Abyss777: We can not use `getById` here because it does not follow filtering.

tananaev:
Merged, thanks. Speaking of `eventId`, maybe we shouldn't remove it there either. Was there a good reason for removing `eventId` in that case?

Abyss777:
As for me it is more logically remove `eventId` and leave `deviceId`.

I think the most frequent use case of `eventId` is link from notification to particular event, after user open the link and check event, he should continue work as usual. If he refresh page or reopen browser, he should not get the event again.

But for `deviceId` I'm assuming another use case: give a link to particular device to follow it (usually it will be only device). Refreshing the page should not put out of following.

tananaev:
But what if he wants to share the event. The problem is that we modify URL after loading the page, which means that user can't just copy the URL and get the same result somewhere else. That's might be annoying because that's not how web usually works.

Anyway, I see your point, so I think we can leave it as it is for now.

PR #545, opened 2017-08-09
Add autocapitalize=none to login email input for mobile browsers

tsmgeek:
Added due to mobiles trying to capup the first letter of the email address.

Should the login api for traccar do `email.toLowerCase()` before validation?


tananaev:
Merged, thanks. This was really annoying, but for some reason I never thought of fixing it.

PR #553, opened 2017-08-17
Optimise ifs and loops

tsmgeek:
Sorry forgot to push this up.

Fixes on the markers code where there is lots of for loops on the markers doing getView(), put that into a variable speeds up Ext quite a bit, ive seen big speedups when iterating hundreds of times on my projects at work.

Added some additional eslint flags. Seems that there are some minor inconsistency across the code that are not going to be fixable with the config ie spaces on indent after variable declaration, some are 4 other 8. I guess this will just take some time.


tsmgeek:
I have a working travis doing tests but not applying fixes.
There are lots of errors. 
I can fix all the things like indents/spaces etc, but no-shadow/no-unused-vars may be a problem, what do you want me to do with those?

tananaev:
I have re-implemented linting.

tsmgeek:
I see you have eslint config in package.json, it will just read. eslint.json if it exists. Is that the new config you want to go with, it does not have crockford defaults as you had with jscs.

tananaev:
Yes, that's the config I want to go with. It's more strict rules. Also, it uses standard eslint rules and doesn't depend on any other third parties.

tsmgeek:
I thought i may be better to still have a .eslintrc.json so that editors will pick it up, I know my vscode will not read the package.json for the config. It is supported by travis.

tananaev:
Merged, thanks.

PR #555, opened 2017-08-18
Optimize calendar load

### LINKED ISSUE #554: Uint8Array usage
We are using `Uint8Array` which is not supported by all browsers. Is there any better alternative?

@Abyss777

Abyss777:
fix #554 
Used `readAsDataURL` to make `FileReader` convert to base64 itself,  but result looks like
````
data:text/calendar;base64,QkVHSU46...
````
and we need bare base64
Used `substr` to split.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,8 +17,7 @@
             "Traccar": false,
             "Strings": false,
             "Locale": false,
-            "proj4": false,
-            "Uint8Array": false
+            "proj4": false
         },
         "rules": {
             "strict": "off",
--- a/web/app/view/dialog/CalendarController.js
+++ b/web/app/view/dialog/CalendarController.js
@@ -26,12 +26,12 @@ Ext.define('Traccar.view.dialog.CalendarController', {
             reader = new FileReader();
             reader.onload = function (event) {
                 fileField.up('window').lookupReference('dataField').setValue(
-                    btoa(String.fromCharCode.apply(null, new Uint8Array(event.target.result))));
+                    event.target.result.substr(event.target.result.indexOf(',') + 1));
             };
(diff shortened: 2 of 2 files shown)


tananaev:
Looks like `readAsDataURL` is not much better:

https://developer.mozilla.org/en-US/docs/Web/API/FileReader/readAsDataURL#Browser_compatibility

I think we need to at least support IE9, ideally IE8 as well.

P.S. I'm not sure if you noticed, but we've added `eslint` support and Travis CI integration to automatically check JavaScript issues.

Abyss777:
There is no simple solution for IE<10 or we need external library or implement this https://msdn.microsoft.com/en-us/library/z9ty6h50(v=vs.85).aspx

Moreover Microsoft does not support IE<11 why we should?

tananaev:
We definitely want to support IE10. There is a lot of businesses who still use outdate IE.

OK, let's merge your pull request. Can you please remove `Uint8Array` from exceptions here:

https://github.com/tananaev/traccar-web/blob/master/package.json

tananaev:
Merged, thanks.

Abyss777:
Removed exception.
IE10 has relatively good support of HTML5. I think it would be easy to support it.

Abyss777:
Just to note that IE10 right decision:
http://docs.sencha.com/extjs/6.2.0/guides/supported_browsers.html
https://openlayers.org/en/latest/doc/tutorials/introduction.html

IE10 is oldest IE browser that can be supported by ExtJS and openlayers without additional tricks or polyfills

tananaev:
Makes sense. New website is also works only on IE10 and above.

PR #556, opened 2017-08-18
Disable some UI features

Abyss777:
Implemented a few attributes to disable some UI features.

- Drivers
- Computed Attributes
- Calendars
- `ui.disableVehicleFeatures` planed as cumulative attribute to disable related features at once. Now it disables only Drivers and some columns in reports. We might extend this in future.

Want to note that columns are hidden, they might be manually shown.


web/app/store/UserAttributes.js
```diff
@@ -83,5 +83,21 @@ Ext.define('Traccar.store.UserAttributes', {
         key: 'ui.disableReport',
         name: Strings.attributeUiDisableReport,
         valueType: 'boolean'
+    }, {
+        key: 'ui.disableVehicleFetures',
+        name: Strings.attributeUiDisableVehicleFetures,
+        valueType: 'boolean'
+    }, {
+        key: 'ui.disableDrivers',
```
tananaev: There is a lot of duplication. Can we share common attributes somewhere?

web/app/Application.js
```diff
@@ -104,6 +104,10 @@ Ext.define('Traccar.Application', {
         return window.matchMedia && window.matchMedia('(max-width: 768px)').matches;
     },
 
+    isVehicleFeaturesDisabled: function () {
```
tananaev: Name sounds a bit weird because "is" and plural. Maybe `getVehicleFeaturesDisabled`?

web/app/view/SettingsMenuController.js
```diff
@@ -55,12 +55,15 @@ Ext.define('Traccar.view.SettingsMenuController', {
             this.lookupReference('settingsGroupsButton').setHidden(false);
             this.lookupReference('settingsGeofencesButton').setHidden(false);
             this.lookupReference('settingsNotificationsButton').setHidden(false);
-            this.lookupReference('settingsCalendarsButton').setHidden(false);
-            this.lookupReference('settingsDriversButton').setHidden(false);
+            this.lookupReference('settingsCalendarsButton').setHidden(
+                Traccar.app.getBooleanAttributePreference('ui.disableCalendars'));
+            this.lookupReference('settingsDriversButton').setHidden(Traccar.app.isVehicleFeaturesDisabled() ||
```
tananaev: This looks weird. Please move all parameters to the next line.

tananaev: Same for all other similar places.

Abyss777:
- Moved common known attributes to two separate stores to avoid duplication
- Combine them in `Root.js`
- Moved pointed checks on next line
- Renamed `getVehicleFeaturesDisabled` function

tananaev:
Merged, thanks.

PR #557, opened 2017-08-18
Update openlayers to 4.3.1

Abyss777:
There was a couple on new openlayers releases.
Analyzed release notes https://github.com/openlayers/openlayers/releases/tag/v4.3.0
Only one touches us. Make it linear without callbacks.


tananaev:
Merged, thanks.

PR #559, opened 2017-08-22
Keep alarms until deleated

BernardtB:
Hi 
Is it possible to add the feature to Traccar , where all the alarms stays on the server until its deleted ,
and to synchronize all the alarms with all the current sessions that is logged into the same account .

So by login you can view all the alarms that the system stored when you were not logged in 
and then logging out without clearing the alarms , so when you login again you can still see the alarms until deleted


tananaev:
What the hell is this? I'm closing this nonsense pull request. Not sure what you tried to do.

BernardtB:
My mistake , I do apologize .

PR #566, opened 2017-08-30
Add Max Zoom attribute

### LINKED ISSUE #521: feature rqst : set max zoom level
Hi,

I'm using traccar this year to follow a competition where the pilots don't know where they have to fly before takeoff. But as they take off in a row and not all together, the live-tracking could give the next ones some advice about the track to follow.

So as far as I can imagine, is it possible to : 
- disable any zooming above a certain value in order to limit the vieable area to a limited detailed map ?
- or... maybe, but it's more difficult... put some adjustable delay to the tracks, so that the actual position is not displayed, but will be desplayed with a delay of 10 minutes, or one/two hours ?

Thanks
Michel

Abyss777:
Implemented `web.maxZoom` attribute to limit zooming for user/server

fix https://github.com/tananaev/traccar/issues/3489
fix #521


tananaev:
Merged, thanks.

PR #567, opened 2017-08-30
Add some missed command related strings

Abyss777:
Added missed types and reordered them to repeat order from https://github.com/tananaev/traccar/blob/master/src/org/traccar/model/Command.java


tananaev:
Merged, thanks.

PR #573, opened 2017-09-06
Move units to attributes and add volume units

Abyss777:
Web part for https://github.com/tananaev/traccar/pull/3507

Some comments in code


web/l10n/en.json
```diff
@@ -64,6 +68,7 @@
     "sharedTypeNumber": "Number",
     "sharedTypeBoolean": "Boolean",
     "sharedTimezone": "Timezone",
+    "sharedVolumeUnit": "Volume",
```
Abyss777: May be we should add "Unit" word to strings, they look not clearly
image

web/app/view/dialog/AttributeController.js
```diff
@@ -53,35 +53,72 @@ Ext.define('Traccar.view.dialog.AttributeController', {
...
+                            default:
+                                break;
+                        }
+                    } else {
+                        config.xtype = 'textfield';
+                    }
+                    break;
             }
-            if (valueField.getXType() !== config.xtype || valueField.convert !== config.convert) {
+            if (valueField.getXType() !== config.xtype) {
```
Abyss777: Handle switching type more correctly here and next

web/app/store/VolumeUnits.js
```diff
@@ -0,0 +1,57 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+Ext.define('Traccar.store.VolumeUnits', {
+    extend: 'Ext.data.Store',
+    fields: ['key', 'name', 'fullName', 'factor'],
```
Abyss777: Added `fullName` fields to distinguish gallons
image

May be we should do the same for distance and speed units

web/l10n/en.json
tananaev: Agree.

Abyss777:
Added "Unit" to strings. Chose "settings" prefix to save one string from translation.
image

Looks better now

tananaev:
Merged, thanks.

PR #574, opened 2017-09-07
Save attributes with correct types

Abyss777:
Now attribute type is according to component (numderfield, checkbox, textfield) where it was created.

Also added `scrollable: true` to `Users` and `Statistics` windows, because buttons do not fit in vertical mobile view (720x1280).

And also removed meaningless `allowBlank: false` from checkboxes


tananaev:
Merged, thanks.

PR #582, opened 2017-09-11
Fix reports with missed values

Abyss777:
Chart generating brakes if some attributes missed in the set https://www.traccar.org/forums/topic/graphic-report-issue/
as well as route grid (just warning, but...)

ExtJS correctly handle missed values, but they should be `undefined`, not `null`

### DIFF
--- a/web/app/controller/Root.js
+++ b/web/app/controller/Root.js
@@ -52,7 +52,7 @@ Ext.define('Traccar.controller.Root', {
         if (value !== undefined) {
             return Traccar.AttributeFormatter.getAttributeConverter(this.attributeKey)(value);
         }
-        return null;
+        return value;
     },
 
     onLaunch: function () {


Abyss777:
By the way, they are rendered as `gap` by default http://docs.sencha.com/extjs/6.2.0/classic/Ext.chart.series.Line.html#cfg-nullStyle
Do not know, why it is called null

web/app/controller/Root.js
```diff
@@ -52,7 +52,7 @@ Ext.define('Traccar.controller.Root', {
         if (value !== undefined) {
             return Traccar.AttributeFormatter.getAttributeConverter(this.attributeKey)(value);
         }
-        return null;
+        return undefined;
```
tananaev: Undefined value shouldn't be used explicitly. I propose we remove return statement instead.

Abyss777:
Oh, sure. Replaced commit.

Abyss777:
lint is failed than
````
/home/local/RG/kas/repos/traccar/traccar-web/web/app/controller/Root.js
  50:25  error  Expected to return a value at the end of method 'calculateAttribute'  consistent-return
````

tananaev:
I know the solution. Just return `value`.

Abyss777:
Good. Thanks for the tip.

tananaev:
Merged, thanks.

PR #586, opened 2017-09-13
Make keyboard zoom and pan working

### LINKED ISSUE #584: Zoom on the map using the numeric keypad
Hello Anton,

Is it possible to configure in a future release the zoom on the different map (Google, OpenStreeMap, ...) via + and - on the numeric keypad ?

Thank you.

Abyss777:
fix #584 

Actually it is enabled by default https://openlayers.org/en/latest/apidoc/ol.interaction.KeyboardZoom.html and even working if user click on "i" or "+ -" controls.
The problem is that `div` containing map is not "focusable". This change makes it "focusable"

Now user can click on any point of map and zoom and pan will work.

### DIFF
--- a/web/app/view/map/BaseMap.js
+++ b/web/app/view/map/BaseMap.js
@@ -175,6 +175,8 @@ Ext.define('Traccar.view.map.BaseMap', {
             view: this.mapView
         });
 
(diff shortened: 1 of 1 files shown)


tananaev:
Merged, thanks.

PR #588, opened 2017-09-19
Implement Notifications per device

Abyss777:
Web part for https://github.com/tananaev/traccar/pull/3534

image

And move all link buttons to submenu
image

Some comments in code


web/app/view/edit/Users.js
```diff
@@ -121,20 +121,24 @@ Ext.define('Traccar.view.edit.Users', {
         }, {
             text: Strings.userAdmin,
             dataIndex: 'admin',
+            renderer: Traccar.AttributeFormatter.getFormatter('admin'),
```
Abyss777: Format boolean columns as translated "Yes/No" instead of just "true/false". Same in other places.

web/app/view/edit/DevicesController.js
```diff
@@ -143,15 +98,15 @@ Ext.define('Traccar.view.edit.DevicesController', {
...
         deviceReadonly = Traccar.app.getPreference('deviceReadonly', false) && !Traccar.app.getUser().get('admin');
         readonly = Traccar.app.getPreference('readonly', false) && !Traccar.app.getUser().get('admin');
         empty = selected.length === 0;
         this.lookupReference('toolbarEditButton').setDisabled(empty || readonly || deviceReadonly);
         this.lookupReference('toolbarRemoveButton').setDisabled(empty || readonly || deviceReadonly);
-        this.lookupReference('toolbarGeofencesButton').setDisabled(empty || readonly);
-        this.lookupReference('toolbarAttributesButton').setDisabled(empty || readonly);
-        this.lookupReference('toolbarDriversButton').setDisabled(empty || readonly);
+        deviceMenu = this.lookupReference('toolbarDeviceMenu');
+        deviceMenu.device = empty ? null : selected[0];
```
Abyss777: Passing here device object to `deviceMenu`

web/app/view/DeviceMenuController.js
```diff
@@ -0,0 +1,104 @@
...
+
+    onGeofencesClick: function () {
+        Ext.create('Traccar.view.BaseWindow', {
+            title: Strings.sharedGeofences,
+            items: {
+                xtype: 'linkGeofencesView',
+                baseObjectName: 'deviceId',
+                linkObjectName: 'geofenceId',
+                storeName: 'Geofences',
+                baseObject: this.getView().ownerCmp.device.getId()
```
Abyss777: There is intermediate component to access it we use `ownerCmp`, frankly it should be `ownerCt` but... http://docs.sencha.com/extjs/6.2.0/classic/Ext.Component.html#property-ownerCt

web/l10n/en.json
```diff
@@ -340,6 +340,7 @@
     "alarmTampering": "Tampering Alarm",
     "alarmRemoving": "Removing Alarm",
     "notificationType": "Type of Notification",
+    "notificationAlways": "Send Always",
```
tananaev: I think we should rename label because it's a bit confusing. Am I understanding correctly that this flag means that notification works for all user devices? If yes, we should probably use "All Devices" or something like this for the label. Internally we can still call it "always", unless we can think of better name.

web/app/view/edit/NotificationsController.js
```diff
@@ -0,0 +1,32 @@
...
+
+    requires: [
+        'Traccar.view.dialog.Notification',
+        'Traccar.model.Notification'
+    ],
+
+    objectModel: 'Traccar.model.Notification',
+    objectDialog: 'Traccar.view.dialog.Notification',
+    removeTitle: Strings.sharedNotification
+
```
tananaev: Remove space here to make it consistent with other JS classes.

web/app/view/edit/DevicesController.js
```diff
@@ -143,15 +98,15 @@ Ext.define('Traccar.view.edit.DevicesController', {
...
+        this.lookupReference('toolbarDeviceMenu').setDisabled(empty);
```
tananaev: Do we need second `lookupReference` here for the same component?

web/app/view/DeviceMenuController.js
tananaev: Maybe we should use `up` method instead?

Abyss777:
Fixed comments

How about "Apply to All Devices" ?

image

tananaev:
It's too verbose and wraps by default. What's wrong with just "All Devices"?

Abyss777:
Can't explain. Just feeling that it looks not clear. Anyway changed

image

tananaev:
Merged, thanks.

Abyss777:
What do you think about moving "Device Distance" menu item from Settings Menu to Device Menu?

tananaev:
I guess it makes sense now that we have submenu in devices.

PR #589, opened 2017-09-20
Move Device Distance to Device Menu

Abyss777:
image

Also adjusted API paths to https://github.com/tananaev/traccar/pull/3538


tananaev:
Merged, thanks.

PR #595, opened 2017-09-28
Add altitude to charts list

Abyss777:
It is not very useful for cars but might be useful for flying vehicles.

image


tananaev:
Merged,  thanks.

PR #596, opened 2017-09-29
Rewrote tools/minify.sh;

jon-stumpf:
I was not familiar with Sencha. In learning how it works and how it installs, I rewrote the shell script based on a common structure that I have used throughout my career.  If you find it useful, please feel free to merge.

### DIFF
--- a/tools/minify.sh
+++ b/tools/minify.sh
@@ -1,14 +1,62 @@
 #!/bin/sh
 
-cd $(dirname $0)/../web
+# Sencha Variables
+if [ -z "${SENCHAROOT}" ]
+then
+	SENCHAROOT="../../.."
+fi
 
-SDK="../../../ext-6.2.0"
+SDK="${SENCHAROOT}/ext-6.2.0"
+CORE="${SDK}/packages/core"
+CLASSIC="${SDK}/classic/classic"
+CMDDIR="${SENCHAROOT}/Cmd"
 
-sencha compile --classpath=app.js,app,$SDK/packages/core/src,$SDK/packages/core/overrides,$SDK/classic/classic/src,$SDK/classic/classic/overrides \
-       exclude -all \
-       and \
-       include -recursive -file app.js \
-       and \
-       exclude -namespace=Ext \
-       and \
-       concatenate -closure app.min.js
+if [ ${#} -gt 1 ]
+then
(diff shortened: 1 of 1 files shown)


tananaev:
Unfortunately it doesn't work like the old script. I can run the old one from anywhere. For example from Traccar root:
```
traccar-web/tools/minify.sh
```

jon-stumpf:
I thought I handled that case.  I will fix it.

jon-stumpf:
Please provide me the context that you experienced the error.  I reran my tests and I am able to run the command from anywhere.  I believe the error you are experiencing has to do with the location of SENCHAROOT and I will need to check some other things in the script.

Please provide me the PWD where you run the script and the absolute path of where your SENCHAROOT is.

Thanks.

tananaev:
I didn't specify any locations. I took your file and ran command that I provided earlier.

jon-stumpf:
I understand that you did not specify a location.  I modeled the condition and ran a different test.  I believe I covered the default behavior correctly.  Please try again.

tananaev:
Now the problem is with CMD path:
```
$ tools/minify.sh 
tools/minify.sh: line 59: tools/../../../Cmd/sencha: No such file or directory
```

I think I'll just leave existing script for now. We can keep pull request open in case someone else want to use more advanced script.

jon-stumpf:
This is why I don't like relative paths.  :)

I see that I made an assumption that is not valid in your default case.  I install third-party software under /opt and use very specific rules from the LSB.  I eliminated the assumption.

If you would just indulge me, I would like to understand what I am not understanding about how you installed Ext.js and Sencha Cmd.  Can you tell me the following paths in your environment?

1. The absolute path of **traccar-web/tools/minify.sh**.
2. The absolute path of the Sencha **ext-6.2.0** directory.
3. The result of $(**which sencha**).

Thank you.

tananaev:
I only installed CMD. It's in my PATH. ExtJS itself is just a folder next to Traccar repo folder.
```
$ which sencha
/Users/tananaev/bin/Sencha/Cmd/sencha
```

jon-stumpf:
Thanks.  My last commit assumes that "**sencha**" is in the PATH rather than use an absolute path.

PR #599, opened 2017-10-06
Add quick period selector to Report Config dialog

### LINKED ISSUE #536: Feature Request: Add buttons to report generator
Прошу прощения, в английском не силен. Прошу добавить кнопки "Сегодня", "Вчера", "Текущая неделя", "Текущий месяц" и т.п. в диалог "Конфигурировать" отчета "Маршрут". Не удобно каждый раз выставлять дату-вермя чтобы посмотреть пройденный маршрут.

Abyss777:
I think it is reasonable to set `to` to the end of period, but not `now`. Because user do not have to reconfigure period a couple of hours later.

image

Now, a new periods can be easily added in future if demanded. (like last week or last month)

fix #536


tananaev:
I have a suggestion. How about we hide from and to and only show it if you select "custom" in the period drop-down. What do you think? It would make UI much simpler.

Abyss777:
Agree. Good idea. What will be default value? Custom?

Abyss777:
Hm, looks like it is better without default value.

Abyss777:
On first start everything is empty

image

On Custom select fields are shown

image

On predefined period select they are hidden

image


Also added "Previous Week" and "Previous Month" to not come back to it again

tananaev:
Just one more small comment about UI. I think it would be better to have period above from and to fields.

web/app/view/dialog/ReportConfigController.js
```diff
@@ -44,7 +44,55 @@ Ext.define('Traccar.view.dialog.ReportConfigController', {
...
+                from.setDate(1);
+                to.setDate(1);
+                to.setMonth(from.getMonth() + 1);
+                break;
+            case 'previousMonth':
+                from.setDate(1);
+                from.setMonth(from.getMonth() - 1);
+                to.setDate(1);
+                break;
+            default:
```
tananaev: Please add break here.

web/app/view/dialog/ReportConfigController.js
```diff
@@ -44,7 +44,55 @@ Ext.define('Traccar.view.dialog.ReportConfigController', {
...
+            case 'today':
+                to.setDate(to.getDate() + 1);
+                break;
+            case 'yesterday':
+                from.setDate(to.getDate() - 1);
+                break;
+            case 'thisWeek':
+                day = from.getDay();
+                first = from.getDate() - day + (day === 0 ? -6 : 1);
+                from.setDate(first);
```
tananaev: What if it crosses a month boundary? Does it work?

web/app/view/dialog/ReportConfigController.js
```diff
@@ -44,7 +44,55 @@ Ext.define('Traccar.view.dialog.ReportConfigController', {
...
+                break;
+            default:
+        }
+        from.setHours(0, 0, 0, 0);
+        to.setHours(0, 0, 0, 0);
+        this.lookupReference('fromDateField').setValue(from);
+        this.lookupReference('fromTimeField').setValue(from);
+        this.lookupReference('toDateField').setValue(to);
+        this.lookupReference('toTimeField').setValue(to);
+        this.lookupReference('fromContainer').setHidden(true);
```
tananaev: I think it's better to do something like `newValue !== 'custom'` here, and remove "custom" case from switch to make it easier to read the code.

Abyss777:
It works correctly with a month boundary

image

There was `return` in `'custom'` case, we should not change anything, just unhide. Moved logic in `if`

tananaev:
Perfect, thanks.

PR #602, opened 2017-10-16
Make liveRoute layer initial visible if needed  

### LINKED ISSUE #601: Live routes don't show when button pressed by default
Anytime you log in to Traccar and the Live routes button is already pressed, the live routes won't show until you "unpress" and press it again, making the "pressed by default" useless and misleading user, that may think there is no device moving.

Abyss777:
liveRoute layer was not visible until press button again

fix #601


tananaev:
Merged, thanks.

PR #603, opened 2017-10-17
Make cell text "selectable" in all grids

Abyss777:
Looks like #513 abandoned...

Implement base Grid Panel class with common configs:
- `enableTextSelection: true`
-  `plugins: 'gridfilters'` - it is used in most grids

Applied workaround for selection in windows https://stackoverflow.com/questions/42760943/workaround-for-bug-extjs-22715-enabletextselection-true-has-no-effect-inside-wi

### DIFF
--- a/web/app/view/Events.js
+++ b/web/app/view/Events.js
@@ -17,7 +17,7 @@
  */
 
 Ext.define('Traccar.view.Events', {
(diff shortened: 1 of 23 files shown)


tananaev:
Merged, thanks.

PR #605, opened 2017-10-20
Adapted to use buffered commands

Abyss777:
Adapted to the server side and fixed one issue: `typesStore` initially not loaded.


tananaev:
Merged.

PR #608, opened 2017-10-23
Enable word wrapping in State panel

### LINKED ISSUE #607: Multiline State Panel cells
Actually making cell "selectable" #603 did not help initial problem (impossible read long data in grid cells)
Only visible data are selectable, cut strings are cut in clipboard. 

What do you think of making state grid multiline?

image

Abyss777:
It is impossible to read long values in State Panel, enabled word wrapping in "Value" column.
fix #607


tananaev:
Merged, thanks.

PR #621, opened 2017-11-13
Add "Show Address" link in State panel

Abyss777:
If address is empty it is replaced with link to retry geocoding.

image

Some comments in code


.eslintrc.json
```diff
@@ -16,7 +16,7 @@
         "max-lines": ["error", 1000],
         "max-len": ["error", 180],
         "max-statements": ["error", 35],
-        "max-params": ["error", 5],
+        "max-params": ["error", 7],
```
Abyss777: We need seven parameter in `onCellClick` handler

web/app/view/State.js
```diff
@@ -63,9 +63,15 @@ Ext.define('Traccar.view.State', {
             renderer: function (value, metaData, record) {
                 if (record.get('attribute') === 'alarm') {
                     metaData.tdCls = 'view-color-red';
+                } else if (record.get('name') === Strings.positionAddress && !value) {
+                    return '<a href="#">' + Strings.stateShowAddress + '</a>';
```
Abyss777: It is really hard to add `onclick` handler here. (or my knowledge is not enough)
But found on sencha forums tip to handle cell click instead of `a` element.

web/app/view/StateController.js
```diff
@@ -184,5 +184,30 @@ Ext.define('Traccar.view.StateController', {
             this.position = null;
             Ext.getStore('Attributes').removeAll();
         }
+    },
+
+    onCellClick: function (grid, td, cellIndex, record, tr, rowIndex, e) {
+        var positionId = this.position.getId();
+        if (record.get('name') === Strings.positionAddress && !record.get('value') &&
+                e.target.tagName.toLowerCase() === 'a') {
```
Abyss777: We need to determine if user clicked on link (not other place of cell) to leave behavior intuitive.

web/app/view/StateController.js
```diff
@@ -184,5 +184,30 @@ Ext.define('Traccar.view.StateController', {
...
+            Ext.Ajax.request({
+                scope: this,
+                method: 'GET',
+                url: 'api/server/geocode',
+                params: {
+                    latitude: this.position.get('latitude'),
+                    longitude: this.position.get('longitude')
+                },
+                success: function (response) {
+                    if (record && this.position && positionId === this.position.getId()) {
```
Abyss777: This is protection in case request completed after user clicked other position or deselected position.

web/app/view/StateController.js
tananaev: Can we assign some class to the element instead?

Abyss777:
I've replaced the commit because I found better solution with help of global events.
Also added the same functionality to route reports

image

web/app/view/State.js
```diff
@@ -63,6 +63,9 @@ Ext.define('Traccar.view.State', {
             renderer: function (value, metaData, record) {
                 if (record.get('attribute') === 'alarm') {
                     metaData.tdCls = 'view-color-red';
+                } else if (record.get('name') === Strings.positionAddress && !value) {
+                    return '<a href="#" onclick="Ext.fireEvent(\'stategeocode\')" >' +
+                    Strings.stateShowAddress + '</a>';
```
tananaev: There is an indentation problem here.

web/app/view/ReportController.js
```diff
@@ -455,7 +481,14 @@ Ext.define('Traccar.view.ReportController', {
     }, {
         text: Strings.positionAddress,
         dataIndex: 'address',
-        renderer: Traccar.AttributeFormatter.getFormatter('address')
+        renderer: function (value, metaData, record) {
+            if (!value) {
+                return '<a href="#" onclick="Ext.fireEvent(\'routegeocode\', ' +
+                record.getId() + ')" >' +
```
tananaev: Here as well.

web/l10n/en.json
```diff
@@ -241,6 +241,7 @@
     "stateTitle": "State",
     "stateName": "Attribute",
     "stateValue": "Value",
+    "stateShowAddress": "Show Address",
```
tananaev: I suggest to rename it to `shared*` because it's used in reports as well.

web/app/view/ReportController.js
```diff
@@ -423,6 +426,29 @@ Ext.define('Traccar.view.ReportController', {
...
+                method: 'GET',
+                url: 'api/server/geocode',
+                params: {
+                    latitude: position.get('latitude'),
+                    longitude: position.get('longitude')
+                },
+                success: function (response) {
+                    position.set('address', response.responseText);
+                    position.commit();
+                    this.fireEvent('selectReport', position);
```
tananaev: Why do we need to fire this event?

Abyss777: Clicking on "Show Address" causing selecting the row that causing selecting position and show it in State panel, when comes response address is replaced in Report panel, but not in State.
Looks not very nice. Just updating State panel here.

tananaev: I see. OK.

tananaev:
Merged, thanks.

Abyss777:
Sorry for indentation. Pity `eslint` not supports such indentations.

PR #626, opened 2017-12-07
Update openlayers to 4.6.1

Abyss777:
Updated openlayers to recently released
`exceedLength` was deprecated (so fast) and renamed to `overflow` https://github.com/openlayers/openlayers/releases/tag/v4.6.0


tananaev:
Merged, thanks.

PR #634, opened 2017-12-28
Implement disabling devices

Abyss777:
Here is UI for device disable.
- Checkbox in device model
- Columns in devices panel and permissions panel
- Make disabled devices visually different in devices panel
image


web/app.css
```diff
@@ -8,6 +8,10 @@
     background-color: rgba(255, 84, 104, 0.3);
 }
 
+.view-item-disabled {
+	opacity: 0.5;
```
tananaev: Indentation problem here.

web/app/view/dialog/DeviceController.js
```diff
@@ -0,0 +1,29 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+Ext.define('Traccar.view.dialog.DeviceController', {
+    extend: 'Traccar.view.dialog.BaseEditController',
+    alias: 'controller.device',
+
+    init: function () {
+        if (Traccar.app.getUser().get('admin')) {
+            this.lookupReference('disabledField').setHidden(false);
```
tananaev: Do we have a check on the server side to make sure regular user can't enable or disable his devices?

web/app/view/edit/Devices.js
```diff
@@ -81,11 +81,14 @@ Ext.define('Traccar.view.edit.Devices', {
     viewConfig: {
         enableTextSelection: true,
         getRowClass: function (record) {
-            var status = record.get('status');
+            var status = record.get('status'), result = '';
```
tananaev: I think it's better to swap this around. I prefer longer assignment at the end so it's easier to read the code.

web/app/view/permissions/Devices.js
```diff
@@ -57,6 +57,14 @@ Ext.define('Traccar.view.permissions.Devices', {
             minWidth: Traccar.Style.columnWidthNormal,
             hidden: true,
             filter: 'string'
+        }, {
+            text: Strings.sharedDisabled,
+            dataIndex: 'disabled',
+            flex: 1,
```
tananaev: Can we separate some config into defaults here?

web/app/view/dialog/DeviceController.js
Abyss777: Regular user do not have access to disabled devices at all. `checkDevice` uses `getUserItems` function that filters disabled devices.
https://github.com/tananaev/traccar/blob/master/src/org/traccar/database/PermissionsManager.java#L302

web/app/view/permissions/Devices.js
Abyss777: Unfortunately not. `default` config also applied to "checkcolumn"  and it will have same width as others. That do not looks good.

web/app/view/dialog/DeviceController.js
tananaev: OK, I guess it good enough.

web/app/view/permissions/Devices.js
tananaev: Right, I remember now.

Abyss777:
Fixed indentation and swapped variables around.

tananaev:
Merged, thanks.

PR #642, opened 2018-01-17
Implement POI layer

### LINKED ISSUE #173: KML overlay
Is there some way I can put an overlay to the maps? Like linestrings in the Google KML?

Abyss777:
Implemented POI layer, it can be used for adding static objects on the map or any other features supported by KML and openlayers.

image

image

Here green line is KML LineString.

I would share this example `test.kml` but not sure where to put it. (appropriate folder)

fix #173

### DIFF
--- a/web/app/model/Server.js
+++ b/web/app/model/Server.js
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 Anton Tananaev (anton@traccar.org)
+ * Copyright 2015 - 2018 Anton Tananaev (anton@traccar.org)
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -60,6 +60,9 @@ Ext.define('Traccar.model.Server', {
         type: 'boolean'
     }, {
         name: 'coordinateFormat',
+        type: 'string'
+    }, {
+        name: 'poiLayer',
         type: 'string'
     }, {
         name: 'attributes'
--- a/web/app/model/User.js
+++ b/web/app/model/User.js
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 - 2017 Anton Tananaev (anton@traccar.org)
+ * Copyright 2015 - 2018 Anton Tananaev (anton@traccar.org)
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -80,6 +80,9 @@ Ext.define('Traccar.model.User', {
     }, {
         name: 'limitCommands',
         type: 'boolean'
+    }, {
+        name: 'poiLayer',
+        type: 'string'
     }, {
         name: 'token',
         type: 'string'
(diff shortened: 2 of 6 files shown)


tananaev:
I merged server side, but then thought maybe it's better to allow users to upload KML files kind of like calendar data?

Abyss777:
There are some disadvantages of such solution
- Some limits of file size uploads
- Impossible to use public POIs, they must be uploaded in server, but they can be large.
- Impossible to share same set of POIs/tracks (only duplicating) between some users (for purposes described in https://github.com/tananaev/traccar/issues/3714)
- I believe it is tricky to point openlayers to binary/base64 data as source for layer.

The only advantage I see is user do not need use some file hosting.

We uploading calendar data because we use it on backend side.

tananaev:
I see. Makes sense if we can use external hosting for this. Have you tested that it works with some common file hosting?

Abyss777:
hm... there is CORS issue, I'll lookup some workaround

Abyss777:
Looks like it is big problem. There is no workaround on client side. Server (with kml file) should allow cross origin requests. I.e. Google Disk, Cloud Mail.Ru and pastebin do not do it.

But it works with Github Gists for example

tananaev:
One option I can think of is to proxy URL through our server, but it adds complexity and maybe we should just find some good options that provide CORS to recommend users.

Abyss777:
I've tested a couple of free web-hosting, they do not allow CORS. May be some others, but it is impossible to test all.

I think it is possible provide https://github.com/Rob--W/cors-anywhere for paid users. I checked https://cors-anywhere.herokuapp.com/ and it works correctly with links from Google Drive. Just need to add link after `/` like this `http://cors-anywhere.herokuapp.com/https://drive.google.com/a/fox5.ru/uc?authuser=0&id=1gc5gdu2xfX8v72cghDtBU4AQgMAzBbdp&export=download`

I would not recommend it for production, it is just demo.

The only simple 100% working solution I've found is Github.

I though about another thing. We do not have a way to access to photos/audio received from devices. May be we should implement some unified solution for access to photos and upload KML files...

Abyss777:
I think it is possible hardcode in this proxy in webapp and it will be transparent for user.

tananaev:
Merged. I think it's good enough for now. I'm not sure that proxy is a good idea. As for media files, maybe it's something we should look into next. It should be fairly easy to give access to those and then we can use from from the UI.

Abyss777:
Thanks.

Attached `test.kml` example here for history.

[test.zip](https://github.com/tananaev/traccar-web/files/1641653/test.zip)

lingan-se:
Hello. This is a very nice implementation. I have tried it a little but are confused. Must you have the *.kml file in /opt/traccar/web folder or can you have it linked from another site example: http://somesite.com/racetrack.kml ? Another question is if I have to put the *.kml path in both "user - account" and "server - settings" (web interface). Is it possible to get the *.kml to "http://mytraccarserver.com/simple"?
I admire your fantastic work with this. Thanks!

Camtek:
or should I put this file before I can call it in the web interface?

PR #643, opened 2018-01-23
Implement scheduled notifications

Abyss777:
Added Calendar field to Notification model


tananaev:
This looks good. I will merge it when the server side is ready as well. It's good that we found more applications for calendar functionality.

tananaev:
Merged.

Abyss777:
Agree, calendars was looking lonely.

PR #644, opened 2018-02-03
Odometer permission to both Admin or Manager

parveenkumardeeva:
Already added to server side code
https://github.com/tananaev/traccar/pull/3709

### DIFF
--- a/web/app/view/DeviceMenuController.js
+++ b/web/app/view/DeviceMenuController.js
@@ -36,7 +36,7 @@ Ext.define('Traccar.view.DeviceMenuController', {
             Traccar.app.getBooleanAttributePreference('ui.disableComputedAttributes'));
         this.lookupReference('menuCommandsButton').setHidden(Traccar.app.getPreference('limitCommands', false));
         this.lookupReference('menuDeviceDistanceButton').setHidden(
-            !(Traccar.app.getUser().get('admin') || Traccar.app.getUser().get('userLimit') !== 0) || Traccar.app.getVehicleFeaturesDisabled());
+            !Traccar.app.getUser().get('admin') && Traccar.app.getUser().get('userLimit') === 0 || Traccar.app.getVehicleFeaturesDisabled());
     },
 
     onGeofencesClick: function () {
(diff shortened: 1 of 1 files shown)


web/app/view/DeviceMenuController.js
```diff
@@ -36,7 +36,7 @@ Ext.define('Traccar.view.DeviceMenuController', {
             Traccar.app.getBooleanAttributePreference('ui.disableComputedAttributes'));
         this.lookupReference('menuCommandsButton').setHidden(Traccar.app.getPreference('limitCommands', false));
         this.lookupReference('menuDeviceDistanceButton').setHidden(
-            !Traccar.app.getUser().get('admin') || Traccar.app.getVehicleFeaturesDisabled());
+            !(Traccar.app.getUser().get('admin') || Traccar.app.getUser().get('userLimit') !== 0) || Traccar.app.getVehicleFeaturesDisabled());
```
tananaev: Why do you need extra brackets?

parveenkumardeeva: Dear @tananaev ,

Just wanted to be sure that If an admin have userlimit = 0 then he should also be able to set device odometer.
BTW, Pardon me if it's not optimized acc. to javascript design because It's my first block of js code.

tananaev: I'm not asking about "or" operation. I'm asking about extra brackets that are unnecessary in any programming language I know in this situation. Correct me if I'm wrong.

parveenkumardeeva: If we put this code:
!Traccar.app.getUser().get('admin') || !Traccar.app.getUser().get('userLimit') !== 0

parveenkumardeeva: If we put this code:
!Traccar.app.getUser().get('admin') || Traccar.app.getUser().get('userLimit') == 0
then after checking first condition for admin it will return false and then it will check for the second condition and if an admin have userlimit set to 0 it will return true and the button will be disabled. So, better return from the brackets after checking for an admin

Did I understood something wrong?

tananaev: Your condition
```js
!(Traccar.app.getUser().get('admin') || Traccar.app.getUser().get('userLimit') !== 0)
```
is the same as
```js
!Traccar.app.getUser().get('admin') && Traccar.app.getUser().get('userLimit') === 0
```

parveenkumardeeva: Ok, Now I see.

tananaev:
Merged, thanks.

PR #645, opened 2018-02-05
Format image and audio attributes as link in State panel

Abyss777:
Make a simple formatting in State panel
It is just a link to open file in new browser tab/window

image


tananaev:
Merged, thanks.

PR #650, opened 2018-02-14
Add Decoder Timezone attribute

Abyss777:
Added Decoder Timezone attribute to device and group.
Used `sharedTimezone` string, do not think we need separate.


tananaev:
Merged, thanks. Yes, I agree that we don't need another string.

PR #655, opened 2018-03-19
Add Calendar column to some grids

Abyss777:
- Added column Calendar (hidden by default) to grids with Geofences and Notifications ( fix https://github.com/tananaev/traccar/issues/3812)
image

- Moved Calendar to Extra fieldset (make it consistent with Geofence and logic)
image


tananaev:
Merged, thanks.

PR #659, opened 2018-04-06
Adding search bar to map

cyrix2000:
Adding search bar using Nominatim geocoder with OSM


tananaev:
1. Formatting is way off everywhere. It would need to be fixed if you want it to be included in Traccar.
2. Where does `Geocoder` class come from?

dudududodododedede:
Any chance this could be updated to work with the latest version of traccar-web @cyrix2000 ?

alexthefifth:
Updated to work with latest traccar (4.13 as of this writing): https://github.com/traccar/traccar-web/pull/860

PR #660, opened 2018-04-06
Add ol-popup and ol-geocoder libs

cyrix2000:
Needed ressources for showing search bar in map window


cyrix2000:
It belongs to #659 which adds a search functionality to the Traccar map.

tananaev:
Why is it a separate pull request? If it belongs to other PR it should be there.

Issue #652, opened 2018-03-12
Feature Request: Add Odometer to Summary, Trips, and Stops Reports


ckrey:
Use case:

Legal requirements for commercially used vehicles require a vehicle log including
Start and End values for Odometer readings in most countries.

Please add 
- TotalDistance
- Odometer
- OBDOdometer
values for start and end positions to
- SummaryReport
- TripsReport
- StopsReport

Examples:
http://www.ird.govt.nz/resources/4/7/474499e8-8e4c-47c0-ab3d-54334ab8c0f1/vehiclelogbook.xls
https://muster-vorlage.ch/fahrtenbuch-vorlage/
...

tananaev:
Closing as the PR is merged.

PR #663, opened 2018-04-14
Odometer in reports

ckrey:
Implements Odometer in Summary, Trips, and Stops Reports #652 
relies on corresponding PR in traccar/traccar


web/l10n/en.json
```diff
@@ -410,4 +412,4 @@
     "categoryTractor": "Tractor",
     "categoryTruck": "Truck",
     "categoryVan": "Van"
-}
```
tananaev: Please don't change end line.

web/app/view/ReportController.js
```diff
@@ -595,6 +611,10 @@ Ext.define('Traccar.view.ReportController', {
         dataIndex: 'startTime',
         xtype: 'datecolumn',
         renderer: Traccar.AttributeFormatter.getFormatter('startTime')
+    }, {
+        text: Strings.reportStartOdometer,
```
tananaev: Same question here. Shouldn't we just call it odometer?

ckrey:
ok, both done

web/l10n/en.json
```diff
@@ -385,6 +385,7 @@
     "reportSpentFuel": "Spent Fuel",
     "reportStartOdometer": "Odometer Start",
     "reportEndOdometer": "Odometer End",
+    "reportOdometer": "Odometer",
```
tananaev: We already have odometer string.

ckrey:
replace by PR #706

PR #665, opened 2018-04-19
Implement Maintenances

Abyss777:
Implemented web part for Maintenances

image

Types store is basically all number position attributes (like for charts)

Share some code with known attributes and fixed a bit `CustomNumberField`

### DIFF
--- a/web/app/store/MaintenanceTypes.js
+++ b/web/app/store/MaintenanceTypes.js
@@ -1,6 +1,6 @@
 /*
- * Copyright 2017 Anton Tananaev (anton@traccar.org)
- * Copyright 2017 Andrey Kunitsyn (andrey@traccar.org)
+ * Copyright 2018 Anton Tananaev (anton@traccar.org)
+ * Copyright 2018 Andrey Kunitsyn (andrey@traccar.org)
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -15,18 +15,9 @@
  * You should have received a copy of the GNU General Public License
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
-Ext.define('Traccar.store.CommonDeviceAttributes', {
+
+Ext.define('Traccar.store.MaintenanceTypes', {
     extend: 'Ext.data.Store',
     model: 'Traccar.model.KnownAttribute',
-
-    data: [{
-        key: 'speedLimit',
-        name: Strings.attributeSpeedLimit,
-        valueType: 'number',
-        dataType: 'speed'
-    }, {
-        key: 'report.ignoreOdometer',
-        name: Strings.attributeReportIgnoreOdometer,
-        valueType: 'boolean'
-    }]
+    proxy: 'memory'
 });
--- a/web/app/view/dialog/AttributeController.js
+++ b/web/app/view/dialog/AttributeController.js
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 - 2017 Anton Tananaev (anton@traccar.org)
+ * Copyright 2016 - 2018 Anton Tananaev (anton@traccar.org)
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -110,7 +110,8 @@ Ext.define('Traccar.view.dialog.AttributeController', {
                     }
                     break;
             }
-            if (valueField.getXType() !== config.xtype) {
+            if (valueField.getXType() !== config.xtype ||
+                    config.xtype === 'customNumberField' && valueField.dataType !== config.dataType) {
                 this.getView().down('form').insert(this.getView().down('form').items.indexOf(valueField), config);
                 this.getView().down('form').remove(valueField);
             } else if (config.xtype === 'customNumberField') {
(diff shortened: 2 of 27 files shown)


web/app/view/edit/Maintenances.js
```diff
@@ -0,0 +1,77 @@
...
+            dataIndex: 'type',
+            filter: {
+                type: 'list',
+                idField: 'key',
+                labelField: 'name',
+                store: 'MaintenanceTypes'
+            },
+            renderer: function (value) {
+                var attribute = Ext.getStore('MaintenanceTypes').getById(value);
+                return attribute && attribute.get('name') ? attribute.get('name') : value;
```
tananaev: I think this can be simplified to:
```js
return attribute && attribute.get('name') || value;
```

web/app/view/dialog/MaintenanceController.js
```diff
@@ -0,0 +1,70 @@
...
+        this.periodConfig = Ext.clone(this.lookupReference('periodField').initialConfig);
+        this.periodConfig.value = 0;
+        this.lookupReference('saveButton').setDisabled(true);
+    },
+
+    onValidityChange: function (form, valid) {
+        this.lookupReference('saveButton').setDisabled(!valid);
+    },
+
+    onNameChange: function (combobox, newValue) {
```
tananaev: This looks pretty complicated. Can you please explain what it does?

web/app/view/dialog/MaintenanceController.js
```diff
@@ -0,0 +1,70 @@
...
+            this.getView().down('fieldset').insert(this.getView().down('fieldset').items.indexOf(startField),
+                Ext.merge({}, this.startConfig, config));
+            this.getView().down('fieldset').remove(startField);
+            this.lookupReference('startField').validate();
+        } else {
+            startField.setConfig(config);
+            startField.validate();
+        }
+
+        periodField = this.lookupReference('periodField');
```
tananaev: This seems to be a duplicate of above code. I think it can be extracted into a separate method to avoid duplication.

web/app/view/dialog/Maintenance.js
```diff
@@ -0,0 +1,74 @@
...
+                reference: 'startField',
+                fieldLabel: Strings.maintenanceStart
+            }, {
+                xtype: 'customNumberField',
+                name: 'period',
+                reference: 'periodField',
+                allowBlank: false,
+                fieldLabel: Strings.maintenancePeriod,
+                validator: function (value) {
+                    return this.parseValue(value) !== 0 ? true : 'Can\'t be zero';
```
tananaev: This should probably be in strings.

web/app/view/dialog/AttributeController.js
```diff
@@ -110,7 +110,8 @@ Ext.define('Traccar.view.dialog.AttributeController', {
                     }
                     break;
             }
-            if (valueField.getXType() !== config.xtype) {
+            if (valueField.getXType() !== config.xtype ||
+                    config.xtype === 'customNumberField' && valueField.dataType !== config.dataType) {
```
tananaev: What is this for?

web/app/store/MaintenanceTypes.js
```diff
@@ -0,0 +1,23 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+Ext.define('Traccar.store.MaintenanceTypes', {
+    extend: 'Ext.data.Store',
+    model: 'Traccar.model.KnownAttribute',
+    proxy: 'memory'
```
tananaev: Is this required?

web/app/view/dialog/MaintenanceController.js
Abyss777: It prepares two configs for `CustomNumberField` of `start` and `period` fields and replaces components if `dataType` is changed.

web/app/view/dialog/Maintenance.js
Abyss777: Is it OK that other tooltips about why field is invalid will still not be translated? I wanted just make it consistent.

image

tananaev: Aren't they translated in ExtJS? Is there any default string that we can use?

Abyss777: Unfortunately I have not found any close default string. This is similar https://docs.sencha.com/extjs/6.2.0/classic/Ext.form.field.Base.html#cfg-invalidText

But I want to point that it can't be zero.

Abyss777:
I do not know how I missed it, but just noticed that I used the same glyph as for Server. Do you have an idea about substitution?

web/app/view/dialog/Maintenance.js
Abyss777: Sorry, I missed that this strings are translated by ExtJS.

tananaev:
Maybe we should change server glyph to something else and keep maintenance as it is?

Abyss777:
Yes, I wanted suggest the same.
Something like this https://fontawesome.com/v4.7.0/icon/server/ or this https://fontawesome.com/v4.7.0/icon/cogs/

tananaev:
I like the first one.

Abyss777:
Changed the icon
image

PR #668, opened 2018-04-24
[WIP] Add "About Traccar" menu option

matjon:
GPLv3 requires to display some license information:
(section 5)
    d) If the work has interactive user interfaces, each must display
    Appropriate Legal Notices; however, if the Program has interactive
    interfaces that do not display Appropriate Legal Notices, your
    work need not make them do so.

This patch adds an initial version of displaying such interactive legal notices.

This patch is unfinished: some lint warnings are still unfixed and strings should be moved to en.json. I would like to hear Your comments though.

This work was funded by Partner Security: http://www.partnersecurity.pl


matjon:
I will post a screenshot in a while.

tananaev:
You would need to clean up the code. It doesn't seem to be ready. A lot of commented out code, not translated strings, strings that shouldn't be translated in PR...

Also, all checks have to pass.

matjon:
OK, but what about the general idea?
As I mentioned above, the code is still unfinished - I will finish it before issuing the final PR.

> strings that shouldn't be translated in PR
You mean I should add new strings to en.json only and do not touch pl.json at all?

The screenshots are here:
http://78.31.92.254:82/share.cgi/Screenshot-2018-4-24%20Traccar.png?ssid=0KYY5pg&fid=0KYY5pg&open=normal&ep=
http://78.31.92.254:82/share.cgi/Screenshot-2018-4-24%20Traccar(1).png?ssid=0KYY5pg&fid=154cwwl&open=normal&ep=

tananaev:
I like general idea, but I think we should make something simpler. Maybe just an HTML file with links to appropriate projects and licenses?

matjon:
I am working on a very long HTML file with all projects Traccar depends on and their license terms (just like "Open source licenses" in many Android programs, like Firefox's about:license, etc.), but it is not ready yet. There is going to be a button that says: "Show all libraries Traccar depends on and their license terms" and after clicking it a HTML file will be displayed inside an iframe (by ShowIframe.js).

The long list of all libraries will not be included on the main "About Traccar" page. The three projects are there because I am able to query their version from JavaScript.

I am trying to make the notice as understandable as possible, unlike those of many Android software.

tananaev:
Why can't we just have a simple list of licenses (just project and license name) linking to original license files?

matjon:
Because links go stale and projects disappear.

I believe that GPL requires us to carry all such licenses together with the main program (as they contain conditions added under section 7 of the GPL) and not just links to them:

> b) The work must carry prominent notices stating that it is released under this License and any conditions added under section 7. This requirement modifies the requirement in section 4 to “keep intact all notices”.

Additionally, the license terms may have changed at the time the user displays them. I have been reviewing libraries which ExtJS uses and had problems with old releases of programs that are difficult to find. After all, many Android programs display just "Open source licenses" and not links to them.

Anyway, the hard part is getting a list of all projects Traccar depends on, and not their license terms.

I have pushed my work in progress to a branch [simpler_UI](https://github.com/matjon/traccar-web/tree/simpler_UI), the main part is in file [LICENSE_LIBRARIES.html](https://github.com/matjon/traccar-web/blob/simpler_UI/web/LICENSE_LIBRARIES.html).

matjon:
I do not know whether to put styles into web/app.css or leave them in place in About.js - existing code seems to use both approaches. Apart from that, the initial part of "About Traccar" menu option is done.

tananaev:
I still think it's overcomplicating something that should be extremely simple.  If you are worried that license links become inaccessible we can just copy them locally like you did with GPL, but I don't see any good reason to show it in a dialog. Just a link to open in a separate browser window would do. Much simpler implementation and easier to read for user.

matjon:
OK, but what about the current pull request? It would be nice to show the "About Traccar" window as a dialog box for interface consistency and visual quality.

I could rewrite `About.js` and turn it into a webpage and display it in a separate window (or maybe even in a popup by `ShowIframe.js`), but I doubt that would be much simpler as the content would be very similar.

It is not possible to force a browser to open a new tab from JavaScript:
https://stackoverflow.com/questions/4907843/open-a-url-in-a-new-tab-and-not-a-new-window-using-javascript

There might also be problems with popup blocking - the browser might prevent opening new windows from JavaScript (but it may not be an issue if we are opening a window from an `onClick()` handler).

The GPL license is another story. It could be displayed in a new window. The `ShowIframe` class is quite simple, however. I may just make the window bigger to make it easier for the user to read.

> but I don't see any good reason to show it in a dialog.

Just visual quality and interface consistency.

-------------------------------

## Concerning a future pull request:

I am going to open a future pull request to show a full list of libraries and their license terms in a new window, something like this:
Future view of about Traccar window
Future view of list of all libraries

> If you are worried that license links become inaccessible we can just copy them locally like you did with GPL

So it's just a choice of

- one big webpage containing all license terms

vs 

- a list of libraries with links pointing to their license terms which are in separate HTML pages.

tananaev:
Here is what I think we should do:
1. A small dialog box with simple HTML content with list of libraries links to licenses.
2. When you click on a license it will open a new browser window with full text of that license.

tananaev:
I don't think we need frames anywhere.

matjon:
OK, but what about the current pull request - the "About Traccar" window?

tananaev:
I don't think we need anything except licenses.

matjon:
But GPLv3 requires us to state the license of the program. The license says
>    d) If the work has interactive user interfaces, each must display
>    Appropriate Legal Notices; however, if the Program has interactive
>    interfaces that do not display Appropriate Legal Notices, your
>    work need not make them do so.

&nbsp;

>  An interactive user interface displays "Appropriate Legal Notices"
> to the extent that it includes a convenient and prominently visible
> feature that (1) displays an appropriate copyright notice, and (2)
> tells the user that there is no warranty for the work (except to the
> extent that warranties are provided), that licensees may convey the
> work under this License, and how to view a copy of this License.  If
> the interface presents a list of user commands or options, such as a
> menu, a prominent item in the list meets this criterion.

this is what the "About Traccar" menu currently does.

Also, the current approach enables us to translate the strings.

Besides, most programs have "About <program name>" menu item that displays a dialog box.

tananaev:
Please name any modern web-based apps that have about screen.

matjon:
I don't know, I don't use many web-based apps.

matjon:
There are few GPLv3 licensed web apps.

matjon:
And GPLv3 requires us to do an "About" / "License" box, either as a dialog box or a separate webpage.
A dialog is AFAIK better for UX.

tananaev:
That's what I'm suggesting. A simple dialog box with links to licences.

matjon:
OK, so the bulk of this pull request is OK? Just I need to remove `ShowIframe.js` and instead open GPLv3 in a new window?

tananaev:
No, I propose to simplify dialog to be just simple list and be in HTML, not a collection of extjs widgets that you have there now.

matjon:
You mean to open the dialog in a new window / tab? But this way it could not be localised.

tananaev:
Yes, but the idea is to just have a list of project names and licenses. Those don't require any translation. Dialog title can be translatable.

matjon:
OK, I may do this, but do not promise that I will do this soon.

When it comes to translation: after all, I may create a custom JavaScript in this HTML file and load language JSON files.

tananaev:
I don't see much need for translation there, but let's see.

PR #670, opened 2018-04-26
Add the "About Traccar" menu option - approach 2

matjon:
GPLv3 requires to display some license information: (section 5)

> d) If the work has interactive user interfaces, each must display
> Appropriate Legal Notices; however, if the Program has interactive
> interfaces that do not display Appropriate Legal Notices, your
> work need not make them do so.

This patch adds an initial version of displaying such interactive legal notices.

The web/LICENSE.html file does not currently have any CSS styling.

This work was funded by Partner Security: http://www.partnersecurity.pl

See the discussion at
        https://github.com/traccar/traccar-web/pull/668
Anton Tananaev wanted a simpler solution.

### DIFF
--- a/web/LICENSE.html
+++ b/web/LICENSE.html
@@ -70,7 +70,7 @@
                         The note below is important: Traccar contributors must be
                         aware that they are submitting work under GNU GPLv3 "or later".
                 -->
-                (note that Traccar source code is distributed under the terms of
+                (note that Traccar's source code is distributed under the terms of
                 the GNU GPL version 3 "or any later version", but use of some libraries
                 requires dropping the "or any later version" clause when distributing
                 the whole work).
--- a/web/LICENSE.html
+++ b/web/LICENSE.html
@@ -1,0 +1,106 @@
+<!DOCTYPE html>
+<!--
+        This program is free software: you can redistribute it and/or modify
+        it under the terms of the GNU General Public License as published by
+        the Free Software Foundation, either version 3 of the License, or
+        (at your option) any later version.
+
+        This program is distributed in the hope that it will be useful,
+        but WITHOUT ANY WARRANTY; without even the implied warranty of
+        MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+        GNU General Public License for more details.
+
+        You should have received a copy of the GNU General Public License
+        along with this program. If not, see <http://www.gnu.org/licenses/>.
+-->
+<html lang="en">
+<head>
+        <meta charset="utf-8">
+        <title>
+                About Traccar
+        </title>
+</head>
+<body>
+        <img src="/logo.svg" alt="Traccar">
+        <p>Modern GPS Tracking Platform</p>
+        <p>Professional, Secure, Open Source</p>
+
+        <!--
+                TODO: Display here information about client and server version.
+        -->
+
+        <!--
+                See: https://en.wikipedia.org/wiki/Copyright_notice
+
+                According to notices in source code, the server and web code when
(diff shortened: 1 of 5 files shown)


matjon:
There is a new "About Traccar" menu option.
When clicked, it opens in my Firefox a new tab which looks so:
https://raw.githubusercontent.com/matjon/traccar-web/about_license_initial2/web/LICENSE.html

I will fix the failing lint in a while.

matjon:
I have force updated the branch to make clear in a commit message that I still own the copyrights to the diff.
About Traccar window
http://78.31.92.254:82/share.cgi?ssid=02CWhCb

tananaev:
That's closer, but not exactly what I suggested.

matjon:
You mean to remove everything other than just a license list?

I would argue that more than just a license list is needed. GPLv3 specifies:

> An interactive user interface displays "Appropriate Legal Notices"
> to the extent that it includes a convenient and prominently visible
> feature that (1) displays an appropriate copyright notice, and (2)
> tells the user that there is no warranty for the work (except to the
> extent that warranties are provided), that licensees may convey the
> work under this License, and how to view a copy of this License.  If
> the interface presents a list of user commands or options, such as a
> menu, a prominent item in the list meets this criterion.

The license list is not finished yet, I may add it in the future.

matjon:
> But the most important I suggested to use dialog for this page. Only new window for full license text.

OK, so we probably didn't understand each other then. You mean to display a HTML inside an iframe in a dialog window? Beg my pardon, but I don't understand why this is better then just using ExtJS components inside a dialog window.

To be clear: I am going to display a short list of just several most important libraries on the interface the user accesses first. To see the full list of libraries, the user would have to click "see full list of libraries" or something similar.

It isn't there, but I am going to add a program version to the "About" window. Just so that the user will be able to know if they are using Traccar 3.14 or 3.16. So in my mind the "About Traccar" dialog was a bit more than just legal terms.

> Also, I think we should collapse copyright to something like
>
>     Copyright (C) Traccar contributors

I don't know the New Zealand law, but I suppose that using

>     Copyright (C) Anton Tananaev

would be more correct. I really mean it that You (and possibly Andrey Kunitsyn) probably have *full* legal rights to the *whole* of Traccar source code, to the point that You are able to relicense it under a proprietary license (provided that You didn't use works collected by others).
I suppose that Traccar is a "collective work" for the purposes of copyright law:

https://en.wikipedia.org/wiki/Collective_work

> A collective work is a work that contains the works of several authors assembled and published under the direction of one natural or legal person who owns the copyright in the work as a whole. Definitions vary considerably from one country to another, but usually treat ownership of the work as a whole as distinct from ownership of the individual contributions, so the individual authors may retain the right to publish their work elsewhere.


I'm sorry, but I'm beginning to be a bit fed up with adding the "About Traccar" menu option.

tananaev:
My suggestion is to use regular HTML in the dialog, not a frame. I kept saying it over and over.

matjon:
I'm sorry, but I am new to ExtJS and don't understand what You mean by "regular HTML in the dialog".
You could mean something like:

     }, {                                                                        
        xtype: 'tbtext',                                                        
        html: [ big chunk of HTML here ] 
     }, {                                                                        

but I am not sure.

tananaev:
Yes, something like that, but I don't think you need xtype.

matjon:
OK, but I don't understand why it would be better or simpler then just using ExtJS widgets. I would claim that ExtJS widgets would be more elegant in this case, putting big chunks of HTML in the html attribute is kind of hacky.

PR #671, opened 2018-04-28
minify.sh: use environment variable EXTJS_PATH

matjon:
Allow setting location of the `ext-6.2.0` directory by an environment variable. People can have `ext-6.2.0` in various locations.


tananaev:
Merged, thanks.

PR #675, opened 2018-05-14
Add speedLimit attribute to Geofence

Abyss777:
Added known attribute to Geofence in continue https://github.com/traccar/traccar/pull/3870


tananaev:
Merged, thanks.

PR #676, opened 2018-05-16
Add polylineDistance Geofence attribute

Abyss777:
Added known Geofence attribute introduced in https://github.com/traccar/traccar/pull/3838


tananaev:
Thanks.

PR #677, opened 2018-05-31
Add support of engine hours in milliseconds

Abyss777:
Added "one-item-store" to leave approach the same.

image

fix https://github.com/traccar/traccar/issues/3532


tananaev:
Merged, thanks.

PR #685, opened 2018-06-26
Add speed to events report

victorferreira:
Now the events report shows the speed limit and the registered speed whenever a vehicle passes the speed limit.

### DIFF
--- a/web/app/view/ReportController.js
+++ b/web/app/view/ReportController.js
@@ -514,6 +514,20 @@ Ext.define('Traccar.view.ReportController', {
         text: Strings.sharedMaintenance,
         dataIndex: 'maintenanceId',
         renderer: Traccar.AttributeFormatter.getFormatter('maintenanceId')
+    }, {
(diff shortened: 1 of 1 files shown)


tananaev:
@Abyss777, do you see any issues here?

Abyss777:
Yes, there are warnings if event does not contain such attributes. I would implement like this:
```javascript
        renderer: function (value) {
            if ('speedLimit' in value) {
                return Traccar.AttributeFormatter.speedFormatter(value['speedLimit']);
            }
            return null;
        }
``` 

If we will use them somewhere else (like events panel) it is better implement it like calculated fields, for now it is fine.

tananaev:
I also vote for calculated fields. Seems like a cleaner solution.

Abyss777:
I would more say universal solution, like we do for Route report, add "known" events attributes as hidden columns. But there would be some work...

tananaev:
Hmm.. yes, completely forgot about that. That would be much better and more consistent.

PR #686, opened 2018-06-28
Implement refactored notifications

Abyss777:
Web part for https://github.com/traccar/traccar/pull/3950

image

`tagfield` automatically set comma-separated values

image

Also improved `arrayListFilter` to handle comma-separated strings

image

Some comments in code...


web/app/Application.js
```diff
@@ -128,6 +129,11 @@ Ext.define('Traccar.Application', {
         return Strings[key] || key;
     },
 
+    getNotificatorString: function (eventType) {
+        var key = 'notification' + eventType.charAt(0).toUpperCase() + eventType.slice(1);
```
Abyss777: It is better use here `'notificator'` but we already have translated strings

web/app/AttributeFormatter.js
```diff
@@ -73,7 +73,16 @@ Ext.define('Traccar.AttributeFormatter', {
     },
 
     deviceIdFormatter: function (value) {
-        return Ext.getStore('Devices').getById(value).get('name');
+        var device, store;
+        if (value !== 0) {
```
Abyss777: Made it consistent with `xxxIdFormatter` mostly to handle if `deviceId = 0`

web/app/controller/Root.js
```diff
@@ -289,6 +290,8 @@ Ext.define('Traccar.controller.Root', {
                     this.beep();
                 }
                 Traccar.app.showToast(array[i].text, device.get('name'));
+            } else {
+                Traccar.app.showToast(array[i].text);
```
Abyss777: Handle a test event without `deviceId`

web/l10n/en.json
```diff
@@ -359,6 +359,7 @@
     "notificationWeb": "Send via Web",
     "notificationMail": "Send via Mail",
     "notificationSms": "Send via SMS",
+    "notificationNotificators": "Way of Notification",
```
tananaev: I think we should just name this "Channels".

web/l10n/en.json
```diff
@@ -359,6 +359,7 @@
     "notificationWeb": "Send via Web",
```
tananaev: What do you think about renaming these to just "web", "email" and "SMS"?

web/app/view/ArrayListFilter.js
```diff
@@ -34,6 +34,13 @@ Ext.define('Traccar.view.ArrayListFilter', {
                         return true;
                     }
                 }
+            } else if (property.indexOf(',') !== -1) {
```
tananaev: What if it's space?

web/app/store/AllNotificators.js
```diff
@@ -0,0 +1,32 @@
...
+
+Ext.define('Traccar.store.AllNotificators', {
+    extend: 'Ext.data.Store',
+    model: 'Traccar.model.KnownNotificator',
+
+    proxy: {
+        type: 'rest',
+        url: 'api/notifications/notificators',
+        listeners: {
+            'exception': function (proxy, response) {
```
tananaev: Do we need quotes around `exception`?

web/app/view/ArrayListFilter.js
Abyss777: Backend does not handle such case (space-separated).
But I can implement it in this custom component if we plan to use it somewhere else.

web/l10n/en.json
Abyss777: OK. Then we can use `"notificator"` as prefix...

web/app/view/ArrayListFilter.js
tananaev: I think it should be consistent. Your code below handles spaces.

web/l10n/en.json
tananaev: Sure.

Abyss777:
All done. Removed quotation from a couple other places.

tananaev:
Cool. I think it looks good. Can I merge both pull requests now?

Abyss777:
Forget screenshot
image

Yes I believe both are ready now.

tananaev:
Merged both. I'm thinking of implementing push notifications for Traccar Manager. It would be nice to have.

Abyss777:
I'm very waiting this. I even installed this project https://pushjet.io/ and wanted to implement own 'notificator' for this.

Moreover you already on the dark side and used firebase in client :)

You can count on my help.

tananaev:
I wouldn't call it a dark side. We still provide alternative without Google libraries.

Abyss777:
Yes, sorry. It is really important to have alternative.

PR #687, opened 2018-07-04
Add Tram, Train and Trolleybus categories

Abyss777:
Added a couple of categories for public transport
Thanks for SVGs.

image

fix https://github.com/traccar/traccar/issues/3930


tananaev:
Merged, thanks.

PR #709, opened 2018-10-25
Fix bug with selecting device second time from the devices grid

alex-petkevich:
Now, if you have only one entry in the devices grid and select a device from the list, the card is positioned only for the first time (the "selected" event is triggered). I added rowclick event handler. Now with any click on the device in devices grid, the map will show where the car is located.

### DIFF
--- a/web/app/view/edit/Devices.js
+++ b/web/app/view/edit/Devices.js
@@ -75,7 +75,8 @@ Ext.define('Traccar.view.edit.Devices', {
     },
 
     listeners: {
-        selectionchange: 'onSelectionChange'
+        selectionchange: 'onSelectionChange',
+        rowclick: 'onClick'
     },
 
     viewConfig: {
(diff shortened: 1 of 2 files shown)


tananaev:
If we handle each click, it probably makes sense to remove same logic from selection callback. Otherwise we send duplicate events every time.

alex-petkevich:
This case we will have another bug - if someone use only keyboard(tab/enter buttons) but not mouse/tochscreen, it will never fired and map will not positioning correct- this is an accessibility issue

tananaev:
There must be a way to implement it properly. Maybe we only trigger if it's already selected?

alex-petkevich:
This case it does not positioning properly if we have only one record in devices grid and move the map view to any place to hide car from displayed area - selection event fired only once, at the moment the row is selected. 
I'll try to add event processing when any key is pressed and remove selection event.

tananaev:
Great, thanks.

alex-petkevich:
I've updated pull request, could you take a look into it?

tananaev:
Merged, thanks.

PR #727, opened 2019-02-07
Orthographic correction

ddelagneau:
eventGeofence enter and exit


tananaev:

PR #734, opened 2019-04-15
Updating folder structure 

mdhanju:
1. Adding userActions.js for all user action.
2. Adding userReducers for all user reducers.
3. Moved pages to containers folder.

TODO: - 
1. Move async requests from components to actions so can be reused by other components


tananaev:
Is it a standard React folder structure? I think it make more sense to group by features.

mdhanju:
Actions/ - Redux folder structure.
Reducers/ - Redux folder structure.
Containers/ - stateful components
Components/ - stateless components (Can be reused in containers)

mdhanju:
@tananaev Any thoughts on PR

tananaev:
I have already added my thoughts. Waiting for your reply.

mdhanju:
Yes it is standard folder structure recommended by react redux.

Here is example from redux example project.
https://github.com/reduxjs/redux/tree/master/examples/real-world/src

tananaev:
It doesn't seem to be scalable. How do you handle a big app with multiple features and screens? Also, can you please point to the documentation where they say it's recommended.

mdhanju:
Containers are screens in which multiple components(features) can be included. 
Here is example from official redux git repo https://github.com/reduxjs/redux/tree/master/examples/real-world/src

tananaev:
You keep referring to the same sample. Is there any documentation about it?

PR #738, opened 2019-04-26
Translate es.json corrections

pdorgambide:
Fixed wrong spanish translation on geofence events.


tananaev:
All translations should be done through Transifex:

https://www.transifex.com/traccar/traccar

PR #742, opened 2019-05-20
Update dependencies to latest

tgroshon:
Several important packages were outdated enough to have transient deps
with security vulnerabilities. This patch upgrades key packages and
makes config adjustments to support them.

### DIFF
--- a/modern/package.json
+++ b/modern/package.json
@@ -5,14 +5,14 @@
   "dependencies": {
     "@material-ui/core": "^3.0.1",
     "@material-ui/icons": "^3.0.1",
-    "leaflet": "^1.3.4",
-    "react": "^16.4.2",
+    "leaflet": "^1.5.1",
+    "react": "^16.8.6",
     "react-container-dimensions": "^1.4.1",
-    "react-dom": "^16.4.2",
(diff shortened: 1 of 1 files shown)


tgroshon:
@tananaev this is my first step to contributing more features to the `modern` rewrite.

Related to this change, how are your feelings on including the `package-lock.json` in the git repo?  It's a fairly common practice, but I saw that file was specifically ignored in `.gitignore` so I wanted to get your thoughts before asking to change it.

tananaev:
Have you tested that it works? I remember I tried updating and there were some conflicts. Does websocket and API proxy work correctly?

As for lock file, I would prefer to leave it out, at least for now during active development.

tgroshon:
I did have to update the `proxy` key which changed on newer versions of `react-scripts` and add the `browserslist`.  I compared a build of my branch with current `modern` and they seem to behave identically:

1. Successful `npm start`
2. No additional console errors
3. Login works (fetch request to `/api/session` on login and main pages work) to my local traccar
4. WebSocket connects and pulls positions from my local traccar

So the upgrade seems good.

For the lock file, no problem. I will leave it out.

tananaev:
Merged, thanks.

PR #743, opened 2019-05-21
User registration and Hooks refactor

tgroshon:
Adds the user registration functionality.  Also performs the following refactors:

1. Run prettier over files
2. Add docstrings
3. Refactor many components to Functions and Hooks
4. Refactor authentication workflow to use Hooks + Context Providers

### DIFF
--- a/modern/package.json
+++ b/modern/package.json
@@ -5,6 +5,7 @@
   "dependencies": {
     "@material-ui/core": "^3.0.1",
     "@material-ui/icons": "^3.0.1",
+    "http-proxy-middleware": "^0.19.1",
     "leaflet": "^1.5.1",
     "react": "^16.8.6",
     "react-container-dimensions": "^1.4.1",
@@ -21,7 +22,6 @@
     "test": "react-scripts test --env=jsdom",
     "eject": "react-scripts eject"
   },
-  "proxy": "http://localhost:8082",
   "browserslist": {
     "production": [
       ">0.2%",
--- a/modern/package.json
+++ b/modern/package.json
@@ -12,7 +12,6 @@
     "react-dom": "^16.8.6",
     "react-leaflet": "^2.3.0",
     "react-redux": "^7.0.3",
-    "react-router-dom": "^5.0.0",
     "react-scripts": "^3.0.1",
     "redux": "^4.0.0"
   },
--- a/modern/src/App.css
+++ b/modern/src/App.css
@@ -1,0 +1,21 @@
+#spinner:before {
+  content: '';
+  box-sizing: border-box;
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  width: 60px;
+  height: 60px;
+  margin-top: -30px;
+  margin-left: -30px;
+  border-radius: 50%;
+  border-top: 6px solid #5fa2dd;
+  border-right: 6px solid transparent;
+  animation: spinner .6s linear infinite;
+}
+
+@keyframes spinner {
+  to {
+    transform: rotate(360deg);
+  }
+}
--- a/modern/src/App.js
+++ b/modern/src/App.js
@@ -1,21 +1,23 @@
-import React, { Component, Fragment } from 'react';
-import { Switch, Route } from 'react-router-dom'
-import CssBaseline from '@material-ui/core/CssBaseline';
-import MainPage from './MainPage';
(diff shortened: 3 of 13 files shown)


tananaev:
Can we please split changes, especially automated re-formatting.

Also, I don't want to add comments unless absolutely necessary.

tgroshon:
@tananaev Thanks for looking at this.  If you are willing, I would like to persuade you on those two points.  Sorry for the long write up, but I want to explain my thinking.

## 1. Commenting ##

I **completely agree** that comments should be avoided wherever possible.  Comments are lies!  However, I disagree in one _specific_ case: Function and Component docstrings.  Evil comments often explain "what" some code does and when they go out-of-date they become lies.  Good docstrings however explain "why" some function or component exists.

In my experience with React, it is _very useful_ to put explanatory docstrings on Components to explain why that component exists and the role it fills.  Because the docstrings are explaining the "why" of a high-level construct, they seldomly need updated or go out-of-date but provide very useful context for anyone approaching the code later.  This balance only works _**if**_ the docstrings stick to answering "why" and not "what" or "how".

Generally, I think I followed the above rules in this PR.  All but one of the "comments" I added are docstrings that I think fall into the "good" category.  I have one non-docstring comment that explains for a particular piece of code "why I chose to do it this way".

## 2. Splitting changes ##

I totally understand that this is not an easy PR to review because of the many kinds of changes.  However, I did split out most of the _logical_ changes into _separate commits_ which are individually pretty easy to review.  If you are willing, please review each commit rather than trying to take the PR as a whole.  I would like to avoid making _separate pull-requests_ for reformatting and refactoring because it will make everything take so much longer.  

### 2.1. Practicalities for this PR ###

If I were to break this PR into "logical pieces", it would probably be 3 separate PR's and branches with later ones based on the previous ones.  It wouldn't make sense to push all three at once because the later branches would still include all the previous branch changes so you would still have to review, update, and merge them in order.

This change took roughly 3 hours of work and included many very positive changes that I would include on any professional React project.  But if you are (understandably) only able to review changes every 12 hours, that means this 3 hours of work takes 15 hours; longer if the change needs updates/adjustments.

Additionally, some of the refactors in this PR _only make sense_ in context of the final change being made: i.e. adding user registration.  Otherwise, the refactors seem arbitrary _unless_ we previously agreed on making them.  If you want separate PRs for refactors, we will need to explicitly plan out and agree on the changes we make beforehand.  Otherwise, the refactors _must_ come bundled with the feature.

### 2.2. The Theory ###
 
This project is new, so it will certainly go through a lot of churn before it stabilizes.  The project is still figuring out it's style guide, code patterns, architecture, etc.  Because of that extra churn, the review process can easily become a bottleneck even if the same process works great in a similary, mature project (like current traccar or traccar-web).  Good review processes maintain stability, but young projects are inherently unstable.

For example, It takes a few minutes to reformat the whole codebase (with a tool like Prettier), but you probably won't be able to review it until that night or the next day.  If the work velocity is low, this is not a problem but if the velocity is high it becomes a throttle.  The long turnaround (which is totally understandable for this being a side project) makes it more rational for the writer to batch up many changes together.  

Generally, small PR's make for easy reviews but hard development.  For changes to the modern branch, I think you want to prioritize dev over reviews; features over stability.

## Conclusion ##

1. Re: comments -  I think the docstrings are a good idea, but if you still don't want them, I can remove.
2. Re: Split Changes - I _really_ don't want to spend my dev time splitting changes up and waiting a day for reviews on those small PRs.

You have a great project here, and I think you made the right decision on using React + Material UI.  I have a _lot_ of experience with React and am really excited to contribute that to your project.

**Recommendation**: I could get _a ton_ of work done over the next few weeks if you added me as a committer to your `modern` branch ;)

tananaev:
1. I can agree to useful comments on react components, but I see some on functions that are completely redundant and in some cases give less information than method name.

2. Large PRs always take more time to review and merge. Usually complexity grows exponentially. I can review small and easy PRs when I have a short break. For bigger ones I would have to find a long block of time, which won't be easy. Reviewing single commits is not possible because I cannot make comments specific to commit.

Unfortunately I don't think I can make you a committer just yet. I would recommend you to work on your own branch and merge changes in small logical parts. Creating a branch and PR takes a couple of clicks.

tananaev:
Any comments? Are you not interested in contributing anymore?

PR #748, opened 2019-07-08
animate marker on updateLatest

jcardus:
Created an animation when the device marker is updated. Please check the video.
[Video 2019-07-08 at 10.03.18.mp4.zip](https://github.com/traccar/traccar-web/files/3368068/Video.2019-07-08.at.10.03.18.mp4.zip)

### DIFF
--- a/web/app/Style.js
+++ b/web/app/Style.js
@@ -81,6 +81,9 @@ Ext.define('Traccar.Style', {
     mapGeofenceWidth: 5,
     mapGeofenceRadius: 9,
 
+    mapAnimateMarkerDuration: 2000,
+    mapAnimateMarkerTimoeout: 15,
+
     coordinatePrecision: 6,
     numberPrecision: 2,
 
--- a/web/app/Style.js
+++ b/web/app/Style.js
@@ -82,7 +82,7 @@ Ext.define('Traccar.Style', {
     mapGeofenceRadius: 9,
 
     mapAnimateMarkerDuration: 2000,
-    mapAnimateMarkerTimoeout: 15,
+    mapAnimateMarkerTimeout: 40,
 
     coordinatePrecision: 6,
     numberPrecision: 2,
--- a/web/app/view/map/MapMarkerController.js
+++ b/web/app/view/map/MapMarkerController.js
@@ -198,10 +198,10 @@ Ext.define('Traccar.view.map.MapMarkerController', {
         timeout = Traccar.Style.mapAnimateMarkerTimeout;
 
         updatePosition = function (position, marker) {
-            var coord, style;
-            coord = marker.get('line').getCoordinateAt(position / (duration / timeout));
+            var coordinate, style;
+            coordinate = marker.get('line').getCoordinateAt(position / (duration / timeout));
             style = marker.getStyle();
-            marker.setGeometry(new ol.geom.Point(coord));
+            marker.setGeometry(new ol.geom.Point(coordinate));
             if (position < duration / timeout) {
                 setTimeout(updatePosition, timeout, position + 1, marker);
             } else {
--- a/web/app/view/map/MapMarkerController.js
+++ b/web/app/view/map/MapMarkerController.js
@@ -192,7 +192,7 @@ Ext.define('Traccar.view.map.MapMarkerController', {
         var start = marker.getGeometry().getCoordinates(),
             end = geometry.getCoordinates(), line = new ol.geom.LineString([start, end]),
             duration = Traccar.Style.mapAnimateMarkerDuration,
-            timeout = Traccar.Style.mapAnimateMarkerTimoeout,
+            timeout = Traccar.Style.mapAnimateMarkerTimeout,
 
             updatePos = function (pos, m) {
                 var coord = m.get('line').getCoordinateAt(pos / (duration / timeout)),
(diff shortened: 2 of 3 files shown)


tananaev:
What would happen if you get updates every second?

jcardus:
I can put the duration of the movement configurable. In that case I would set duration to 500 ms.

tananaev:
I think the right solution is to cancel old animation.

jcardus:
you mean "cancel all animation"?

yes, which would be done setting the value to 0.

tananaev:
If new update is triggered earlier than animation finished, we need to cancel old animation. That's what I meant.

jcardus:
Oh ok, that makes sense. Gonna work on that.

Should I still put the duration of the movement configurable? Any suggestion for the name of the setting?

tananaev:
We usually put constants like this into "Style.js". Not ideal, but better than hardcoded.

jcardus:
Can you take a look again?

web/app/view/map/MapMarkerController.js
```diff
@@ -144,7 +144,7 @@ Ext.define('Traccar.view.map.MapMarkerController', {
                 marker = this.latestMarkers[deviceId];
                 style = marker.getStyle();
                 if (style.getImage().fill !== this.getDeviceColor(device) ||
-                        style.getImage().category !== device.get('category')) {
```
tananaev: Why are you changing styling everywhere? Please revert.

web/app/Style.js
```diff
@@ -81,6 +81,9 @@ Ext.define('Traccar.Style', {
     mapGeofenceWidth: 5,
     mapGeofenceRadius: 9,
 
+    mapAnimateMarkerDuration: 2000,
+    mapAnimateMarkerTimoeout: 15,
```
tananaev: Typo

web/load.js
```diff
@@ -174,7 +174,7 @@
         addScriptFile('https://cdn.jsdelivr.net/npm/proj4@' + proj4jsVersion + '/dist/proj4.js');
     }
 
-    window.Images = ['arrow', 'default', 'animal', 'bicycle', 'boat', 'bus', 'car', 'crane', 'helicopter', 'motorcycle',
+    window.Images = ['arrow', 'default', 'animal', 'bicycle', 'boat', 'bus', 'car', 'car2', 'crane', 'helicopter', 'motorcycle',
```
tananaev: What is this change for?

web/app/Style.js
tananaev: Also, updating every 15 ms seem too often. It can cause issues when you have many markers. Maybe something between 30-50 is more reasonable.

web/app/view/map/MapMarkerController.js
```diff
@@ -188,6 +188,35 @@ Ext.define('Traccar.view.map.MapMarkerController', {
         }
     },
 
+    animateMarker: function (marker, geometry, course) {
+        var start = marker.getGeometry().getCoordinates(),
```
tananaev: Please split definitions and assignment, like we do in other places.

web/app/view/map/MapMarkerController.js
```diff
@@ -188,6 +188,35 @@ Ext.define('Traccar.view.map.MapMarkerController', {
...
+            end = geometry.getCoordinates(), line = new ol.geom.LineString([start, end]),
+            duration = Traccar.Style.mapAnimateMarkerDuration,
+            timeout = Traccar.Style.mapAnimateMarkerTimoeout,
+
+            updatePos = function (pos, m) {
+                var coord = m.get('line').getCoordinateAt(pos / (duration / timeout)),
```
tananaev: Same here.

jcardus:
Yes, instead of canceling the old animation and starting a new one I'm updating the path. 
When a new event comes in and the old animation hasn't finished, the marker will jump to the latest event. The animation will finish in the newest position.

web/app/view/map/MapMarkerController.js
```diff
@@ -188,6 +188,39 @@ Ext.define('Traccar.view.map.MapMarkerController', {
...
+            style = m.getStyle();
+            m.setGeometry(new ol.geom.Point(coord));
+            if (pos < duration / timeout) {
+                setTimeout(updatePos, timeout, pos + 1, m);
+            } else {
+                if (style.getImage().angle !== m.get('nextCourse')) {
+                    style.setImage(Traccar.DeviceImages.getImageIcon(
+                        style.getImage().fill, style.getImage().zoom, m.get('nextCourse'), style.getImage().category));
+                }
+                m.set('isAnimating', false);
```
tananaev: Can we rename this to just `animating`.

web/app/view/map/MapMarkerController.js
```diff
@@ -188,6 +188,39 @@ Ext.define('Traccar.view.map.MapMarkerController', {
...
+    animateMarker: function (marker, geometry, course) {
+        var start, end, duration, timeout, line, updatePos;
+
+        start = marker.getGeometry().getCoordinates();
+        end = geometry.getCoordinates();
+        line = new ol.geom.LineString([start, end]);
+        duration = Traccar.Style.mapAnimateMarkerDuration;
+        timeout = Traccar.Style.mapAnimateMarkerTimeout;
+
+        updatePos = function (pos, m) {
```
tananaev: Please use full words for names. Pos should be Positions. M should be Marker. etc

jcardus:
Thanks Anton, I think we're ok now. I also reused the rotateMarker function with the 'self' trick, the function had 0 usages with the previous change.

tananaev:
Merged, thanks for the pull request.

tananaev:
@jcardus looks like the change broke map centering:

https://www.traccar.org/forums/topic/follow-behavior-and-webshockets/#post-50650

Would you be able to fix that?

jcardus:
Hi Anton,

Sorry, just saw this now.

I will fix it.

PR #749, opened 2019-07-13
Add support for custom ArcGIS REST map services

edvalley:
Adding support for custom ArcGIS REST map services. Of course, still left to add the whole localization strings. Suggest the names (maybe shorter) you like for variables and display texts. Thanks a lot.


web/app/view/map/BaseMap.js
```diff
@@ -38,14 +38,21 @@ Ext.define('Traccar.view.map.BaseMap', {
         bingKey = server.get('bingKey');
 
         switch (type) {
-            case 'custom':
+            case 'customXyzFormat':
```
tananaev: Why are you changing existing layout key? It will break everyone's configuration.

web/l10n/en.json
```diff
@@ -232,6 +232,8 @@
     "mapTitle": "Map",
     "mapLayer": "Map Layer",
     "mapCustom": "Custom Map",
+    "mapCustomXyzFormat": "Custom (XYZ)",
```
tananaev: I would keep original string instead. We can potentially change text, but I don't see any reason for adding a new string.

web/l10n/en.json
```diff
@@ -232,6 +232,8 @@
...
+    "mapCustomArcgisRest": "Custom (ArcGIS REST)",
```
tananaev: Can we drop "rest" to make this shorter?

web/app/view/map/BaseMap.js
edvalley: Oh, you're right, I didn't think about that, it will break everyone's configuration. I'll do it the way you suggest bellow. Thanks...

web/l10n/en.json
edvalley: I agree. Doing it the way you suggest. Thanks.

web/l10n/en.json
edvalley: Of course. Good. Thanks.

edvalley:
Fixed things up. Remember, if you approve it, I'll fix all localization stuff. Thanks.

edvalley:
Sorry, would you like to keep "Custom map" instead of "Custom"? I'm afraid that it loss sense in some languages. Thanks.

tananaev:
I think it looks good. Thanks for pull request.

edvalley:
Thank you! Should I add more commits to fix localization or do I need to issue another pull request?

tananaev:
Localization is done on Transifex website. Please don't send any localization PRs.

edvalley:
Oh, I didn't know about that. That's great! Thank you very much.

PR #751, opened 2019-07-19
Fix misleading displayed label

edvalley:
I just noted that my PR [https://github.com/traccar/traccar-web/pull/749](https://github.com/traccar/traccar-web/pull/749), has introduced a misleading displayed label. No matter what map layer is selected, the label for the custom map URL field, always displays as `Custom (XYZ)`. Please, suggest the name you like for the key.
Now, I remember the reason why I created a new key `mapCustomXyz` for `Custom (XYZ)`, instead of reusing `mapCustom`. Of course, as you said, that change would have broken everyone's configuration.
Thanks.

### DIFF
--- a/web/app/view/dialog/Server.js
+++ b/web/app/view/dialog/Server.js
@@ -47,7 +47,7 @@ Ext.define('Traccar.view.dialog.Server', {
                 xtype: 'unescapedTextField',
                 reference: 'mapUrlField',
                 name: 'mapUrl',
-                fieldLabel: Strings.mapCustom
+                fieldLabel: Strings.mapCustomLabel
             }, {
                 xtype: 'numberfield',
                 reference: 'latitude',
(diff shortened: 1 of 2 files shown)


tananaev:
Can you explain the problem a bit more.

edvalley:
Of course, let me take some screenshots to post here.

edvalley:
This way it should be (The custom map url label displays as `Custom map`)
ok
This way it is now (The custom map url label displays as `Custom (XYZ)`, no matter what map layer you have selected)
bad
That happens because the display text for the key `mapCustom` has been changed to `Custom (XYZ)`, and it is being used for both, the map layer drop down item text, and the label for the custom map url.

tananaev:
Merged, thanks.

edvalley:
Thanks to you.

PR #758, opened 2019-09-17
Fix some translation problems in pt-br and pt

lcsvcn:
I have make a PR some months ago adding the Scooter category, as it rises popularity, but someone has made a PR changing the translation to something the icon doesn't represent.

It is a Eletric Scooter, as Grin, Yellow, Lime, Bird uses.
Scooter as "Patinete Elétrico":
image

Not Scooter as a type of motorcycle with less motor power, as the meaning of Lambreta means.
Scooter as "Lambreta": 
image


tananaev:
Please use Transifex:

PR #767, opened 2020-03-04
add Pushover attrs

floetenbaer:
see pull request in traccar/traccar


tananaev:
If it's still relevant, please create a new PR that add it to both web apps.

PR #769, opened 2020-03-24
Feature(modern): Use local Environment

dt-ap:
Using _.env.local_, **User** could freely choose the **backend** URL name (other than localhost)


tananaev:
Merged, thanks.

PR #771, opened 2020-03-24
Update README.md

spacecabbie:
Added steps to clone partial repo (aka modern dir only)

### DIFF
--- a/modern/README.md
+++ b/modern/README.md
@@ -5,6 +5,12 @@ It uses [React](https://reactjs.org/), [Material UI](https://material-ui.com/) a
 To run the project in development mode:
 
 - Make sure you have Traccar backend running.
+- empty out /web
+- git init
+- git config core.sparseCheckout true
+- git remote add -f origin https://github.com/traccar/traccar-web
+- git pull origin master
+- cd modern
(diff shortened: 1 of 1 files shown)


tananaev:
This is not really a necessary part to run the web app, so I don't want to overcomplicate readme doc with it.

spacecabbie:
> This is not really a necessary part to run the web app, so I don't want to overcomplicate readme doc with it.

I do understand your point. Just wanted to let as much people able to test it including the not so technical abble. since this is the development dir it could be replaced with the proper readme when pushed as release.

tananaev:
Yeah, but even with development I don't really see any good reason for it. What problem does it solve?

spacecabbie:
None whatsoever,

Your looking at it from programming perspective and then yes it has no value. But to help a user it allows quick testing which in turn adds to feedback and bug reporting.

Like myself for example I consider myself in between programmer/user. It took me a while to figure this out, I don't work daily with git etc.
Now i have it running i can add basic options to the template. and contribute even though it's a little. 

I get your point though leaving it up to you ofc.

tananaev:
Why do you want to partially clone repo instead of full repo? I still don't understand it.

Issue #770, opened 2020-03-24
Proposal(modern): Using Modern React Tools.


dt-ap:
@tananaev I propose two different modern react tools for this project.

### [Redux Toolkit](https://redux-toolkit.js.org/)
With this library there will be less boilerplate when creating redux actions/reducer/actions-creator. Furthermore, the library is already packaged with some middlewares:
- [redux-thunk](https://github.com/reduxjs/redux-thunk) for redux asynchronous operations.
- [immer](https://immerjs.github.io/immer/docs/introduction) for easier immutable operations.

### [Hooks](https://reactjs.org/docs/hooks-intro.html)
This is a personal preference. Using React component class is okay too, I guess. I like hooks more than using [Render Props](https://reactjs.org/docs/render-props.html) or [Higher Order Component](https://reactjs.org/docs/higher-order-components.html).

If you interested, I will write two different PR for each of these tools and you could make the decision to use it or not.

tananaev:
Redux Tookit seems like a good idea. Removed boilerplate. Feel free to send a pull request.

Hooks not convinced yet. If you can give an example where it solves a problem. I guess it might make sense for some really small components where we don't want to write a class.

RodolfoSilva:
@tananaev for example: Today we have a redux with positions, devices, selectedDevices.... A lot of boilerplate to use that. When you can use a structure like redux but more readable and isolated.

An example: It is if we want rewrite the devices state to hooks we can use something like this

```
import React from 'react';

const DevicesContext = React.createContext();

const reducer = (state, action) => {
  switch(action.type) {
    case 'UPDATE_DEVICES':
      return {
        ...state,
        ...action.payload,
      };
    default:
      throw new Error(`Unknown action ${action.type}`);
  }
};

export function DevicesProvider(props: DeviceProviderProps) {
  const { children } = props;

  const [state, dispatch] = React.useReducer(reducer, {});

  const updateDevices = React.useCallback((devices) => dispatch({ action: 'UPDATE_DEVICES', payload: devices }), [dispatch]);
  
  return <DevicesContext.Provider value={{ devices: state, updateDevices }}>{children}</DevicesContext.Provider>
}

export function useDevices() {
  const context = React.useContext(DevicesContext);

  if (context === undefined) {
    throw new Error(`useDevices must be used within a DevicesContext`)
  }

  return context
}
```

When we need to use anything related with devices information we just need import and use the `useDevices`;

```
import React from 'react';
[...]

PR #772, opened 2020-03-24
Experimental: Use Redux Toolkit

dt-ap:
See: issue #770


modern/jsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "baseUrl": "src"
+    "baseUrl": "./"
```
tananaev: Why do we need changes in this file?

dt-ap: On the [SocketController lines 3](https://github.com/traccar/traccar-web/blob/4d200181a6a97c8143d0843f07d90072ef35e25c/modern/src/SocketController.js#L3) or [DeviceList lines 14](https://github.com/traccar/traccar-web/blob/4d200181a6a97c8143d0843f07d90072ef35e25c/modern/src/DeviceList.js#L14)
 
You can see that I am importing from **store** using `import ... from 'src/store';`. If I use `"baseUrl": "src"`, I could import it like: `import ... from 'store';`

But I'm afraid it would create a conflict with some library or nodejs modules, So I think it better to use `import ... from 'src/store';`

tananaev: Why do we need it in the first place?

dt-ap: If, later, the project become more complex with so many subfolders. We would not have to import like this: `import ... from '../../../store'; or `import ... from '../../../SomeComponent'`

tananaev: That makes sense, but:
1. Let's split that into a separate pull request
2. If we do it, let's update all imports

dt-ap: OK then, I will remove it.

tananaev:
Merged, thanks.

PR #773, opened 2020-03-24
Add typescript

RodolfoSilva:
Hey @tananaev I'm sending this PR to add typescript support, if you accept this, I'll send a new PR with all others files converted.


tananaev:
I'm not sure about typescript. Also, you haven't provided any arguments.

RodolfoSilva:
Right, 
* Typescript will provide a better developer experience.
* Prevents some bugs, because now we can have types to help avoid typos and other things.
* Documentation, with typescript is very easy to understand what we need for execute a function or whats is the data structure of properties or entity.
* Type checking in compilation time.

tananaev:
For now I'm not convinced that it's a good idea. It adds another layer.

LWlook:
This greatly reduces the likelihood of errors and I would love to see Typescript in such a project. I would never again program in a not typed language.
And it don't really adds another layer (depends on what you see as layer). It only helps the programmer to program, in the output typescript no longer occurs as it is compiled.

PR #774, opened 2020-03-27
Experimental: Use Hooks

dt-ap:
Convert four components from using **Class Component** to **Function Component** using **Hooks**.

### DIFF
--- a/modern/src/DeviceList.js
+++ b/modern/src/DeviceList.js
@@ -1,30 +1,26 @@
-import React, { Component, Fragment } from 'react';
-import { connect } from 'react-redux';
+import React, { Fragment } from 'react';
+import Avatar from '@material-ui/core/Avatar';
+import Divider from '@material-ui/core/Divider';
+import IconButton from '@material-ui/core/IconButton';
 import List from '@material-ui/core/List';
 import ListItem from '@material-ui/core/ListItem';
 import ListItemAvatar from '@material-ui/core/ListItemAvatar';
-import ListItemText from '@material-ui/core/ListItemText';
-import Avatar from '@material-ui/core/Avatar';
-import LocationOnIcon from '@material-ui/icons/LocationOn';
 import ListItemSecondaryAction from '@material-ui/core/ListItemSecondaryAction';
-import IconButton from '@material-ui/core/IconButton';
+import ListItemText from '@material-ui/core/ListItemText';
+import LocationOnIcon from '@material-ui/icons/LocationOn';
 import MoreVertIcon from '@material-ui/icons/MoreVert';
-import Divider from '@material-ui/core/Divider';
+
+import { useDispatch, useSelector } from 'react-redux';
 import { devicesActions } from './store';
 
-const mapStateToProps = state => ({
-  devices: Object.values(state.devices.items)
-});
+const DeviceList = () => {
+  const devices = useSelector(state => Object.values(state.devices.items));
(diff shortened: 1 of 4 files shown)


modern/src/DeviceList.js
```diff
@@ -1,96 +1,79 @@
...
-          <MenuItem onClick={() => this.handleMenuRemove()}>{t('sharedRemove')}</MenuItem>
-        </Menu>
-        <RemoveDialog
-          open={this.state.removeDialogOpen}
-          onClose={() => { this.setState({ removeDialogOpen: false }) }} />
-      </Fragment>
-    );
-  }
+  return (
+    <>
```
tananaev: Is this fragment?

modern/src/MainToolbar.js
```diff
@@ -32,149 +33,124 @@ const styles = theme => ({
...
-                </ListItemIcon>
-                <ListItemText primary={t('sharedNotifications')} />
-              </ListItem>
-            </List>
-          </div>
-        </Drawer>
-      </Fragment>
-    );
-  }
+  return (<>
```
tananaev: Please move tag to the next line

modern/src/MainToolbar.js
```diff
@@ -32,149 +33,124 @@ const styles = theme => ({
...
+          <ListItem button disabled>
+            <ListItemIcon>
+              <SettingsIcon />
+            </ListItemIcon>
+            <ListItemText primary={t('sharedNotifications')} />
+          </ListItem>
+        </List>
+      </div>
+    </Drawer>
+  </>);
```
tananaev: Same with closing

modern/src/DeviceList.js
dt-ap: Yes

modern/src/MainPage.js
```diff
@@ -35,57 +36,45 @@ const styles = theme => ({
...
-              </ContainerDimensions>
-            </div>
-          </div>
+  return loading ? (<div>Loading...</div>) : (
+    <div className={classes.root}>
+      <SocketController />
+      <MainToobar />
+      <div className={classes.content}>
+        {/* <Drawer
+          anchor={isWidthUp('sm', this.props.width) ? "left" : "bottom"} NOTE: What's this do?
```
tananaev: There is a desktop and mobile version. Please keep the original code. Are there any other changes you've made?

dt-ap: I think there are not. But without using `isWidthUp` the permanent drawer moves to bottom automatically

tananaev: Let's revert this change. If you want to change something, I'm happy to review and discuss it, but let's not mix migration and changes together. It makes it impossible to review.

modern/src/MainPage.js
```diff
@@ -35,57 +37,42 @@ const styles = theme => ({
...
-              </ContainerDimensions>
-            </div>
-          </div>
+  return loading ? (<div>Loading...</div>) : (
+    <div className={classes.root}>
+      <SocketController />
+      <MainToobar />
+      <div className={classes.content}>
+        <Drawer
+          anchor={isWidthUp('sm', width) ? "left" : "bottom"}
```
dt-ap: @tananaev Material-UI use different API if you want the hooks alternative. Here the [docs](https://material-ui.com/components/use-media-query/)

tananaev:
Merged, thanks.

PR #775, opened 2020-04-19
Add long./lat./zoom URI parameters to simple UI

kristian:
Instead of reading the longitude, latitude and zoom options from the
user / server configuration, one could use the new longitude=,
latitude= and zoom= URI parameters, in order to define the launch map
view. This is especially useful, in case you would like to embed
multiple maps with different views into e.g. another website.

### DIFF
--- a/web/simple/app.js
+++ b/web/simple/app.js
@@ -29,8 +29,12 @@ if (!Array.prototype.find) {
   });
 }
 
+var getQueryParameter = function(name) {
+	return (window.location.search.match('[?&]' + name + '=([^&]*)') || [])[1];
+};
+
 var url = window.location.protocol + '//' + window.location.host;
-var token = (window.location.search.match(/token=([^&#]+)/) || [])[1];
+var token = getQueryParameter('token');
 
 var style = function (label) {
     return new ol.style.Style({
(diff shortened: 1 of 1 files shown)


web/simple/app.js
```diff
@@ -98,10 +98,10 @@ ajax('GET', url + '/api/server', function(server) {
     ajax('GET', url + '/api/session?token=' + token, function(user) {
 
         map.getView().setCenter(ol.proj.fromLonLat([
-            user.longitude || server.longitude || 0.0,
-            user.latitude || server.latitude || 0.0
+            parseFloat((window.location.search.match(/[?&]longitude=([^&#]+)/) || [])[1]) || user.longitude || server.longitude || 0.0,
```
tananaev, line 101: Can you extract `(window.location.search.match(/[?&]XXX=([^&#]+)/) || [])[1]` into some function? Easier to read and avoid duplication.

kristian: Makes sense. Done as requested. I also simplified the RegEx a little, as window.location.search may never contain any unescaped # (only window.location.href could), so expanding the group with [^&]+ is enough.

tananaev:
Merged, thanks.

PR #779, opened 2020-07-28
 Implement simple trips report

karls4:


tananaev:
I feel like we need to remove code duplication between reports. Maybe split it into smaller components?

karls4:
I will try remove the duplication.

tananaev:
Thanks. Happy to discuss the approach first if you have some ideas.

karls4:
image

What about having a report type picker?

tananaev:
You mean the last part? Yes, that makes sense. I also feel like there is an opportunity to generalize the table as well. Note that field for route report will be configurable in future, so maybe we can just have generic report table that takes a list of fields as an input?

karls4:
I have added a url parameter to the report page to take the report type eg "route", "trips"
Added a report selector to the top of the report parameters sections
depending on what type of report is shown decides which table is shown

tananaev:
That's probably too much. I think we should still keep a page per report, but split sub-components into more generic ones.

tananaev:
Any updates?

karls4:
Hi, no updates yet. I might get a chance this week to get a look at it.

karls4:
image

What if I split this section into its own component and reuse it on the two report pages?

Do we need anything else added to it? 

Groups, show markers etc

tananaev:
That makes sense. We can add new stuff later in a separate PR. I would recommend keep PRs as small as possible and don't combine multiple changes.

tananaev:
We already have trips report.

PR #781, opened 2020-08-11
Add More Attributes to the device page

karls4:


modern/src/DevicePage.js
```diff
@@ -88,6 +96,33 @@ const DevicePage = () => {
               label={t('deviceIdentifier')}
               variant='filled' />
           }
+          {(!id || device) &&
```
tananaev, line 99: We should probably move this check up, so it's not repeated so many times.

modern/src/DevicePage.js
```diff
@@ -38,10 +40,16 @@ const DevicePage = () => {
     });
   }, [id]);
 
+
+
```
tananaev, line 44: Please remove extra lines.

modern/src/DevicePage.js
```diff
@@ -71,22 +76,47 @@ const DevicePage = () => {
...
-          }
-          {(!id || device) &&
-            <TextField
-              margin='normal'
-              fullWidth
-              defaultValue={device && device.uniqueId}
-              onChange={(event) => setUniqueId(event.target.value)}
-              label={t('deviceIdentifier')}
-              variant='filled' />
+            <div>
```
tananaev, line 79: Is there a way to avoid `div`? Can we use fragment here?

karls4: Sorry I was unaware of fragment.

modern/src/DevicePage.js
```diff
@@ -71,22 +76,47 @@ const DevicePage = () => {
...
-          }
-          {(!id || device) &&
-            <TextField
-              margin='normal'
-              fullWidth
-              defaultValue={device && device.uniqueId}
-              onChange={(event) => setUniqueId(event.target.value)}
-              label={t('deviceIdentifier')}
-              variant='filled' />
+            <React.Fragment>
```
tananaev, line 79: Please use empty tag instead for consistency with other code.

karls4:
image

See attached. Its plain and simple at the moment

tananaev:
Also already implemented.

PR #788, opened 2020-10-08
WIP: drivers page

dt-ap:
Just started working on drivers. I create new page with route `/drivers`.
Need your input for the UI and the flow

driver_screenshot


tananaev:
Unfortunately we can't merge this. It needs to be more complete.

PR #790, opened 2020-10-21
Add 2FA using TOTP

mivtdole:
This is part of the adding 2FA using TOTP to traccar


tananaev:
Closing as outdated. It needs to be added to the modern app now.

PR #794, opened 2020-11-15
Implementing export to excel reports

mail2bishnoi:
Hi Anton,

I have started working on export to Excel. But there is a challenge in that to share the data (deviceId, from, to, etc) between sibling components (ReportFilterForm and table component), I think to do it by using redux and refactoring the existing code. Can you please give me some advice about this?

export-to-excel


tananaev:
I've already implemented export.

PR #795, opened 2020-11-17
Added report title to all the reports

mail2bishnoi:
Hi Anton,

I have added the title to all the report tables. Please review the code once and suggest improvements.

With this, I would like to work on chart reports. Would you like to recommend any chart library?

Thanks.

report-title-screen

### DIFF
--- a/modern/src/reports/EventReportPage.js
+++ b/modern/src/reports/EventReportPage.js
@@ -1,10 +1,11 @@
 import React, { useState } from 'react';
-import { TableContainer, Table, TableRow, TableCell, TableHead, TableBody, Paper } from '@material-ui/core';
+import { Table, TableRow, TableCell, TableHead, TableBody, Paper } from '@material-ui/core';
 import { FormControl, InputLabel, Select, MenuItem } from '@material-ui/core';
 import t from '../common/localization';
 import { formatPosition } from '../common/formatter';
 import ReportFilter from './ReportFilter';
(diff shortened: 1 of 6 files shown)


tananaev:
What is the value in this? And if we do add title, it should probably go in the toolbar.

tananaev:
Haven't investigated charts, but let's discuss before you start implementation.

mail2bishnoi:
> What is the value in this? And if we do add title, it should probably go in the toolbar.

Because sometime I confuse that which report I am looking at, and I have to look at url. Thats why I think its better to add a report heading.

By toolbar you mean page's title?

tananaev:
I mean toolbar, not page title.

mail2bishnoi:
> 
> 
> I mean toolbar, not page title.

Sorry, but I didn't get which toolbar you are talking about.

tananaev:
Each page has a toolbar (`MainToolbar` class). I'm not sure how you can miss it. It's on the top of each page.

mail2bishnoi:
Now I understood. 
Yes, that can also be done and its advantage will be that we can set the title for each page. But for now, I only thought about the headings for the reports table.

I will implement the what you suggest.

mail2bishnoi:
Should I proceed this or close the PR?

tananaev:
Up to you. It would be nice to add title to the toolbar, but I don't think it's super urgent. If you go ahead, make sure it's a universal solution.

PR #797, opened 2020-11-20
Implementing chart report initial structure

mail2bishnoi:
Hi Anton,

I started working on chart reports. Although it is in a very early stage, a few points I would like to discuss with you -

1. Since the chart type attribute does not go into the input of the API, I did not include it in the report filter form but added it to the chart component itself.
2. Which is the best way to convert a nested position object to a simple object? Because we need to draw a graph with attributes like distance, totalDistance, etc

chart-report

chart-report-2

Best Regards


modern/src/common/formatter.js
```diff
@@ -9,7 +9,7 @@ export const formatPosition = (value, key) => {
     case 'fixTime':
     case 'deviceTime':
     case 'serverTime':
-      return moment(value).format('LLL');
+      return moment(value).format('YYYY-MM-DD HH:mm');
```
tananaev, line 12: This is not a universal solution. We need a better one.

modern/src/reports/ChartReportPage.js
```diff
@@ -0,0 +1,74 @@
...
+import { Card, CardHeader, CardContent } from '@material-ui/core';
+import { Box, Divider } from '@material-ui/core';
+import { makeStyles } from '@material-ui/core/styles';
+import ReportFilter from './ReportFilter';
+import ReportLayoutPage from './ReportLayoutPage';
+import ChartType from './ChartType';
+
+import {LineChart, Line, XAxis, YAxis, CartesianGrid, Tooltip, Legend, ResponsiveContainer } from 'recharts';
+
+const useStyles = makeStyles((theme) => ({
```
tananaev, line 11: Why do we need this?

modern/src/reports/ChartReportPage.js
```diff
@@ -0,0 +1,74 @@
...
+  const handleSubmit = async (deviceId, from, to, mail, headers) => {
+    const query = new URLSearchParams({ deviceId, from, to, mail });
+    const response = await fetch(`/api/reports/route?${query.toString()}`, { headers });
+    if (response.ok) {
+      const contentType = response.headers.get('content-type');
+      if (contentType) {
+        if (contentType === 'application/json') {
+          setItems(await response.json());
+        } else {
+          window.location.assign(window.URL.createObjectURL(await response.blob()));
```
tananaev, line 28: Why is this here? Please make sure you clean up all unnecessary stuff.

modern/src/reports/ChartReportPage.js
```diff
@@ -0,0 +1,74 @@
...
+  const [items, setItems] = useState([]);
+  const [type, setType] = useState('speed');
+
+  return (
+    <ReportLayoutPage reportFilterForm={ReportFilterForm} setItems={setItems} >
+      <Card>
+        <CardHeader action={<ChartType type={type} setType={setType}/>} />
+        <Divider />
+        <CardContent>
+          <Box height={400} position="relative">
```
tananaev, line 56: Looks like you have a lot of hardcoded styles. You have to clean up all that.

modern/src/reports/ChartReportPage.js
```diff
@@ -0,0 +1,74 @@
...
+    }
+  };
+
+  return <ReportFilter handleSubmit={handleSubmit} showOnly />;
+};
+
+const CustomizedAxisTick = ({ x, y, stroke, payload }) =>{
+  return (
+    <g transform={`translate(${x},${y})`}>
+      <text x={0} y={0} dy={16} textAnchor="end" fill="#666" transform="rotate(-35)">{payload.value}</text>
```
tananaev, line 40: Please explain what this is for.

modern/src/reports/ChartType.js
```diff
@@ -0,0 +1,53 @@
...
+    minWidth: 160,
+  },
+}));
+
+const ChartType = ({ type, setType }) => {
+  const classes = useStyles();
+
+  return (
+    <FormControl className={classes.formControl}>
+      <Select value={type} onChange={e => setType(e.target.value)} >
```
tananaev, line 17: I think we should start with a smaller subset for now. I don't think it's a good way to maintain this list specifically for charts. Maybe we can have a list somewhere else as a more generic list that can be converted into select.

tananaev:
Added some comments. As for your questions.

> Since the chart type attribute does not go into the input of the API, I did not include it in the report filter form but added it to the chart component itself.

I think it should still live on the left side. Maybe a separate card below the main one.

> Which is the best way to convert a nested position object to a simple object? Because we need to draw a graph with attributes like distance, totalDistance, etc

How does it work now? Can't you just use a map function?

modern/src/common/formatter.js
mail2bishnoi: Changes reverted. Tried a new solution.

modern/src/reports/ChartReportPage.js
mail2bishnoi: Sorry. Forgot to remove it during code cleanup.

modern/src/reports/ChartReportPage.js
mail2bishnoi: Cleaned up.

modern/src/reports/ChartReportPage.js
mail2bishnoi: I'll do some workaround.

modern/src/reports/ChartReportPage.js
mail2bishnoi: To show the custom X-axis lable.

modern/src/reports/ChartType.js
mail2bishnoi: Removed this component and added chartTypes list and then created a select using chartTypes.

mail2bishnoi:
Hi Anton,

I have tried to incorporate your suggestions. I have shifted the chart type to the left side.
With this, I have tried to format the position. Please review once and give suggestions for the improvements.

An issue related to the chart library is open due to which the problem can sometimes occur in the chart, such as the line going out of the chart boundary.

[https://github.com/recharts/recharts/issues/2149](url)

Best Regards

tananaev:
Can you please provide an updated screenshot.

mail2bishnoi:
PFA
1. Char with no issues.
chart1


2. Chart with the issue in recharts (line going outside chart boundary).
chart2

tananaev:
I was thinking of something like this:
Screen Shot 2020-11-24 at 9 06 30 PM

mail2bishnoi:
It can also be used.

Is the rest of the approach correct?

tananaev:
I think it looks good overall. There are probably some minor comments that we can once you have main structure ready.

The bug in the library is a bit worrying. Are there any other better libraries?

mail2bishnoi:
Ok. Thanks.

I'll look for some other chart library and update you.

mail2bishnoi:
Hi Anton,

Incorporated your suggestions. Please review and suggest improvements.

with this, I have fixed the chart issue by converting speed into the number.

chart3

modern/src/common/chartTypes.js
```diff
@@ -0,0 +1,7 @@
+import t from '../common/localization';
+
+export const chartTypes = [
```
tananaev, line 3: Let's inline this for now in the charts module. I think longer term we want something more general to list all known position attributes.

modern/src/common/formatter.js
```diff
@@ -80,3 +80,26 @@ export const formatVolume = (value, unit) => {
 export const formatHours = (value) => {
   return moment.duration(value).humanize();
 };
+
+const speedConverter = (value, unit) => {
```
tananaev, line 84: There's too much duplication with the functions above. Need to unify that. And also I think we should move converter functions into a separate file.

modern/src/reports/ChartReportPage.js
```diff
@@ -0,0 +1,68 @@
...
+  const speedUnit = useAttributePreference('speedUnit');
+
+  const handleSubmit = async (deviceId, from, to, mail, headers) => {
+    const query = new URLSearchParams({ deviceId, from, to, mail });
+    const response = await fetch(`/api/reports/route?${query.toString()}`, { headers });
+    if (response.ok) {
+      const data = await response.json();
+      let formattedData = data.map((obj)=>{
+        return Object.assign(obj, 
+                              {speed: Number(getConverter('speed')(obj.speed, speedUnit))},
```
tananaev, line 20: Fix formatting here.

modern/src/reports/ChartReportPage.js
```diff
@@ -0,0 +1,68 @@
...
+  const [items, setItems] = useState([]);
+  const [type, setType] = useState('speed');
+
+  return (
+    <ReportLayoutPage reportFilterForm={ReportFilterForm} setItems={setItems} setType={setType} showChartType>
+      <Paper>
+        <Box height={400}>
+          <ResponsiveContainer>
+            <LineChart data={items}>
+              <XAxis dataKey="fixTime" tick={<CustomizedAxisTick/>} height={60} />
```
tananaev, line 54: There are some hardcoded numbers like height and grid. Is there a way to get rid of them? Can you please clarify what each one is for.

modern/src/reports/ReportLayoutPage.js
```diff
@@ -29,6 +35,18 @@ const ReportLayoutPage = ({ reportFilterForm:ReportFilterForm, setItems, childre
             <Paper className={classes.form}>
               <ReportFilterForm setItems={ setItems } />
             </Paper>
+            {showChartType && (
+              <Paper className={classes.chart}>
+                <FormControl variant="filled" margin="normal" fullWidth>
+                <InputLabel>{t('reportChartType')}</InputLabel>
+                <Select defaultValue="speed" onChange={e => setType(e.target.value)}>
+                {chartTypes.map(item => (
```
tananaev, line 43: I think we need indentation here.

modern/src/reports/ReportLayoutPage.js
```diff
@@ -29,6 +35,18 @@ const ReportLayoutPage = ({ reportFilterForm:ReportFilterForm, setItems, childre
             <Paper className={classes.form}>
               <ReportFilterForm setItems={ setItems } />
             </Paper>
+            {showChartType && (
+              <Paper className={classes.chart}>
+                <FormControl variant="filled" margin="normal" fullWidth>
+                <InputLabel>{t('reportChartType')}</InputLabel>
+                <Select defaultValue="speed" onChange={e => setType(e.target.value)}>
```
tananaev, line 42: Looks like we define this default value in state and here. Can we avoid that?

modern/src/common/chartTypes.js
mail2bishnoi: done.

modern/src/common/formatter.js
mail2bishnoi: done.

modern/src/reports/ChartReportPage.js
mail2bishnoi: done.

modern/src/reports/ChartReportPage.js
mail2bishnoi: Box->height is for setting chart height. I have tried setting the chart height using CSS styling but it's not working.

rest of the hardcoded style is chart specific. Xaxis->height is for setting x-axis height so that date and time can be rotated and still visible.

modern/src/reports/ReportLayoutPage.js
mail2bishnoi: done

modern/src/reports/ReportLayoutPage.js
mail2bishnoi: done

modern/src/common/converter.js
```diff
@@ -0,0 +1,24 @@
...
+    case 'mph':
+      factor = 1.15078;
+    case 'kn':
+    default:
+      factor = 1;
+  }
+  return (value * factor).toFixed(2);
+};
+
+export const getConverter = (key) => {
```
tananaev, line 15: I don't think we need this. Please just export separate functions instead.

modern/src/reports/ChartReportPage.js
```diff
@@ -0,0 +1,73 @@
...
+
+  const handleSubmit = async (deviceId, from, to, mail, headers) => {
+    const query = new URLSearchParams({ deviceId, from, to, mail });
+    const response = await fetch(`/api/reports/route?${query.toString()}`, { headers });
+    if (response.ok) {
+
+      const data = await response.json();
+
+      let formattedData = data.map(obj => {
+        return Object.assign(
```
tananaev, line 22: I don't like reusing the same model. Let's create a separate new object with only the data that we need.

modern/src/reports/ChartReportPage.js
```diff
@@ -0,0 +1,73 @@
...
+  const speedUnit = useAttributePreference('speedUnit');
+
+  const handleSubmit = async (deviceId, from, to, mail, headers) => {
+    const query = new URLSearchParams({ deviceId, from, to, mail });
+    const response = await fetch(`/api/reports/route?${query.toString()}`, { headers });
+    if (response.ok) {
+
+      const data = await response.json();
+
+      let formattedData = data.map(obj => {
```
tananaev, line 21: nit Please don't use shortened words. You can use `it` or `position` or `item` here instead.

modern/src/reports/ChartReportPage.js
```diff
@@ -0,0 +1,73 @@
...
+        return Object.assign(
+          obj, 
+          {speed: Number(getConverter('speed')(obj.speed, speedUnit))},
```
tananaev, line 24: Formatting is still off here.

modern/src/common/converter.js
mail2bishnoi: done

modern/src/reports/ChartReportPage.js
mail2bishnoi: done

modern/src/reports/ChartReportPage.js
mail2bishnoi: done

modern/src/reports/ChartReportPage.js
```diff
@@ -0,0 +1,74 @@
...
+  const handleSubmit = async (deviceId, from, to, mail, headers) => {
+    const query = new URLSearchParams({ deviceId, from, to, mail });
+    const response = await fetch(`/api/reports/route?${query.toString()}`, { headers });
+    if (response.ok) {
+
+      const positions = await response.json();
+
+      let formattedPositions = positions.map(position => {
+        return {
+          speed:Number(speedConverter(position.speed, speedUnit)),
```
tananaev, line 23: Can you please make sure formatting matches existing code. You need a space after `:`.

modern/src/reports/ReportLayoutPage.js
```diff
@@ -29,6 +34,18 @@ const ReportLayoutPage = ({ reportFilterForm:ReportFilterForm, setItems, childre
             <Paper className={classes.form}>
               <ReportFilterForm setItems={ setItems } />
             </Paper>
+            {showChartType && (
```
tananaev, line 37: I just noticed one big thing here. I don't think it makes sense to include chart here. It's used only in the chart report.

modern/src/reports/ChartReportPage.js
mail2bishnoi: done.

modern/src/reports/ReportLayoutPage.js
mail2bishnoi: What would you like to suggest where it should be shifted?

Should I move it again to the right side as before?

tananaev: Maybe we need to split layout differently somehow to allow it?

mail2bishnoi: Hi Anton,

I am unable to understand how to implement this? Can you give me a direction?

One solution I thought that includes the Report Filer component and Chart component in a single ChartReportPage.

Thanks

tananaev: Maybe you can pass filter using props. Something like this:
```js
        filter={
          <ReportFilterForm setItems={ setItems } />
          <Paper className={classes.chart}>
            <FormControl variant="filled" margin="normal" fullWidth>
            ...
        }
```

mail2bishnoi: Done. Please check and suggest improvements.

mail2bishnoi:
Hi Anton,

Any updates on this?

tananaev:
Sorry, I added comments, but forgot to submit the review.

modern/src/reports/ReportLayoutPage.js
```diff
@@ -27,7 +27,7 @@ const ReportLayoutPage = ({ reportFilterForm:ReportFilterForm, setItems, childre
         <Grid container spacing={2}>
           <Grid item xs={12} md={3} lg={2}>
             <Paper className={classes.form}>
-              <ReportFilterForm setItems={ setItems } />
+              <ReportFilterForm setItems={ setItems } type={ type } setType={ setType }/>
```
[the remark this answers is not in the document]
mail2bishnoi: Sorry dear. Just forgot.

modern/src/reports/ChartReportPage.js
```diff
@@ -0,0 +1,85 @@
...
+  );
+}
+
+const ChartReportPage = () => {
+
+  const [items, setItems] = useState([]);
+  const [type, setType] = useState('speed');
+
+  return (
+    <ReportLayoutPage reportFilterForm={ReportFilterForm} setItems={setItems} type={type} setType={setType}>
```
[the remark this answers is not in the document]
mail2bishnoi: Chart type

mail2bishnoi:
Hi Anton,

Your suggestion included.

Thank you so much for providing feedback and teaching me how to code on a production-level project.

modern/src/reports/ReportLayoutPage.js
```diff
@@ -18,7 +18,7 @@ const useStyles = makeStyles(theme => ({
   },
 }));
 
-const ReportLayoutPage = ({ reportFilterForm:ReportFilterForm, setItems, children }) => {
+const ReportLayoutPage = ({ reportFilterForm:ReportFilterForm, setItems, type, setType, children }) => {
```
tananaev, line 21: I don't like that we need to pass through a parameter that is not generic. We need to fix that somehow. Maybe component breakdown should be different.

mail2bishnoi:
Hi Anton,

Didn't find any generic solution to the above problem.


Can I use the below solution for passing data and chart type?
`const [items, setItems]=useState({data:[], type:'speed'})`;

tananaev:
Instead of passing a component name to `reportFilterForm` can we pass the whole JSX?

mail2bishnoi:
Hi Anton,

This time I have tried to refactor all the report components to make them more generic. I believe I did it in the best way.

Please review the code and suggest improvements.

Best Regards

tananaev:
Merged. Thanks for the hard work. It looks really good now.

pc500:
I have noticed that the "Speed" value in the modern UI does not correctly show MPH.  It appears to be roughly 8% "slow".  You can see the same from the new vs old UI.

image

image

PR #800, opened 2020-11-27
Update pt_BR.json and pt.json

RafaelMiquelino:
Translate reportDaily string on pt_BR and pt language files.


tananaev:
All localization should be done through Transifex:

PR #806, opened 2021-01-05
Computed Attributes

mail2bishnoi:
Hi Anton,

I have implemented the computed attribute screen. Please review and suggest improvements.

the position attribute list is not complete. I have added only three attributes. If the rest of the things are ok then I'll add the remaining attributes.

### DIFF
--- a/modern/src/settings/ComputedAttributePage.js
+++ b/modern/src/settings/ComputedAttributePage.js
@@ -1,0 +1,92 @@
+import React, { useState } from 'react';
+import { Accordion, AccordionSummary, AccordionDetails, makeStyles, Typography,  FormControl, InputLabel, MenuItem, Select, TextField } from "@material-ui/core";
+import t from '../common/localization';
+import EditItemView from '../EditItemView';
+
+import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
+import positionAttributes from '../attributes/positionAttributes';
+
+
+const useStyles = makeStyles(() => ({
+  details: {
+    flexDirection: 'column',
+  },
+}));
+
+const ComputedAttributePage =() => {
+
+  const classes = useStyles();
+  const [item, setItem] = useState();
+  const [key, setKey] = useState();
+
+  const options = Object.entries(positionAttributes).map(([key, value]) => ({
+    key,
+    name: value.name,
+    type: value.type,
+  }));
+
+  return (
+    <EditItemView endpoint="/attributes/computed" item={item} setItem={setItem}>
+      {item &&
+        <Accordion defaultExpanded>
+          <AccordionSummary expandIcon={<ExpandMoreIcon />}>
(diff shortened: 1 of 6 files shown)


modern/src/settings/ComputedAttributePage.js
```diff
@@ -0,0 +1,92 @@
...
+              variant="filled" />
+            <FormControl variant="filled" margin="normal" fullWidth>
+              <InputLabel>{t('sharedAttribute')}</InputLabel>
+              <Select 
+                value={item.attribute || ''} 
+                onChange={(e) => {
+                  const newValue = e.target.value;
+                  const positionAttribute = positionAttributes[newValue];
+                  setKey(newValue);
+                  if(positionAttribute && positionAttribute.type) {
```
tananaev, line 52: Please fix formatting. Also I would recommend extracting this into a separate method. Too much code for inline method.

tananaev (commented):
Added some comments. Can you please also provide screenshots of both screens.

mail2bishnoi:
Hi Anton,

Your suggestions included.
computed_attribute_list

computed_attribute

modern/src/settings/ComputedAttributesPage.js
```diff
@@ -0,0 +1,69 @@
...
+          <TableCell>{t('sharedDescription')}</TableCell>
+          <TableCell>{t('sharedAttribute')}</TableCell>
+          <TableCell>{t('sharedExpression')}</TableCell>
+          <TableCell>{t('sharedType')}</TableCell>
+        </TableRow>
+      </TableHead>
+      <TableBody>
+        {items.map((item) => (
+          <TableRow key={item.id}>
+            <TableCell className={classes.columnAction} padding="none">
```
tananaev, line 43: We only allow admins to create/edit attributes. Can we hide this and add button if the user is not an admin?

tananaev (commented):
One more comment. Everything else looks good.

mail2bishnoi:
Done.

modern/src/settings/ComputedAttributesPage.js
```diff
@@ -40,11 +42,15 @@ const ComputedAttributeView = ({ updateTimestamp, onMenuClick }) => {
       <TableBody>
         {items.map((item) => (
           <TableRow key={item.id}>
-            <TableCell className={classes.columnAction} padding="none">
-              <IconButton onClick={(event) => onMenuClick(event.currentTarget, item.id)}>
-                <MoreVertIcon />
-              </IconButton>
-            </TableCell>
+            {adminEnabled &&
+              (
```
tananaev, line 46: Do we need these brackets? Same question for other places as well.

mail2bishnoi: No. Removed.

tananaev:
Looks good. Merged. Thanks for all the contributions.

PR #813, opened 2021-01-26
user self registration

mikems66:


modern/src/LoginPage.js
```diff
@@ -118,6 +137,12 @@ const LoginPage = () => {
             </div>
           </FormControl>
         </form>
+        {registerDialogShown && <RegisterDialog open={registerDialogShown} onResult={handleRegisterResult} />}
+        <Snackbar
```
tananaev, line 141: Please follow existing code formatting style.

tananaev: Also, is it possible to move snackbar to the `RegisterDialog` component?

modern/src/RegisterDialog.js
```diff
@@ -0,0 +1,109 @@
...
+          required
+          fullWidth
+          error={validationErrors.password}
+          label={t('userPassword')}
+          name='password'
+          value={formFields.password || ''}
+          type='password'
+          autoComplete='current-password'
+          onChange={handleChange}
+          helperText={validationErrors.password && 'Password is required'} />
```
tananaev, line 98: Please don't use unlocalized text anywhere.

modern/src/RegisterDialog.js
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  const handleRegister = async (event) => {
+    event.preventDefault();
+    setErrorResponse(null);
+    let objErrors = {};
+    if (formFields.name.trim() === '') {
+      objErrors.name = true;
+    }
+    if (!
```
tananaev, line 28: Why is there a new line here?

modern/src/RegisterDialog.js
```diff
@@ -0,0 +1,109 @@
...
+    const response = await fetch('/api/users', {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify(formFields)
+    });
+
+    if (response.ok) {
+      onResult(true)
+    } else {
+      setErrorResponse(t('errorGeneral'));
```
tananaev, line 49: I would skip this error check for now. We need more standardized error handing for all calls.

modern/src/RegisterDialog.js
```diff
@@ -0,0 +1,109 @@
...
+  const [validationErrors, setValidationErrors] = useState({});
+  const [errorResponse, setErrorResponse] = useState(null);
+
+  const handleChange = (event) => {
+    setFormFields({ ...formFields, [event.target.name]: event.target.value });
+    setValidationErrors({ ...validationErrors, [event.target.name]: false });
+  }
+
+  const handleRegister = async (event) => {
+    event.preventDefault();
```
tananaev, line 22: Is this required?

modern/src/RegisterDialog.js
```diff
@@ -0,0 +1,109 @@
...
+import Button from '@material-ui/core/Button';
+import Dialog from '@material-ui/core/Dialog';
+import DialogActions from '@material-ui/core/DialogActions';
+import DialogContent from '@material-ui/core/DialogContent';
+import DialogContentText from '@material-ui/core/DialogContentText';
+import TextField from '@material-ui/core/TextField';
+import Alert from '@material-ui/lab/Alert';
+
+const RegisterDialog = ({ open, onResult }) => {
+  const [formFields, setFormFields] = useState({});
```
tananaev, line 12: Why not have separate fields?

modern/src/RegisterDialog.js
```diff
@@ -0,0 +1,109 @@
...
+  const [validationErrors, setValidationErrors] = useState({});
```
tananaev, line 13: Same here.

modern/src/LoginPage.js
```diff
@@ -47,10 +50,19 @@ const LoginPage = () => {
...
+  const [registerDialogShown, setRegisterDialogShown] = useState(false);
+  const [snackbarOpen, setSnackbarOpen] = useState(false);
 
   const classes = useStyles();
   const history = useHistory();
 
+  const server = useSelector(state => state.session.server);
+
+  let registrationPref = false;
+  if (server) {
```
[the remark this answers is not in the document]
mikems66: Thanks for your feedback.
I did it in the useSelector.

modern/src/LoginPage.js
mikems66: I hope formatting is good now.
I moved snack bar to RegisterDialog.

modern/src/RegisterDialog.js
mikems66: I changed this to existing localized string. If better labels are needed, let me know.

modern/src/RegisterDialog.js
mikems66: removed line

modern/src/RegisterDialog.js
mikems66: Error checking is skipped now.

modern/src/RegisterDialog.js
mikems66: No, removed.

modern/src/RegisterDialog.js
mikems66: I didn't change, I think this way code is more compact.

modern/src/RegisterDialog.js
mikems66: Same as above.

modern/src/RegisterDialog.js
```diff
@@ -0,0 +1,106 @@
...
+      setSnackbarOpen(true);
+    }
+  }
+
+  if (snackbarOpen) {
+    return <Snackbar
+      anchorOrigin={{ vertical: 'bottom', horizontal: 'center', }}
+      open={snackbarOpen} autoHideDuration={6000}
+      onClose={() => { onResult(true) }} message={t('loginCreated')}
+    />
```
tananaev, line 54:
```suggestion
    return (
      <Snackbar
        anchorOrigin={{ vertical: 'bottom', horizontal: 'center' }}
        open={snackbarOpen}
        autoHideDuration={6000}
        onClose={() => { onResult(true) }}
        message={t('loginCreated')} />
    );
```

modern/src/RegisterDialog.js
```diff
@@ -0,0 +1,106 @@
...
+          error={validationErrors.email}
+          label={t('userEmail')}
+          name='email'
+          value={formFields.email || ''}
+          autoComplete='email'
+          onChange={handleChange}
+          helperText={validationErrors.email && t('sharedRequired')} />
+        <TextField
+          margin='normal'
+          required
```
tananaev, line 86: Doesn't thins already handle "required"? Why do we need extra validation?

mikems66: It allowes all blanks to be entered as name and password. But maybe that's not important, so I removed validation (except e-mail).

tananaev: There is also a standard email validation, as far as I know. Can we use that?

modern/src/RegisterDialog.js
tananaev: I think we should separate fields to match what we have in other places. And errors is not needed at all, I think.

mikems66: I will separate fields.

modern/src/RegisterDialog.js
mikems66: If you mean use type='email', seems this doesn't work with Dialog component (submit still works). I could disable submit button like this:  
disabled={!userName || !/(.+)@(.+)\.(.{2,})/.test(email) || !password}  ... let me know if that's ok.

tananaev: Yeah, let's do something like that, but maybe move that regex to a separate method for readability.

tananaev:
Merged, thank you. I will done some minor updates to it on master.

mikems66:
No problem. I could do more stuff, let me know if you have any suggestions.

tananaev:
One of the biggest things that's not implemented yet is geofences, but that might be too complicated.

PR #815, opened 2021-02-04
Initial Maintenance Screen

mail2bishnoi:
Hi Anton,

I have started developing the maintenance screen. However, it is not complete yet. For this, we have to make custom number components. Can you guide me on where and how to start?

Thanks.

### DIFF
--- a/modern/src/attributes/positionAttributes.js
+++ b/modern/src/attributes/positionAttributes.js
@@ -13,8 +13,8 @@ export default {
     name: t('positionIgnition'),
     type: 'boolean',
   },
-  'serviceOdometer': {
-    name: t('positionServiceOdometer'),
+  'odometer': {
+    name: t('positionOdometer'),
     type: 'number',
     dataType: 'distance',
   },
--- a/modern/src/attributes/positionAttributes.js
+++ b/modern/src/attributes/positionAttributes.js
@@ -12,5 +12,10 @@ export default {
   'ignition': {
     name: t('positionIgnition'),
     type: 'boolean',
-  },  
+  },
+  'serviceOdometer': {
+    name: t('positionServiceOdometer'),
+    type: 'number',
+    dataType: 'distance',
+  },
 };
--- a/modern/src/common/converter.js
+++ b/modern/src/common/converter.js
@@ -1,4 +1,4 @@
-export const speedConverter = (value, unit) => {
+export const speedConverter = (value, unit, back = false) => {
   let factor;
   switch (unit) {
     case 'kmh':
@@ -11,10 +11,10 @@ export const speedConverter = (value, unit) => {
     default:
       factor = 1;
   }
-  return (value * factor).toFixed(2);
+  return back ? (value / factor).toFixed(2):(value * factor).toFixed(2);
 };
 
-export const distanceConverter = (value, unit) => {
+export const distanceConverter = (value, unit, back = false) => {
   let factor;
   switch (unit) {
     case 'mi':
@@ -27,5 +27,5 @@ export const distanceConverter = (value, unit) => {
     default:
       factor = 0.001;
   }
-  return (value * factor).toFixed(2);  
+  return back ? value / factor : value * factor;  
 }
--- a/modern/src/common/converter.js
+++ b/modern/src/common/converter.js
@@ -3,11 +3,29 @@ export const speedConverter = (value, unit) => {
   switch (unit) {
     case 'kmh':
       factor = 1.852;
+      break;
     case 'mph':
       factor = 1.15078;
+      break;
     case 'kn':
     default:
       factor = 1;
   }
   return (value * factor).toFixed(2);
 };
+
+export const distanceConverter = (value, unit) => {
+  let factor;
+  switch (unit) {
+    case 'mi':
+      factor = 0.000621371;
+      break;
+    case 'nmi':
+      factor = 0.000539957;
+      break;
+    case 'km':
+    default:
+      factor = 0.001;
+  }
+  return (value * factor).toFixed(2);  
+}
--- a/modern/src/common/converter.js
+++ b/modern/src/common/converter.js
@@ -1,4 +1,4 @@
-export const speedConverter = (value, unit, back = false) => {
+const speedConverter = (value, unit, back = false) => {
   let factor;
   switch (unit) {
     case 'kmh':
@@ -11,10 +11,10 @@ export const speedConverter = (value, unit, back = false) => {
     default:
       factor = 1;
   }
-  return back ? (value / factor).toFixed(2):(value * factor).toFixed(2);
+  return back ? value / factor: value * factor;
 };
 
-export const distanceConverter = (value, unit, back = false) => {
+const distanceConverter = (value, unit, back = false) => {
   let factor;
   switch (unit) {
     case 'mi':
@@ -28,4 +28,20 @@ export const distanceConverter = (value, unit, back = false) => {
       factor = 0.001;
   }
   return back ? value / factor : value * factor;  
+}
+
+export const speedFromKnots = (value, unit) => {
+  return speedConverter(value, unit);
+}
+
+export const speedToKnots = (value, unit) => {
+  return speedConverter(value, unit, true);
+}
+
+export const distanceFromMeters = (value, unit) => {
+  return distanceConverter(value, unit);
+}
+
+export const distanceToMeters = (value, unit) => {
+  return distanceConverter(value, unit, true);
 }
--- a/modern/src/common/converter.js
+++ b/modern/src/common/converter.js
@@ -1,47 +1,39 @@
-const speedConverter = (value, unit, back = false) => {
-  let factor;
+const speedConverter = unit => {
   switch (unit) {
     case 'kmh':
-      factor = 1.852;
-      break;
+      return 1.852;
     case 'mph':
-      factor = 1.15078;
-      break;
+      return 1.15078;
     case 'kn':
     default:
-      factor = 1;
+      return 1;
   }
-  return back ? value / factor: value * factor;
 };
 
-const distanceConverter = (value, unit, back = false) => {
-  let factor;
+const distanceConverter = unit => {
   switch (unit) {
     case 'mi':
-      factor = 0.000621371;
-      break;
+      return 0.000621371;
     case 'nmi':
-      factor = 0.000539957;
-      break;
+      return 0.000539957;
     case 'km':
     default:
-      factor = 0.001;
-  }
-  return back ? value / factor : value * factor;  
+      return 0.001;
+  } 
 }
 
 export const speedFromKnots = (value, unit) => {
-  return speedConverter(value, unit);
+  return value * speedConverter(unit);
 }
 
 export const speedToKnots = (value, unit) => {
-  return speedConverter(value, unit, true);
+  return value / speedConverter(unit);
 }
 
 export const distanceFromMeters = (value, unit) => {
-  return distanceConverter(value, unit);
+  return value * distanceConverter(unit);
 }
 
 export const distanceToMeters = (value, unit) => {
-  return distanceConverter(value, unit, true);
+  return value / distanceConverter(unit);
 }
--- a/modern/src/settings/MaintenancePage.js
+++ b/modern/src/settings/MaintenancePage.js
@@ -7,6 +7,7 @@ import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
 import EditAttributesView from '../attributes/EditAttributesView';
 import positionAttributes from '../attributes/positionAttributes';
 import { useAttributePreference } from '../common/preferences';
+import { distanceConverter, speedConverter } from '../common/converter';
 
 const useStyles = makeStyles(() => ({
   details: {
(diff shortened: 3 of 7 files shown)


tananaev:
Can you please post some screenshots.

What do you mean by a custom number component?

mail2bishnoi:
Hi Anton,

Please look at the marked fields.

This one is I'm developing

under_development


and this is the existing screen

existing

tananaev:
You can probably just add it to the label or something. There is no need to overcomplicate things.

mail2bishnoi:
Hi Anton,

I am working on the maintenance screen. But while editing the existing record, I'm facing problems in converting the start and period fields according to the user preferences when the screen loads the first time. With this, when creating a new record, we have to convert the start and period fields according to the user preferences before making an API call.
Can you please help me with how to handle it?

tananaev:
What is the problem?

mail2bishnoi:
Hi Anton,

The first problem is that when creating a record, on clicking the save button, we directly pass the item state to the API (in handleSave method of EditItemView.js), but now we have to convert the start and period(in case of speed and distance) attribute of the item state before passing to API. Where to edit the attributes of the item.

The other problem is to convert the start and period when editing the record, but only when the screen loads the first time. I am unable to build the logic of conversion when the screen loads the first time.

tananaev:
You can convert from raw value to display value when you need to display and back when you store new value. I'm not sure I understand what the difficulty is.

mail2bishnoi:
Hi Anton 

I have completed the maintenance screen. Please review and suggest improvements.

tananaev:
Please upload updated screenshots.

mail2bishnoi:
maintenance_creaate


maintenance_list

modern/src/attributes/positionAttributes.js
```diff
@@ -12,5 +12,10 @@ export default {
   'ignition': {
     name: t('positionIgnition'),
     type: 'boolean',
-  },  
+  },
+  'serviceOdometer': {
```
tananaev, line 16: This is rarely used attribute. I would recommend to use "odometer" instead.

modern/src/common/converter.js
```diff
@@ -1,13 +1,31 @@
-export const speedConverter = (value, unit) => {
+export const speedConverter = (value, unit, back = false) => {
```
tananaev, line 1: I think we should split functions. For example make `speedFromKnots` and `speedToKnots` and then keep the factor switch in some shared internal function.

modern/src/common/converter.js
```diff
@@ -1,13 +1,31 @@
...
+      break;
     case 'kn':
     default:
       factor = 1;
   }
-  return (value * factor).toFixed(2);
+  return back ? (value / factor).toFixed(2):(value * factor).toFixed(2);
 };
+
+export const distanceConverter = (value, unit, back = false) => {
```
tananaev, line 17: Same here.

modern/src/settings/MaintenancesPage.js
```diff
@@ -0,0 +1,91 @@
...
+  const distanceUnit = useAttributePreference('distanceUnit');
+
+  useEffectAsync(async () => {
+    const response = await fetch('/api/maintenance');
+    if (response.ok) {
+      setItems(await response.json());
+    }
+  }, [updateTimestamp]);
+
+  const convertAttribute = (key, value) => {
```
tananaev, line 34: Don't we have something similar already?

modern/src/settings/MaintenancePage.js
```diff
@@ -0,0 +1,158 @@
...
+
+  const options = [];
+
+  Object.entries(positionAttributes).map(([key, value]) => {
+    if (value.type === 'number') {
+      options.push({key, name: value.name, type: value.type})
+    }
+  });
+
+  const handleChange = event => {
```
tananaev, line 35: We should probably rename it to be more clear.

modern/src/settings/MaintenancePage.js
```diff
@@ -0,0 +1,158 @@
...
+  const classes = useStyles();
+  const [item, setItem] = useState();
+  const [labels, setLabels] = useState({start: '', period: ''});
+
+  const speedUnit = useAttributePreference('speedUnit');
+  const distanceUnit = useAttributePreference('distanceUnit');
+
+  const options = [];
+
+  Object.entries(positionAttributes).map(([key, value]) => {
```
tananaev, line 29: This seems unnecessary to call every time.

modern/src/settings/MaintenancePage.js
```diff
@@ -0,0 +1,158 @@
...
+        case 'speed':
+          setLabels({ ...labels, start: speedUnit, period: speedUnit});
+          break;
+        default:
+          break;
+      }
+    }
+  }
+
+  const rawToValue = (rawValue) => {
```
tananaev, line 54: Seems like a lot of duplication with `valueToRaw`.

modern/src/settings/MaintenancePage.js
```diff
@@ -0,0 +1,158 @@
...
+
+  const handleChange = event => {
+    const newValue = event.target.value;
+    setItem({...item, type: newValue, start: 0, period: 0});
+
+    const attribute = positionAttributes[newValue];
+    if (attribute && attribute.dataType) {
+      switch (attribute.dataType) {
+        case 'distance':
+          setLabels({ ...labels, start: distanceUnit, period: distanceUnit});
```
tananaev, line 43: Labels need to be localized.

tananaev (commented):
Added some comments. There's probably more issues that we would need to fix.

modern/src/attributes/positionAttributes.js
mail2bishnoi: done

modern/src/common/converter.js
mail2bishnoi: done

modern/src/common/converter.js
mail2bishnoi: done

modern/src/settings/MaintenancePage.js
mail2bishnoi: done

modern/src/settings/MaintenancePage.js
mail2bishnoi: Done.

modern/src/settings/MaintenancesPage.js
mail2bishnoi: I didn't find anything.

modern/src/settings/MaintenancePage.js
mail2bishnoi: Tried to use useEffect hook but without success.

mail2bishnoi:
Hi Anton,

Tried to incorporate your suggestions. Please review.

modern/src/common/converter.js
```diff
@@ -1,13 +1,47 @@
-export const speedConverter = (value, unit) => {
+const speedConverter = (value, unit, back = false) => {
```
tananaev, line 1: I think these functions should just return `factor`. It will make the code shorted. No breaks needed.

modern/src/common/converter.js
```diff
@@ -1,13 +1,47 @@
...
+      break;
     case 'kn':
     default:
       factor = 1;
   }
-  return (value * factor).toFixed(2);
+  return back ? value / factor: value * factor;
 };
+
+const distanceConverter = (value, unit, back = false) => {
```
tananaev, line 17: Same here.

modern/src/settings/MaintenancesPage.js
```diff
@@ -0,0 +1,91 @@
...
+          <TableCell>{t('sharedType')}</TableCell>
+          <TableCell>{t('maintenanceStart')}</TableCell>
+          <TableCell>{t('maintenancePeriod')}</TableCell>
+        </TableRow>
+      </TableHead>
+      <TableBody>
+        {items.map(item => (
+          <TableRow key={item.id}>
+            <TableCell className={classes.columnAction} padding="none">
+              <IconButton onClick={(event) => onMenuClick(event.currentTarget, item.id)}>
```
tananaev, line 66:
```suggestion
              <IconButton onClick={event => onMenuClick(event.currentTarget, item.id)}>
```

modern/src/settings/MaintenancePage.js
```diff
@@ -0,0 +1,161 @@
...
+              </FormControl>            
+              <TextField
+                margin="normal"
+                type="number"
+                value={rawToValue(item.start) || ''}
+                onChange={event => setItem({...item, start: valueToRaw(event.target.value)})}
+                label={t('maintenanceStart')}
+                variant="filled"
+                InputProps={{
+                endAdornment: <InputAdornment position="start">{labels.start}</InputAdornment>,
```
tananaev, line 127: What does `position` mean here?

modern/src/settings/MaintenancePage.js
```diff
@@ -0,0 +1,161 @@
...
+      if (value.type === 'number') {
+        otherList.push({key, name: value.name, type: value.type});
+      }
+    }
+    return otherList;
+  }
+
+  const onMaintenanceTypeChange = event => {
+    const newValue = event.target.value;
+    setItem({...item, type: newValue, start: 0, period: 0});
```
tananaev, line 41:
```suggestion
    setItem({ ...item, type: newValue, start: 0, period: 0 });
```

modern/src/settings/MaintenancePage.js
```diff
@@ -0,0 +1,161 @@
...
+  const rawToValue = (value) => {
+
+    const attribute = positionAttributes[item.type];
+    if (attribute && attribute.dataType) {
+      switch (attribute.dataType) {
+        case 'speed':
+          return speedFromKnots(value, speedUnit);
+        case 'distance':
+          return distanceFromMeters(value, distanceUnit);
+        default:
```
tananaev, line 67: Do we need default if we already have "global" return?

modern/src/settings/MaintenancePage.js
```diff
@@ -0,0 +1,161 @@
...
+  const valueToRaw = (value) => {
+
+    const attribute = positionAttributes[item.type];
+    if (attribute && attribute.dataType) {
+      switch (attribute.dataType) {
+        case 'speed':
+          return speedToKnots(value, speedUnit);
+        case 'distance':
+          return distanceToMeters(value, distanceUnit);
+        default:
```
tananaev, line 84: Same here.

modern/src/settings/MaintenancePage.js
```diff
@@ -0,0 +1,161 @@
...
+        case 'speed':
+          setLabels({ ...labels, start: t(prefixString('shared', speedUnit)), period: t(prefixString('shared', speedUnit))});
+          break;
+        default:
+          break;
+      }
+    }
+  }
+
+  const rawToValue = (value) => {
```
tananaev, line 58:
```suggestion
  const rawToValue = value => {
```

modern/src/settings/MaintenancePage.js
```diff
@@ -0,0 +1,161 @@
...
+          return distanceFromMeters(value, distanceUnit);
+        default:
+          return value;
+      }
+    }
+
+    return value;
+  }
+
+  const valueToRaw = (value) => {
```
tananaev, line 75:
```suggestion
  const valueToRaw = value => {
```

tananaev (commented):
Added some minor comments. Mostly formatting.

modern/src/common/converter.js
mail2bishnoi: Done.

modern/src/common/converter.js
mail2bishnoi: Done.

modern/src/settings/MaintenancesPage.js
mail2bishnoi: Done.

modern/src/settings/MaintenancePage.js
mail2bishnoi: image

modern/src/settings/MaintenancePage.js
mail2bishnoi: Done.

modern/src/settings/MaintenancePage.js
mail2bishnoi: Removed.

modern/src/settings/MaintenancePage.js
mail2bishnoi: Removed.

modern/src/settings/MaintenancePage.js
mail2bishnoi: Done.

modern/src/settings/MaintenancePage.js
mail2bishnoi: Done.

mail2bishnoi:
Hi Anton,

Incorporated your suggestions. Please review.

modern/src/settings/MaintenancePage.js
tananaev: But isn't it at the end on the screenshot?

mail2bishnoi: This position attribute work in the top-down direction. For start and end, we use startAndornment and endAdornment.

image

tananaev:
Merged, thank you.

PR #818, opened 2021-02-17
Initial Statistics Screen

mail2bishnoi:
Hi Anton,

Implemented Statistics Screen. Please review and suggest improvements.

### DIFF
--- a/modern/src/reports/ReportFilter.js
+++ b/modern/src/reports/ReportFilter.js
@@ -4,7 +4,7 @@ import t from '../common/localization';
 import { useSelector } from 'react-redux';
 import moment from 'moment';
 
-const ReportFilter = ({ children, handleSubmit, showOnly }) => {
+const ReportFilter = ({ children, handleSubmit, showOnly, showDevices }) => {
   const devices = useSelector(state => Object.values(state.devices.items));
   const [deviceId, setDeviceId] = useState();
   const [period, setPeriod] = useState('today');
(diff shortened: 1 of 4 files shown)


modern/src/reports/ReportFilter.js
```diff
@@ -4,7 +4,7 @@ import t from '../common/localization';
 import { useSelector } from 'react-redux';
 import moment from 'moment';
 
-const ReportFilter = ({ children, handleSubmit, showOnly }) => {
+const ReportFilter = ({ children, handleSubmit, showOnly, showDevices }) => {
```
tananaev, line 7: If you want to re-use the filter, I would recommend adding more nesting. I don't like adding another configuration here.

For example, you can have `ReportFilter` without device and `DeviceReportFilter` with device that uses `ReportFilter`.

tananaev (commented):
Please also share screenshots.

modern/src/reports/ReportFilter.js
mail2bishnoi: I would not like to complicate things for now. Just separated the statistics page and removed the configuration from ReportFilter.

image

tananaev: Can we revert these changes then?

mail2bishnoi: Done.

tananaev: Doesn't looks like it's reverted.

mail2bishnoi:
Something went bad. Closing this PR and opening a new one.

mail2bishnoi:
Closing this PR and opening a newone.

PR #822, opened 2021-03-05
Transform circle fence to polygon fence for display

mikems66:


modern/src/map/GeofenceMap.js
```diff
@@ -4,6 +4,15 @@ import { useEffect, useState } from 'react';
 import { map } from './Map';
 import { useEffectAsync } from '../reactHelper';
 import { reverseCoordinates } from './mapUtil';
+import circle from '@turf/circle';
+
+
+const circleToPolygon = (item) => {
+  let coord = item.area.replace(/CIRCLE|\(|\)|,/g, " ").trim().split(/ +/);
```
tananaev, line 11:
```suggestion
  let coordinates = item.area.replace(/CIRCLE|\(|\)|,/g, " ").trim().split(/ +/);
```

modern/src/map/GeofenceMap.js
```diff
@@ -4,6 +4,15 @@ import { useEffect, useState } from 'react';
...
+  var options = {steps: 64, units: 'meters'};
```
tananaev, line 12:
```suggestion
  var options = { steps: 64, units: 'meters' };
```

I would also reduce the number of steps.

modern/src/map/GeofenceMap.js
```diff
@@ -4,6 +4,15 @@ import { useEffect, useState } from 'react';
...
+  let c = circle([Number(coord[1]), Number(coord[0])], Number(coord[2]), options);
```
tananaev, line 13: Please use full words for variable names. Maybe `polygon` or something here.

modern/src/map/GeofenceMap.js
```diff
@@ -72,9 +81,10 @@ const GeofenceMap = () => {
   }, []);
 
   useEffect(() => {
+    let circleFence = geofences.filter((item => item.area.indexOf('CIRCLE') > -1)).map(circleToPolygon);
     map.getSource(id).setData({
       type: 'FeatureCollection',
-      features: geofences.map(item => [item.name, reverseCoordinates(wellknown(item.area))]).filter(([, geometry]) => !!geometry).map(([name, geometry]) => ({
```
tananaev, line 77: Should we implement this circle logic in `wellknown` instead? Where is that defined?

modern/src/map/GeofenceMap.js
mikems66: ok

modern/src/map/GeofenceMap.js
mikems66: ok, steps is now 32

modern/src/map/GeofenceMap.js
mikems66: ok

modern/src/map/GeofenceMap.js
mikems66: "wellknown" is a Node module that creates geojson object from text, and as I understand it, it doesn't handle circles, so I don't think we can use it.

tananaev: In that case we should probably move the whole logic outside of this UI component.

mikems66: Is this ok; 
new function createGeofenceFeatures in map/mapUtil.js, which will create features for all geometry types.

tananaev: I think you forgot to push the changes.

mikems66: Here it is, actually new function name is geofenceToFeature.

tananaev:
Looks good. Merged. Thank you for the contribution.

PR #823, opened 2021-03-09
Initial Device Driver Mapping Screen Development

mail2bishnoi:
Hi Anton,

I have tried to create a device and driver mapping screen. Although this could be provided in the connection part of the device's screen, I find this way more user-friendly. If the design of this screen is correct, then I will try to make it more generic so that the code can be reused in the remaining mapping screens like Computed Attributes, Saved Commands, etc.

device_driver_mapping

device_driver_screen


tananaev:
This is not how we assign something to devices anymore. You need something like this:

https://github.com/traccar/traccar-web/blob/master/modern/src/DevicePage.js#L121-L129

PR #828, opened 2021-03-15
Initial draft of cutome route table

mail2bishnoi:
Hi Anton,

I have started the development of a custom route table for the show and hide the columns. Please review and suggest changes.

### DIFF
--- a/modern/src/common/style.js
+++ b/modern/src/common/style.js
@@ -1,0 +1,1 @@
+export const columnWidthNormal = 100;
--- a/modern/src/common/style.js
+++ b/modern/src/common/style.js
@@ -1,1 +1,0 @@
-export const columnWidthNormal = 100;
--- a/modern/src/reports/RouteReportPage.js
+++ b/modern/src/reports/RouteReportPage.js
@@ -1,10 +1,10 @@
 import React, { useState } from 'react';
 import { DataGrid } from '@material-ui/data-grid';
 import t from '../common/localization';
-import { formatPosition, formatDistance, formatSpeed, formatBoolean, formatDate, formatNumber } from '../common/formatter';
+import { formatDistance, formatSpeed, formatBoolean, formatDate, formatNumber, coordinateFormatter } from '../common/formatter';
 import ReportFilter from './ReportFilter';
 import ReportLayoutPage from './ReportLayoutPage';
-import { useAttributePreference } from '../common/preferences';
+import { useAttributePreference, usePreference } from '../common/preferences';
 
 const Filter = ({ setItems }) => {
 
@@ -29,6 +29,7 @@ const Filter = ({ setItems }) => {
 const RouteReportPage = () => {
   const distanceUnit = useAttributePreference('distanceUnit');
   const speedUnit = useAttributePreference('speedUnit');
+  const coordinateFormat = usePreference('coordinateFormat');
 
   const columns = [{
     headerName: t('positionFixTime'),
@@ -41,13 +42,13 @@ const RouteReportPage = () => {
     field: 'latitude',
     type: 'number',
     flex: 1,
-    valueFormatter: ({ value }) => formatNumber(value, 5),
+    valueFormatter: ({ value }) => coordinateFormatter('latitude', value, coordinateFormat),
   }, {
     headerName: t('positionLongitude'),
     field: 'longitude',
     type: 'number',
     flex: 1,
-    valueFormatter: ({ value }) => formatNumber(value, 5),
+    valueFormatter: ({ value }) => coordinateFormatter('longitude', value, coordinateFormat),
   }, {
     headerName: t('positionSpeed'),
     field: 'speed',
--- a/modern/src/reports/RouteReportPage.js
+++ b/modern/src/reports/RouteReportPage.js
@@ -59,7 +59,6 @@ const RouteReportPage = () => {
     field: 'address',
     type: 'string',
     flex: 1,
-    valueFormatter: ({ value }) => formatPosition(value, 'address'),
   }, {
     headerName: t('positionIgnition'),
     field: 'ignition',
--- a/modern/src/reports/RouteReportPage.js
+++ b/modern/src/reports/RouteReportPage.js
@@ -29,49 +29,49 @@ const Filter = ({ setItems }) => {
 const RouteReportPage = () => {
   const distanceUnit = useAttributePreference('distanceUnit');
 
-  const columns = [
-    {
+  const columns = [{
       headerName: t('positionFixTime'),
       field: 'fixTime',
+      type: 'dateTime',
       flex: 1,
-      valueFormatter: params => formatPosition(params.value, 'fixTime'),
-    },
-    {
+      valueFormatter: ({ value }) => formatPosition(value, 'fixTime'),
+    }, {
       headerName: t('positionLatitude'),
       field: 'latitude',
+      type: 'number',
       flex: 1,
-      valueFormatter: params => formatPosition(params.value, 'latitude'),
-    },
-    {
+      valueFormatter: ({ value }) => formatPosition(value, 'latitude'),
+    }, {
       headerName: t('positionLongitude'),
       field: 'longitude',
+      type: 'number',
       flex: 1,
-      valueFormatter: params => formatPosition(params.value, 'longitude'),
-    },
-    {
+      valueFormatter: ({ value }) => formatPosition(value, 'longitude'),
+    }, {
       headerName: t('positionSpeed'),
       field: 'speed',
+      type: 'number',
       flex: 1,
-      valueFormatter: params => formatPosition(params.value, 'speed'),
-    },
-    {
+      valueFormatter: ({ value }) => formatPosition(value, 'speed'),
+    }, {
       headerName: t('positionAddress'),
       field: 'address',
+      type: 'string',
       flex: 1,
-      valueFormatter: params => formatPosition(params.value, 'address'),
-    },
-    {
+      valueFormatter: ({ value }) => formatPosition(value, 'address'),
+    }, {
       headerName: t('positionIgnition'),
       field: 'ignition',
+      type: 'boolean',
       flex: 1,
-      valueFormatter: params => params.getValue('attributes').ignition ? 'Yes' : 'No',
-    },
-    {
+      valueFormatter: ({ getValue }) => getValue('attributes').ignition ? 'Yes' : 'No',
+    }, {
       headerName: t('deviceTotalDistance'),
-      hide: true,
       field: 'totalDistance',
+      type: 'number',
+      hide: true,
       flex: 1,
-      valueFormatter: params => formatDistance(params.getValue('attributes').totalDistance, distanceUnit),
+      valueFormatter: ({ getValue }) => formatDistance(getValue('attributes').totalDistance, distanceUnit),
     },    
   ]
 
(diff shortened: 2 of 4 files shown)


modern/src/reports/RouteReportPage.js
```diff
@@ -26,35 +27,59 @@ const Filter = ({ setItems }) => {
...
+      headerName: t('positionIgnition'),
+      field: 'ignition',
+      width: 130,
+      valueFormatter: params => params.getValue('attributes').ignition ? 'Yes' : 'No',
+    },
+    {
+      headerName: t('deviceTotalDistance'),
+      hide: true,
+      field: 'totalDistance',
+      width: 160,
```
tananaev, line 73: What can we do with hardcoded values?

modern/src/reports/RouteReportPage.js
```diff
@@ -26,35 +27,59 @@ const Filter = ({ setItems }) => {
...
-                <TableCell>{formatPosition(item, 'latitude')}</TableCell>
-                <TableCell>{formatPosition(item, 'longitude')}</TableCell>
-                <TableCell>{formatPosition(item, 'speed')}</TableCell>
-                <TableCell>{formatPosition(item, 'address')}</TableCell>
-              </TableRow>
-            ))}
-          </TableBody>
-        </Table>
-      </TableContainer>
+      <DataGrid rows={items} columns={columns} pageSize={25}/>
```
tananaev, line 82: Same here.

mail2bishnoi:
Hi Anton,

Done some changes. Please review.

modern/src/reports/RouteReportPage.js
mail2bishnoi: Removed pagignation

modern/src/reports/RouteReportPage.js
mail2bishnoi: Done Some changes.

modern/src/common/style.js
```diff
@@ -0,0 +1 @@
+export const columnWidthNormal = 100;
```
tananaev, line 1: No, I don't like this solution. We should figure out a way without hardcoded width.

mail2bishnoi:
Hi Anton,

Done some changes.

How should I define the minimum width for the column so that it can completely visible on the screen?

modern/src/reports/RouteReportPage.js
```diff
@@ -26,35 +27,63 @@ const Filter = ({ setItems }) => {
 };
 
 const RouteReportPage = () => {
+  const distanceUnit = useAttributePreference('distanceUnit');
+
+  const columns = [
```
tananaev, line 32: Let's format it like this:
```
const columns = [{
  ...
}, {
  ...
} ...
```

modern/src/reports/RouteReportPage.js
```diff
@@ -26,35 +27,63 @@ const Filter = ({ setItems }) => {
...
+    {
+      headerName: t('positionFixTime'),
+      field: 'fixTime',
+      flex: 1,
+      valueFormatter: params => formatPosition(params.value, 'fixTime'),
```
tananaev, line 37: Can we do deconstruction like this:
```
valueFormatter: ({ value }) => ...
```

tananaev: We should also specify types for each column.

modern/src/reports/RouteReportPage.js
```diff
@@ -26,35 +27,63 @@ const Filter = ({ setItems }) => {
...
+    },
+    {
+      headerName: t('positionLatitude'),
+      field: 'latitude',
+      flex: 1,
```
tananaev, line 42: Is there a way to default to 1?

modern/src/reports/RouteReportPage.js
```diff
@@ -26,35 +27,63 @@ const Filter = ({ setItems }) => {
...
-                <TableCell>{formatPosition(item, 'latitude')}</TableCell>
-                <TableCell>{formatPosition(item, 'longitude')}</TableCell>
-                <TableCell>{formatPosition(item, 'speed')}</TableCell>
-                <TableCell>{formatPosition(item, 'address')}</TableCell>
-              </TableRow>
-            ))}
-          </TableBody>
-        </Table>
-      </TableContainer>
+        <DataGrid
```
tananaev, line 82: Please fix indentation.

mail2bishnoi:
Hi Anton,

Incorporated your suggestions. Please review.

modern/src/reports/RouteReportPage.js
```diff
@@ -26,35 +27,63 @@ const Filter = ({ setItems }) => {
 };
 
 const RouteReportPage = () => {
+  const distanceUnit = useAttributePreference('distanceUnit');
+
+  const columns = [{
+      headerName: t('positionFixTime'),
```
tananaev, line 33: The indentation is wrong.

modern/src/reports/RouteReportPage.js
```diff
@@ -26,35 +27,63 @@ const Filter = ({ setItems }) => {
...
+      flex: 1,
+      valueFormatter: ({ getValue }) => getValue('attributes').ignition ? 'Yes' : 'No',
+    }, {
+      headerName: t('deviceTotalDistance'),
+      field: 'totalDistance',
+      type: 'number',
+      hide: true,
+      flex: 1,
+      valueFormatter: ({ getValue }) => formatDistance(getValue('attributes').totalDistance, distanceUnit),
+    },
```
tananaev, line 75: It should also end with `}]`.

modern/src/reports/RouteReportPage.js
tananaev: You also haven't answered this.

mail2bishnoi:
Hi Anton,

Code formatting is done.

flex: 1 -> Could not find any way to set it to 1 by default.

https://material-ui.com/components/data-grid/columns/#fluid-width

modern/src/reports/RouteReportPage.js
```diff
@@ -26,35 +27,62 @@ const Filter = ({ setItems }) => {
...
+    type: 'boolean',
+    flex: 1,
+    valueFormatter: ({ getValue }) => getValue('attributes').ignition ? 'Yes' : 'No',
+  }, {
+    headerName: t('deviceTotalDistance'),
+    field: 'totalDistance',
+    type: 'number',
+    hide: true,
+    flex: 1,
+    valueFormatter: ({ getValue }) => formatDistance(getValue('attributes').totalDistance, distanceUnit),
```
tananaev, line 74: What is `getValue` and how does grid get value for this column?

mail2bishnoi: In DataGrid we pass rows as group of objects in an array.  getValue is a function which is used to get the value from current object by passing a key. In this case we pass the key 'attributes' which return attribute object.

tananaev: I feel like the right solution is to use [valueGetter](https://material-ui.com/components/data-grid/rendering/#value-getter). Otherwise wouldn't ordering/filtering not work if you don't get proper value?

mail2bishnoi: Ok. I'll do changes accordingly.

mail2bishnoi:
Hi Anton,

Changes are done. Please review.

modern/src/reports/RouteReportPage.js
```diff
@@ -26,35 +27,62 @@ const Filter = ({ setItems }) => {
...
+    field: 'address',
+    type: 'string',
+    flex: 1,
+    valueFormatter: ({ value }) => formatPosition(value, 'address'),
+  }, {
+    headerName: t('positionIgnition'),
+    field: 'ignition',
+    type: 'boolean',
+    flex: 1,
+    valueGetter: ({ getValue }) => getValue('attributes').ignition ? 'Yes' : 'No',
```
tananaev, line 67: I think you misunderstood how it supposed to work. You need both `valueGetter` and `valueFormatter`. One gives you raw data and then another one formats it for display. Also, using `getValue` in `valueGetter` kind of defeats the purpose, I think.

mail2bishnoi: According to Material UI docs-

field is the only required property since it's the column identifier. It's also used to match with GridRowData values.

About valueGetter - Sometimes a column might not have a corresponding value and you just want to render a combination of different fields. To do that, you can set the valueGetter attribute of GridColDef as in the example below:

Note: You need to set a sortComparator for the column sorting to work when setting the valueGetter attribute.

https://material-ui.com/components/data-grid/rendering/#header-cell

mail2bishnoi: In our case, we don't have totalDistance and ignition as direct fields but these are the properties of attributes. I think I have to implement it in this way only.

tananaev: In our case we don't have direct fields, but we do have values. We just need proper getters for those values.

mail2bishnoi:
Hi Anton,

I have tried one more refactoring. Please review.

modern/src/reports/RouteReportPage.js
```diff
@@ -26,35 +27,64 @@ const Filter = ({ setItems }) => {
...
+    type: 'string',
+    flex: 1,
+    valueFormatter: ({ value }) => formatPosition(value, 'address'),
+  }, {
+    headerName: t('positionIgnition'),
+    field: 'ignition',
+    type: 'boolean',
+    flex: 1,
+    valueGetter: ({ row }) => row.attributes.ignition,
+    valueFormatter: ({ value }) => value ? 'Yes' : 'No',
```
tananaev, line 68: OK, I think we are almost there, but we shouldn't have hardcoded values like this. By the way, instead of `formatPosition` can we use specific formatters directly?

mail2bishnoi:
Hi Anton, 

Please check now.

modern/src/reports/RouteReportPage.js
```diff
@@ -26,35 +27,65 @@ const Filter = ({ setItems }) => {
...
+    field: 'speed',
+    type: 'number',
+    flex: 1,
+    valueFormatter: ({ value }) => formatSpeed(value, speedUnit),
+  }, {
+    headerName: t('positionAddress'),
+    field: 'address',
+    type: 'string',
+    flex: 1,
+    valueFormatter: ({ value }) => formatPosition(value, 'address'),
```
tananaev, line 62: This doesn't require formatting at all.

modern/src/reports/RouteReportPage.js
```diff
@@ -26,35 +27,65 @@ const Filter = ({ setItems }) => {
...
+    field: 'fixTime',
+    type: 'dateTime',
+    flex: 1,
+    valueFormatter: ({ value }) => formatDate(value),
+  }, {
+    headerName: t('positionLatitude'),
+    field: 'latitude',
+    type: 'number',
+    flex: 1,
+    valueFormatter: ({ value }) => formatNumber(value, 5),
```
tananaev, line 44: Can we create a separate method for coordinates formatting?

modern/src/reports/RouteReportPage.js
```diff
@@ -26,35 +27,65 @@ const Filter = ({ setItems }) => {
...
+  }, {
+    headerName: t('positionLongitude'),
+    field: 'longitude',
+    type: 'number',
+    flex: 1,
+    valueFormatter: ({ value }) => formatNumber(value, 5),
```
tananaev, line 50: Same here.

tananaev (commented):
Promise it's the last comments.

mail2bishnoi:
Created separate methods for coordinate formatting. please check and review.

mail2bishnoi:
Done. Please check.

tananaev:
Merged. Thanks for fixing all the comments.

mail2bishnoi:
Thank you.

Should I replace tables in all the reports with DataGrid?

tananaev:
Makes sense to me.

mail2bishnoi:
ok. Will open a separate PR next week.

Issue #827, opened 2021-03-15
Open street maps are blurry and have poor quality in the new UI compared to openstreetmaps.org and the old UI


pc500:
**Describe the bug**
Open street maps are blurry and have poor quality in the new UI compared to openstreetmaps.org and the old UI

**To Reproduce**
Open a map in the new UI + old UI as shown below, on a 1080p monitor with no DPI scaling (100%) enabled on Windows.

![image](https://user-images.githubusercontent.com/33823824/111091728-28b70580-84f1-11eb-8616-3543ef3423bf.png)

**Expected behavior**
A clearer map as shown in the old UI

**Screenshots**

**Desktop (please complete the following information):**
Windows 10, latest auto-updated Chrome (as of this writing).

**Smartphone (please complete the following information):**

**Additional context**
See https://www.traccar.org/forums/topic/open-street-maps-are-blurry-and-have-poor-quality-in-the-new-ui-compared-to-openstreetmapsorg/#post-60637 for more info.

mtahasahin:
I looked into this bug and found that it is a result of two things:

1. OSM tile size is not specified in the code. MapboxGL defaults to 512 while it should be 256. Fixed in https://github.com/traccar/traccar-web/pull/829
2. Since it is a raster map the zoom level should always be an integer. Otherwise the map will be stretched and look blurry. This is not specific to OSM, any raster map will suffer from this.

I could think of two possible solutions. 
1. Allow only integer zoom levels for Map component. This will solve the problem for raster maps but also limit vector maps to integer zoom levels. (While vector maps can be displayed in arbitrary zoom levels because, well, they are vectors.) Easy to implement.
2. If a raster map style is selected allow only integer zoom levels. Otherwise allow any zoom level. This is a better solution but might add a little complexity to code.

Your thoughts? @tananaev

tananaev:
Is it possible to limit zoom levels only for easter maps? If yes, let's do it.

PR #829, opened 2021-03-30
Added tileSize attribute for OSM map style

mtahasahin:
In the style object for OSM, tileSize attribute is not specified, and MapboxGL defaults to 512x512(which should be 256x256 for OSM) and this causes a blurry map as mentioned at https://github.com/traccar/traccar-web/issues/827. This PR improves map quality by fixing it and eliminates one of the two reasons that causes blurry maps for OSM.  The other reason is related to non-integer zoom levels for raster tiles and requires a more complex solution.

### DIFF
--- a/modern/src/map/mapStyles.js
+++ b/modern/src/map/mapStyles.js
@@ -1,11 +1,11 @@
-export const styleCustom = (url, attribution, tileSize) => ({
+export const styleCustom = (url, attribution) => ({
   version: 8,
   sources: {
     osm: {
       type: 'raster',
       tiles: [url],
       attribution: attribution,
-      tileSize: tileSize,
+      tileSize: 256,
     },
   },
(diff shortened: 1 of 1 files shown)


modern/src/map/mapStyles.js
```diff
@@ -1,10 +1,11 @@
-export const styleCustom = (url, attribution) => ({
+export const styleCustom = (url, attribution, tileSize) => ({
   version: 8,
   sources: {
     osm: {
       type: 'raster',
       tiles: [url],
       attribution: attribution,
+      tileSize: tileSize,
```
tananaev, line 8: I think it's better to just hardcode 256 for now. Is there any place where we use different number?

mtahasahin: No, a different number is not used. I hardcoded the value as you said.

tananaev:
Merged, thank you.

PR #830, opened 2021-03-31
Fix Blurry Raster Maps

mtahasahin:
This PR fixes https://github.com/traccar/traccar-web/issues/827 by limiting zoom levels for raster maps to integers. 

How it works:
- If the current map is a raster map, set wheelZoomRate to 1. This way every scroll wheel tick increments or decrements zoom level by 1. 
- An event listener is added to the map which will listen for "idle" event(which is fired when the map is fully rendered after an action), then sets the zoom level to nearest integer value of current level. This handles zoom changes by mobile devices, touchpads etc.
- If the current map is not a raster map, set wheelZoomRate to default value, remove the event listener.

### DIFF
--- a/modern/src/map/Map.js
+++ b/modern/src/map/Map.js
@@ -42,18 +42,7 @@ const onIdle = function () {
   }
 };
 
-const setAllowedZoomLevels = map => {
-  if (Object.values(map.getStyle().sources).some(e => e.type === 'raster')) {
-    map.scrollZoom.setWheelZoomRate(1);
-    map.on('idle', onIdle);
-  } else {
-    map.scrollZoom.setWheelZoomRate(1/450);
-    map.off('idle', onIdle);
-  }
-};
-
 const initMap = async () => {
-  setAllowedZoomLevels(map);
   const background = await loadImage('images/background.svg');
   await Promise.all(deviceCategories.map(async category => {
     if (!map.hasImage(category)) {
@@ -110,6 +99,17 @@ const Map = ({ children }) => {
       removeReadyListener(listener);
     };
   }, []);
+
+  useEffect(() => {
+    if (mapReady && Object.values(map.getStyle().sources).some(e => e.type === 'raster')) {
+        map.scrollZoom.setWheelZoomRate(1);
+        map.on('idle', onIdle);
+      }
+    return () => {
+      map.scrollZoom.setWheelZoomRate(1/450);
+      map.off('idle', onIdle);
+    }
+  }, [mapReady]);
 
   useLayoutEffect(() => {
     const currentEl = containerEl.current;
--- a/modern/src/map/Map.js
+++ b/modern/src/map/Map.js
@@ -42,10 +42,8 @@ const onIdle = function () {
   }
 };
 
-const isRasterMap = map => Object.values(map.getStyle().sources).some(e => e.type === 'raster');
-
 const setAllowedZoomLevels = map => {
-  if(isRasterMap(map)) {
+  if (Object.values(map.getStyle().sources).some(e => e.type === 'raster')) {
     map.scrollZoom.setWheelZoomRate(1);
     map.on('idle', onIdle);
   } else {
--- a/modern/src/map/Map.js
+++ b/modern/src/map/Map.js
@@ -35,7 +35,27 @@ const updateReadyValue = value => {
   readyListeners.forEach(listener => listener(value));
 };
 
+const onIdle = function () {
+  const zoom = Math.round(map.getZoom())
+  if (zoom !== map.getZoom()) {
+    map.zoomTo(zoom)
+  }
+};
+
+const isRasterMap = map => Object.values(map.getStyle().sources).some(e => e.type === 'raster');
+
+const setAllowedZoomLevels = map => {
+  if(isRasterMap(map)) {
+    map.scrollZoom.setWheelZoomRate(1);
+    map.on('idle', onIdle);
+  } else {
+    map.scrollZoom.setWheelZoomRate(1/450);
+    map.off('idle', onIdle);
+  }
+};
+
 const initMap = async () => {
+  setAllowedZoomLevels(map);
   const background = await loadImage('images/background.svg');
   await Promise.all(deviceCategories.map(async category => {
     if (!map.hasImage(category)) {
--- a/modern/src/map/Map.js
+++ b/modern/src/map/Map.js
@@ -35,13 +35,6 @@ const updateReadyValue = value => {
   readyListeners.forEach(listener => listener(value));
 };
 
-const onIdle = function () {
-  const zoom = Math.round(map.getZoom())
-  if (zoom !== map.getZoom()) {
-    map.zoomTo(zoom)
-  }
-};
-
 const initMap = async () => {
   const background = await loadImage('images/background.svg');
   await Promise.all(deviceCategories.map(async category => {
@@ -102,12 +95,18 @@ const Map = ({ children }) => {
 
   useEffect(() => {
     if (mapReady && Object.values(map.getStyle().sources).some(e => e.type === 'raster')) {
-        map.scrollZoom.setWheelZoomRate(1);
-        map.on('idle', onIdle);
+      const onIdle = function () {
+        const zoom = Math.round(map.getZoom());
+        if (zoom !== map.getZoom()) {
+          map.zoomTo(zoom);
+        }
+      };
+      map.scrollZoom.setWheelZoomRate(1);
+      map.on('idle', onIdle);
+      return () => {
+        map.scrollZoom.setWheelZoomRate(1/450);
+        map.off('idle', onIdle);
       }
-    return () => {
-      map.scrollZoom.setWheelZoomRate(1/450);
-      map.off('idle', onIdle);
     }
   }, [mapReady]);
 


modern/src/map/Map.js
```diff
@@ -35,7 +35,27 @@ const updateReadyValue = value => {
...
 };
 
+const onIdle = function () {
+  const zoom = Math.round(map.getZoom())
+  if (zoom !== map.getZoom()) {
+    map.zoomTo(zoom)
+  }
+};
+
+const isRasterMap = map => Object.values(map.getStyle().sources).some(e => e.type === 'raster');
```
tananaev, line 45: Can we inline this?

modern/src/map/Map.js
```diff
@@ -35,7 +35,27 @@ const updateReadyValue = value => {
   readyListeners.forEach(listener => listener(value));
 };
 
+const onIdle = function () {
```
tananaev, line 38: I would move it to `setAllowedZoomLevels`. It's really only relevant there.

modern/src/map/Map.js
```diff
@@ -35,7 +35,27 @@ const updateReadyValue = value => {
...
+    map.scrollZoom.setWheelZoomRate(1);
+    map.on('idle', onIdle);
+  } else {
+    map.scrollZoom.setWheelZoomRate(1/450);
+    map.off('idle', onIdle);
+  }
+};
+
 const initMap = async () => {
+  setAllowedZoomLevels(map);
```
tananaev, line 58: How do we remove the listener?

modern/src/map/Map.js
```diff
@@ -35,7 +35,27 @@ const updateReadyValue = value => {
...
+  const zoom = Math.round(map.getZoom())
+  if (zoom !== map.getZoom()) {
+    map.zoomTo(zoom)
+  }
+};
+
+const isRasterMap = map => Object.values(map.getStyle().sources).some(e => e.type === 'raster');
+
+const setAllowedZoomLevels = map => {
+  if(isRasterMap(map)) {
```
tananaev, line 48:
```suggestion
  if (isRasterMap(map)) {
```

modern/src/map/Map.js
mtahasahin: line 53 `map.off('idle', onIdle);` removes the listener.

modern/src/map/Map.js
mtahasahin: I would, but it is an event listener function and when I want to remove the event listener I need to provide the same function to `map.off()` function. If I move it to `setAllowedZoomLevels`, a new function is created every time and `map.off()` can't remove the event listener.

mtahasahin:
Are there any other changes that need to be made?

tananaev:
Sorry, forgot to submit my comment.

modern/src/map/Map.js
mtahasahin: When the map is first loaded or a map style change is made using the switcher, `setAllowedZoomLevels` gets called. It checks whether the map is using raster tiles, if it is, it adds an event listener. If it is not, it removes the event listener. Adding the same function for the same event listener multiple times does not result in duplicate event listeners. Or trying to remove an event listener that is not binded do not cause any error. I didn't understand why we need to remove the event listener when map is not displayed because the map itself handles firing of those listener functions, when the map is gone so are all the event listener functions binded to it.

tananaev: Map is never gone. It's only hidden. So, the right thing to do is to remove it. That's why we use react components for anything that's attached to the map.

mtahasahin: Why? We want the event listener function to be there when we want to see the map again. What kind of a problem can it cause?

tananaev: I can't think of any specific problem, but I also don't see any good reason to not do it the right way.

mtahasahin: OK. What about this: in `Map` component I listen for changes on `mapReady` state using `useEffect` hook. If the map is ready and uses raster tiles add the event listener. Return a function that will remove the event listener as a cleanup function. Sounds good?

tananaev: Not sure I understand. Can you please update PR?

mtahasahin: done.

modern/src/map/Map.js
```diff
@@ -93,6 +100,17 @@ const Map = ({ children }) => {
...
 
+  useEffect(() => {
+    if (mapReady && Object.values(map.getStyle().sources).some(e => e.type === 'raster')) {
+        map.scrollZoom.setWheelZoomRate(1);
+        map.on('idle', onIdle);
+      }
+    return () => {
+      map.scrollZoom.setWheelZoomRate(1/450);
+      map.off('idle', onIdle);
+    }
```
tananaev, line 111: Can we do something like this?

```suggestion
    if (mapReady && Object.values(map.getStyle().sources).some(e => e.type === 'raster')) {
        map.scrollZoom.setWheelZoomRate(1);
        const onIdle = function () { ... };
        map.on('idle', onIdle);
        return () => {
          map.off('idle', onIdle);
        }
      } else {
         map.scrollZoom.setWheelZoomRate(1/450);
      }
```

modern/src/map/Map.js
```diff
@@ -35,6 +35,13 @@ const updateReadyValue = value => {
...
 };
 
+const onIdle = function () {
+  const zoom = Math.round(map.getZoom())
+  if (zoom !== map.getZoom()) {
+    map.zoomTo(zoom)
+  }
+};
+
 const initMap = async () => {
```
tananaev, line 45: Also, what happens when we change map layer?

tananaev (commented):
What ha

modern/src/map/Map.js
mtahasahin: Yes but shouldn't `onIdle` event listener and `setWheelZoomRate` functions go together?

modern/src/map/Map.js
mtahasahin: map layer changes -> mapReady becomes false -> clean up function triggers and event listener gets removed -> other map layer loads -> mapReady becomes true -> if it's a raster map event listener gets added again otherwise no action

modern/src/map/Map.js
tananaev: Don't we only need it for fixed zoom rate?

mtahasahin: We use `onIdle` event listener and `setWheelZoomRate` to display map on integer zoom levels. (One for mouse wheel, one for other devices.) When we revert one back to original we should revert the other one too.

tananaev: Not sure I understand this. One is a callback. The other one is configuration.

mtahasahin: - Let's say map is loaded and it's zoom level is 0. When user scroll down every wheel tick will change the zoom level by 1. This is done by `setWheelZoomRate(1)` configuration. 
- Let's say this time user uses his touchpad to zoom in or out. Now, `setWheelZoomRate` is useless and user can zoom to an arbitrary zoom level. To prevent this, we add an event listener for "idle" event which fires when the map is rendered after an action. This listener checks the current zoom level, if it is not an integer forces map to zoom to the nearest integer. This way whenever a non-integer zoom level is selected, it will automatically zooms to nearest integer.

tananaev: I see. I think my solution should still work, but we can rewrite it to something like this:
```
if (mapReady && Object.values(map.getStyle().sources).some(e => e.type === 'raster')) {
        const onIdle = function () { ... };
        map.scrollZoom.setWheelZoomRate(1);
        map.on('idle', onIdle);
        return () => {
          map.scrollZoom.setWheelZoomRate(1/450);
          map.off('idle', onIdle);
        }
      }
```

mtahasahin: OK. Will update the PR now.

modern/src/map/Map.js
```diff
@@ -93,6 +93,23 @@ const Map = ({ children }) => {
...
+      const onIdle = function () {
+        const zoom = Math.round(map.getZoom());
+        if (zoom !== map.getZoom()) {
+          map.zoomTo(zoom);
+        }
+      };
+      map.scrollZoom.setWheelZoomRate(1);
+      map.on('idle', onIdle);
+      return () => {
+        map.scrollZoom.setWheelZoomRate(1/450);
```
tananaev, line 107: Can we save the original zoom instead of hardcoding this value?

modern/src/map/Map.js
```diff
@@ -93,6 +93,23 @@ const Map = ({ children }) => {
     };
   }, []);
 
+  useEffect(() => {
+    if (mapReady && Object.values(map.getStyle().sources).some(e => e.type === 'raster')) {
+      const onIdle = function () {
+        const zoom = Math.round(map.getZoom());
+        if (zoom !== map.getZoom()) {
+          map.zoomTo(zoom);
+        }
```
tananaev, line 102: Can this just be
```suggestion
        map.zoomTo(Math.round(map.getZoom()));
```

tananaev (commented):
Sorry, a couple of other minor comments.

modern/src/map/Map.js
mtahasahin: this causes infinite loop since "idle" event will be fired after every zoom.

modern/src/map/Map.js
mtahasahin: This is not the zoom value but the rate how fast mouse wheel will zoom. It is not something that users can change.

tananaev: I understand, but where did you get the value from? What if developers change it in the future?

mtahasahin: `scrollZoom` class has a private property called `_wheelZoomRate`. I can use that to access the original value but what if developers change that?

tananaev: Interesting. No getter? I guess let's leave it as is.

tananaev:
Merged, thank you.

tananaev:
I have tried the feature and controls are very weird. Mouse wheel still lets me to zoom by small iterations, but also jumps quickly if scroll a bit faster. It doesn't seem like a good user experience. Is there any better solution? If not, I think it's better to have blurry maps with smooth zooming experience.

mtahasahin:
Just to be sure, you didn't like that when you scroll fast, the map zooms in/out fast, right? Because other than that, it should behave exactly as the map in [openstreetmap.org](https://openstreetmap.org). Included a sample video. If there are any other issues, it might be browser-specific. I can check it if you tell me what browser you are using.

https://user-images.githubusercontent.com/61321144/113567014-71586f00-9616-11eb-8da9-9c51c8a7feb1.mp4

tananaev:
I am using latest chrome browser. I can slowly move mouse wheel and it zooms smoothly and when I stop it animates to nearest round zoom level.

mtahasahin:
Couldn't reproduce with 3 Windows machines with latest Chrome. Are you using Mac?

tananaev:
Yes, using Mac.

mtahasahin:
I looked into this and it turns out that the mouse wheel event differs greatly from browser/OS to browser/OS. It causes different zoom speeds even with default configuration. See (https://github.com/mapbox/mapbox-gl-js/issues/7572). For now, let's revert the PR. If I find a browser/OS agnostic fix or Mapbox starts supporting it natively I'll open a new PR.

tananaev:
Reverted it for now. Feel free to send a new PR once there's a better solution.

PR #832, opened 2021-04-06
Data grid implementations for remaining reports

mail2bishnoi:
Hi Anton,

I have implemented Data-Grid for the remaining reports. Please review and suggest improvements.

For some of the reports, I have used 'getRowId' props for DataGrid because we have to provide a unique id for each row in the data set. 
https://github.com/mui-org/material-ui-x/issues/378

Thanks with regards.

### DIFF
--- a/modern/src/reports/EventReportPage.js
+++ b/modern/src/reports/EventReportPage.js
@@ -1,8 +1,8 @@
 import React, { useState } from 'react';
-import { TableContainer, Table, TableRow, TableCell, TableHead, TableBody, Paper } from '@material-ui/core';
+import { DataGrid } from '@material-ui/data-grid';
 import { FormControl, InputLabel, Select, MenuItem } from '@material-ui/core';
 import t from '../common/localization';
-import { formatPosition } from '../common/formatter';
+import { formatDate } from '../common/formatter';
 import ReportFilter from './ReportFilter';
 import ReportLayoutPage from './ReportLayoutPage';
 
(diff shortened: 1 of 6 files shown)


modern/src/reports/TripReportPage.js
```diff
@@ -30,40 +30,93 @@ const TripReportPage = () => {
...
-            ))}
-          </TableBody>
-        </Table>
-      </TableContainer>
+      <DataGrid
+        rows={items} 
+        columns={columns} 
+        hideFooter 
+        autoHeight
+        getRowId={() => Math.random()} />
```
tananaev, line 119: Why not just use index for id?

modern/src/reports/EventReportPage.js
```diff
@@ -59,30 +59,36 @@ const EventReportPage = () => {
...
-                <TableCell>{formatPosition(item, 'eventTime')}</TableCell>
-                <TableCell>{item.type}</TableCell>
-                <TableCell>{}</TableCell>
-                <TableCell>{}</TableCell>
-              </TableRow>
-            ))}
-          </TableBody>
-        </Table>
-      </TableContainer>
+      <DataGrid
```
tananaev, line 87: Please add paper to the background. See route report. Same for other reports.

mail2bishnoi: Done.

modern/src/reports/TripReportPage.js
mail2bishnoi: Sorry, I didn't understand which index. 

I think we can pass row object to getRowId callback but row doesn't have index.

tananaev: We cannot get item index in the list?

mail2bishnoi: Tried witout success.

modern/src/reports/EventReportPage.js
```diff
@@ -59,30 +59,38 @@ const EventReportPage = () => {
...
-                <TableCell>{formatPosition(item, 'eventTime')}</TableCell>
-                <TableCell>{item.type}</TableCell>
-                <TableCell>{}</TableCell>
-                <TableCell>{}</TableCell>
-              </TableRow>
-            ))}
-          </TableBody>
-        </Table>
-      </TableContainer>
+      <Paper>
```
tananaev, line 87: Can we move paper part to the `ReportLayoutPage` instead of duplicating it?

mail2bishnoi: Yes. Done

tananaev:
Merged, thanks.

PR #835, opened 2021-04-11
Update GeofenceMap.js

Aliber009:
when making api calls with post or put , changes are not diplayed in real time unless you reload the page

### DIFF
--- a/modern/src/map/GeofenceMap.js
+++ b/modern/src/map/GeofenceMap.js
@@ -14,7 +14,7 @@ const GeofenceMap = () => {
     if (response.ok) {
       setGeofences(await response.json());
     }
-  }, []);
+  }, [geofences]); //this will render created or updated geofences when making api calls 
 
   useEffect(() => {
     map.addSource(id, {


modern/src/map/GeofenceMap.js
```diff
@@ -14,7 +14,7 @@ const GeofenceMap = () => {
     if (response.ok) {
       setGeofences(await response.json());
     }
-  }, []);
+  }, [geofences]); //this will render created or updated geofences when making api calls
```
tananaev, line 17: Why do we need to re-load all geofences?

Aliber009: exemple, if we post/put request a circle, it wont render until we reload the page .
this is the only way i figure out how to make it update without reloading

tananaev: So who would update `geofences` then to trigger this?

Aliber009:
Ah, you are right, even this is working ,but  maybe it should be triggered
when mapbox draw data is updated

Le dim. 11 avr. 2021 à 21:36, Anton Tananaev ***@***.***> a
écrit :

tananaev:
Closing as I think it's outdated.

PR #838, opened 2021-04-13
Update GeofenceEditMap.js

Aliber009:
I converted the class component in my wished last edit, to a function component, so the edits are easy to see now
I hope it works

### DIFF
--- a/modern/src/map/GeofenceEditMap.js
+++ b/modern/src/map/GeofenceEditMap.js
@@ -1,21 +1,103 @@
-import '@mapbox/mapbox-gl-draw/dist/mapbox-gl-draw.css'
+import * as React from 'react';
 import MapboxDraw from '@mapbox/mapbox-gl-draw';
-import { useEffect } from 'react';
-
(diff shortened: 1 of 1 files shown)


tananaev:
The description is not very clear to me. Can you please properly explain all the changes in this PR and why we need them.

Aliber009:
the first class, is to allow adding a button in the draw control (the button I added is the circle Mode) Second: I have implemented a circle custom mode in my draw modes, and trigger the button to change the mode.
Later I will add map.on(draw.create, .. draw.update and draw.delete to handle all draw updates then from the data we will create the layers in GeofenceMap.js
Please if you can test.

tananaev:
Honestly, the formatting is not up to standard. Please fix before we can have a discussion. Also, why do we need a circle?

Aliber009:
Circle is the most commun geofence, and it should be added at first 
easy to draw , and can trigger approximities of devices

tananaev:
Outdated.

PR #839, opened 2021-04-13
Event type Localization and geofence store creation

mail2bishnoi:
Hi Anton,

I have implemented 'event type' localization in the events reports.

Along with this, I have created the geofence store and formatted the geofenceId to display the geofence name.

I was trying to move the 'formatGeofence' method to some centralized location i.e. in the formatter.js file, but we can not use hooks outside the react component. Please suggest how can I do this?


modern/src/MainController.js
```diff
@@ -0,0 +1,22 @@
+import {  useDispatch, useSelector } from 'react-redux';
+import { connect } from 'react-redux';
+import { geofencesActions } from './store';
+import { useEffectAsync } from './reactHelper';
+
+
+const MainController = () => {
```
tananaev, line 7: Where do we load devices? Maybe we can load geofences there as well.

mail2bishnoi: Yes we can but my thought process was to keep the SocketController separate from rest of the things like Geofences, Groups, Maintenance etc. If you would like to keep all the fetches in  SocketController only then I'll move it there.

tananaev: Maybe we need a better name then. Something like caching controller?

mail2bishnoi: CachingController/ StoreController

mail2bishnoi:
Renamed MainController.

modern/src/reports/EventReportPage.js
```diff
@@ -57,8 +59,17 @@ const Filter = ({ setItems }) => {
 
 const EventReportPage = () => {
 
+  const geofences = useSelector(state => state.geofences.items);
   const [items, setItems] = useState([]);
 
+  const formatGeofence = value => {
+    if (value !== 0) {
```
tananaev, line 66:
```suggestion
    if (value > 0) {
```

modern/src/CachingController.js
```diff
@@ -0,0 +1,22 @@
+import {  useDispatch, useSelector } from 'react-redux';
+import { connect } from 'react-redux';
+import { geofencesActions } from './store';
+import { useEffectAsync } from './reactHelper';
+
+
```
tananaev, line 6:
```suggestion

```

modern/src/App.js
```diff
@@ -30,13 +30,16 @@ import MaintenancesPage from './settings/MaintenancesPage';
 import MaintenancePage from './settings/MaintenancePage';
 import StatisticsPage from './admin/StatisticsPage';
```
tananaev, line 32: Probably remove the space here for consistency.

modern/src/CachingController.js
```diff
@@ -0,0 +1,22 @@
...
+  useEffectAsync(async () => {
+    if (authenticated) {
+      const response = await fetch('/api/geofences');
+      if (response.ok) {
+        dispatch(geofencesActions.update(await response.json()));
+      }
+    }
+  }, [authenticated]);
+  return null;
```
tananaev, line 19:
```suggestion

  return null;
```

mail2bishnoi:
Formatting is done.

tananaev:
Merged, thanks.

PR #840, opened 2021-04-16
add admin permissions to userPage

Aliber009:
added a checkbox to change user to admin or admin to user permission in the userPage


tananaev:
Why are there so many changes in the code?

Aliber009:
ops sorry I pull in the wrong file

PR #842, opened 2021-04-17
Update UserPage.js for user Management

Aliber009:
Adding all permissions properties like in the classic app for user management utilities


tananaev:
Please fix formatting.

tananaev:
Outdated and still not fixed.

PR #843, opened 2021-04-18
l10n: Add word "address"

Valdnet:
Correct the sentence by adding the word "address" to the text string.


tananaev:
This is incorrect. The message is inviting user to check their inbox, not address.

PR #844, opened 2021-04-20
Created mostly used stores

mail2bishnoi:
Hi Anton,

I have created mostly used stores that can be used in subsequent PRs. Please review and suggest improvements.


tananaev (commented):
I don't think we need all those stores. I would prefer to keep global stores to a minimum.

mail2bishnoi:
Hi Anton,

I have filtered out some of them. I think below mentioned should be there-
1. groups
2. drivers
3. maintenance

about rest of them I'm not sure. Please tell which stores I should keep and which should be removed.

tananaev:
Why do we need maintenance there?

mail2bishnoi:
In the Events report we have Maintenance Id column. To format that we need Maintenance.

tananaev:
OK, let's cut down to those 3 for now.

mail2bishnoi:
Done.

tananaev:
Merged, thanks.

PR #855, opened 2021-05-24
Fix issue with event time not being displayed in event report

Xenoamor:
Without this change the Time column is always empty in the events report view.
The change matches what is in ReportController.js


tananaev:
Merged, thank you.

PR #860, opened 2021-06-09
Add "Search in Map" button

alexthefifth:
Many other GPS tracking applications have this feature (i.e. GPSGate)

Uses OSM's Nominatim for the geocoder, see Usage Policy: https://operations.osmfoundation.org/policies/nominatim/

Uses the Popup and Geocoder OpenLayers libraries

demo:
animation

### DIFF
--- a/web/app/view/map/BaseMap.js
+++ b/web/app/view/map/BaseMap.js
@@ -30,7 +30,7 @@ Ext.define('Traccar.view.map.BaseMap', {
     },
 
     initMap: function () {
-        var server, layer, type, bingKey, lat, lon, zoom, maxZoom, target, poiLayer, self = this;
+        var server, layer, type, bingKey, lat, lon, zoom, maxZoom, target, poiLayer, geocoder, popup, self = this;
 
         server = Traccar.app.getServer();
 
@@ -186,6 +186,27 @@ Ext.define('Traccar.view.map.BaseMap', {
         }
 
         this.map.addControl(new ol.control.LayerSwitcher());
+
+        popup = new ol.Overlay.Popup();
+        this.map.addOverlay(popup);
+
+        geocoder = new Geocoder('nominatim', {
+            provider: 'bing',
+            key: server.get('bingKey'),
+            lang: 'en-US',
+            placeholder: 'Search for ...',
+            targetType: 'glass-button',
+            limit: 5,
+            autoComplete: true,
+            keepOpen: true
+        });
+        this.map.addControl(geocoder);
+
+        geocoder.on('addresschosen', function (evt) {
+            window.setTimeout(function () {
+                popup.show(evt.coordinate, evt.address.formatted);
+            }, 3000);
+        });
 
         target = this.map.getTarget();
         if (typeof target === 'string') {
--- a/web/app/view/map/BaseMap.js
+++ b/web/app/view/map/BaseMap.js
@@ -30,7 +30,7 @@ Ext.define('Traccar.view.map.BaseMap', {
     },
 
     initMap: function () {
-        var server, layer, type, bingKey, lat, lon, zoom, maxZoom, target, poiLayer, geocoder, popup, self = this;
+        var server, layer, type, bingKey, lat, lon, zoom, maxZoom, target, poiLayer, geocoder, self = this;
 
         server = Traccar.app.getServer();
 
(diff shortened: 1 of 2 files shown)


web/app/view/map/BaseMap.js
```diff
@@ -187,6 +187,27 @@ Ext.define('Traccar.view.map.BaseMap', {
 
         this.map.addControl(new ol.control.LayerSwitcher());
 
+        popup = new ol.Overlay.Popup();
+        this.map.addOverlay(popup);
+
+        geocoder = new Geocoder('nominatim', {
+            provider: 'bing',
+            key: server.get('bingKey'),
```
tananaev, line 195: This doesn't make sense. Why are you using bind key for nominatim?

web/app/view/map/BaseMap.js
```diff
@@ -187,6 +187,27 @@ Ext.define('Traccar.view.map.BaseMap', {
...
+            limit: 5,
+            autoComplete: true,
+            keepOpen: true
+        });
+        this.map.addControl(geocoder);
+
+        geocoder.on('addresschosen', function (evt) {
+            window.setTimeout(function () {
+                popup.show(evt.coordinate, evt.address.formatted);
+            }, 3000);
```
tananaev, line 208: Why is there a timeout?

web/load.js
```diff
@@ -178,6 +178,12 @@
     addStyleFile('https://cdn.traccar.com/js/ol/' + olVersion + '/ol.css');
     addScriptFile('https://cdn.traccar.com/js/ol/' + olVersion + '/ol.js');
 
+    addStyleFile('https://unpkg.com/ol-popup@4.0.0/src/ol-popup.css');
+    addScriptFile('https://unpkg.com/ol-popup@4.0.0');
+
+    addStyleFile('https://cdn.jsdelivr.net/npm/ol-geocoder@latest/dist/ol-geocoder.min.css');
+    addScriptFile('https://unpkg.com/ol-geocoder');
```
tananaev, line 185: We'll have to put this on our CDN.

web/app/view/map/BaseMap.js
alexthefifth: The timeout was taken from the sample code -- now removed

web/load.js
alexthefifth: This should be ok, both libraries are MIT licensed:

https://github.com/jonataswalker/ol-geocoder
https://github.com/walkermatt/ol-popup

web/app/view/map/BaseMap.js
alexthefifth: Changed to using OSM

web/load.js
tananaev: Please create a PR and update this PR with CDN links.

web/app/view/map/BaseMap.js
```diff
@@ -187,6 +187,24 @@ Ext.define('Traccar.view.map.BaseMap', {
 
         this.map.addControl(new ol.control.LayerSwitcher());
 
+        geocoder = new Geocoder('nominatim', {
+            provider: 'osm',
+            placeholder: 'Search for ...',
```
tananaev, line 192: Looks like hardcoded string. You need to use localized string here.

web/app/view/map/BaseMap.js
```diff
@@ -187,6 +187,24 @@ Ext.define('Traccar.view.map.BaseMap', {
...
+            targetType: 'glass-button',
+            featureStyle: new ol.style.Style(),
+            limit: 5,
+            keepOpen: true
```
tananaev, line 196: Are all these defaults? Do we have to provide them?

web/app/view/map/BaseMap.js
```diff
@@ -187,6 +187,24 @@ Ext.define('Traccar.view.map.BaseMap', {
...
+        });
+        this.map.addControl(geocoder);
+
+        geocoder.on('addresschosen', function (evt) {
```
tananaev, line 200: Please use full words for variables. `event` in this specific case.

web/app/view/map/BaseMap.js
```diff
@@ -187,6 +187,24 @@ Ext.define('Traccar.view.map.BaseMap', {
...
+            var popup = new ol.Overlay.Popup({
+                insertFirst: false
```
tananaev, line 202: What does this do?

tananaev:
We already have search in the modern app.

PR #865, opened 2021-06-23
Update craco.config.js

Aliber009:
no .htm file is generated when building , the standard craco config worked out


tananaev:
Not sure what you mean by it's not generated? HTML file is static. It doesn't need to be generated. And I get it in the output.

tananaev:
Actually correction. Looks like it is generated, but I still get it in the output without any modifications.

Aliber009:
it didnt work with npm run build , also the prod needs to be configured with .babel

tananaev:
Works fine for me by running `npm run build_release`>

PR #870, opened 2021-06-28
Change language: enables the user to change language on the modern front-end

rhh-ft:
Enables the user to change user interface language on the modern front-end just like on the old one


modern/src/LanguageContext.js
```diff
@@ -0,0 +1,6 @@
+import React from 'react';
+
+export const LanguageContext = React.createContext({
```
tananaev, line 3: Can we import `createContext` individually? Something like this:
```
import { createContext } from 'react';
```

modern/src/LanguageProvider.js
```diff
@@ -0,0 +1,18 @@
+import {LanguageContext} from "./LanguageContext";
+import React, {useState} from "react";
+import {defaultLanguage, setSelectedLanguage} from "./common/localization";
+
+const LanguageProvider = (props) => {
```
tananaev, line 5: I think we should combine `LanguageContext` and `LanguageProvider` into a single file.

It would also be nice to migrate existing "localization" code here, so we have a single place for all language related stuff.

modern/src/LanguageProvider.js
```diff
@@ -0,0 +1,18 @@
...
+    const [language, setLanguage] = useState(defaultLanguage);
+
+    const handleLanguageChange = (nextLanguage) => {
+        setSelectedLanguage(nextLanguage);
+        setLanguage(nextLanguage);
+    };
```
tananaev, line 11: Once we do the above, we probably won't need this anymore.

modern/src/LanguageProvider.js
```diff
@@ -0,0 +1,18 @@
...
+
+    return <LanguageContext.Provider value={{language, setLanguage: handleLanguageChange}}>
+        {props.children}
+    </LanguageContext.Provider>
```
tananaev, line 15:
```suggestion
    return (
      <LanguageContext.Provider value={{language, setLanguage: handleLanguageChange}}>
        {props.children}
      </LanguageContext.Provider>
    )
```

tananaev: btw please use 2-spaces indentation everywhere. Make sure the style is consistent with the rest of the code.

modern/src/components/registration/LoginForm.js
```diff
@@ -57,6 +60,11 @@ const LoginForm = () => {
     }
   }
 
+  const handleLanguageChange = e => {
+    setLanguage(e.target.value);
+    languageContext.setLanguage(e.target.value);
```
tananaev, line 65: We should be able to use single source of truth.

tananaev (commented):
Overall looks good. Added some minor comments.

modern/src/LanguageProvider.js
rhh-ft: LanguageContext and LanguageProvider have been combined into a single file.

Are you thinking about migrating the existing code from the "localization.js" file into LanguageProvider and use react context for all localization related stuff?

modern/src/components/registration/LoginForm.js
rhh-ft: Now it only uses the context

modern/src/LanguageProvider.js
rhh-ft: I don't understand this comment.
Could you please explain how the code can be simplified?

modern/src/LanguageProvider.js
tananaev: Yes, I don't want to have two different places to manage localization. I think all the code from `localization.js` should be integrated into `LanguageProvider`. And I mean not just copy, but integrate into the component.

modern/src/LanguageProvider.js
tananaev: Once you move the code from `localization.js`, you won't need this method. You would just call `setLanguage`.

tananaev (commented):
Answered comments.

tananaev:
@rhh-ft any updates?

tananaev:
@rhh-ft should we take over this PR? Is there any plan to address comments from your end?

rhh-ft:
@tananaev Hi sorry for the late response I have been on holiday. It would be better if you guys would take over at this point

oyhan:
Hi @tananaev @rhh-ft 
I've played this PR.
As an end user is it not better to keep language preference (including rtl/ltr) after the page reloads?

tananaev:
It is better. I don't think anyone is arguing otherwise.

Issue #869, opened 2021-06-28
Fix issues with inputs


dkyeremeh:
Fix issues with inputs

dkyeremeh:
@tananaev @mail2bishnoi... Working on this

PR #871, opened 2021-06-29
Inputs fix

dkyeremeh:
## Summary
1. Addresses issues with inputs #869
2. Updated styling of disabled buttons
3. Updated padding of grids containing inputs to accommodate new input margin

@tananaev Kindly let me know if I've missed any part the inputs were not showing correctly

Screenshot 2021-06-29 at 6 20 38 AM
Screenshot 2021-06-29 at 6 21 06 AM

### DIFF
--- a/modern/src/components/registration/LoginForm.js
+++ b/modern/src/components/registration/LoginForm.js
@@ -59,7 +59,7 @@ const LoginForm = () => {
 
   return (
     <StartPage>
-      <Grid container direction='column' spacing={3}>
+      <Grid container direction='column' spacing={2}>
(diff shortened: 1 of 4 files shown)


tananaev (commented):
One tiny thing I noticed is that label text start is not aligned with error message start. It's not an issue with default component because of the field background, but it looks a bit odd with our design. Can we fix it?

dkyeremeh:
Nice catch. Will address it

dkyeremeh:
@tananaev done

tananaev:
Merged, thank you.

Issue #878, opened 2021-07-22
Home screen issues


tananaev:
1. Top and bottom padding cuts list items.

<img width="671" alt="Screen Shot 2021-07-21 at 11 29 37 PM" src="https://user-images.githubusercontent.com/1161863/126599065-ebacacfd-4a80-499a-b4ad-394c3b90545e.png">

2. Large virtual list doesn't work well with map background.

https://user-images.githubusercontent.com/1161863/126598976-a513ce8c-d298-4fcd-ba48-2b15c1c6142e.mp4

^ download video file if it doesn't play directly in GitHub

For the second issue should we just have white background? @dkyeremeh what do you think?

There's also a problem that map doesn't center properly. It should center in the visible only, which means that we need to apply map padding.

cc @mail2bishnoi

dkyeremeh:
I'll create a PR for issue 1.

I was expecting issue 2 to come up. We'll be using a SectionList or infinite scroll to increase the performance.

dkyeremeh:
@tananaev can we close this issue?

tananaev:
Yes, thanks for fixing it.

PR #885, opened 2021-07-23
Home screen fix

dkyeremeh:
Addresses #878 

- Fixed the top and bottom cutoff
- Changed the background to accommodate the virtual scroll

@tananaev I've attached a video of how it looks now

https://user-images.githubusercontent.com/52928432/126792449-45722c5d-b303-4a01-a2d7-01b383dafb97.mov

### DIFF
--- a/modern/src/DevicesList.js
+++ b/modern/src/DevicesList.js
@@ -151,7 +151,7 @@ const DeviceView = ({ updateTimestamp, onMenuClick }) => {
             height={height}
             itemCount={items.length}
             itemData={{ items, onMenuClick }}
-            itemSize={72 + 1}
+            itemSize={72}
             overscanCount={10}
             innerRef={listInnerEl}
           >
--- a/modern/src/DevicesList.js
+++ b/modern/src/DevicesList.js
@@ -18,7 +18,7 @@ import { devicesActions } from './store';
 import EditCollectionView from './EditCollectionView';
 import { useEffectAsync } from './reactHelper';
 import { formatPosition } from './common/formatter';
-import { getDevices, getItemPosition } from './common/selectors';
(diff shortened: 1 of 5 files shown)


modern/src/common/selectors.js
```diff
@@ -1,3 +1,7 @@
 export const getIsAdmin = (state) => state.session.user?.administrator;
 
 export const getUserId = (state) => state.session.user?.id;
+
+export const getDevices = (state) => Object.values(state.devices.items);
+
+export const getItemPosition = (itemId) => (state) => state.positions.items[itemId];
```
tananaev, line 7: Can we rename with to just `getPosition`? And we don't need to use `itemId` here. It's just position id.

modern/src/DevicesList.js
```diff
@@ -145,7 +151,9 @@ const DeviceView = ({ updateTimestamp, onMenuClick }) => {
             height={height}
             itemCount={items.length}
             itemData={{ items, onMenuClick }}
-            itemSize={72 + 1}
+            itemSize={72}
+            overscanCount={10}
+            innerRef={listInnerEl}
```
tananaev, line 156: Why do we need this parameter?

dkyeremeh: @tananaev It's to allow styling of the list container. We need space at the bottom and top of the list. Putting the space anywhere else result in the list item being cut. This is the problem you pointed out in point 1 of https://github.com/traccar/traccar-web/issues/878

modern/src/common/selectors.js
dkyeremeh: Noted

dkyeremeh:
@tananaev PR updated

tananaev:
Merged, thank you.

PR #888, opened 2021-08-04
Quick device menu

mail2bishnoi:
Hi @tananaev 

I have started working on the device quick menu PR. Please review the initial draft.

Can you please tell me how to pass the event(which we pass to onMarkerClick function), on clicking the list item to show the quick menu?

### DIFF
--- a/modern/public/styles.css
+++ b/modern/public/styles.css
@@ -10,3 +10,7 @@ canvas {
 .mapboxgl-ctrl {
   margin: 10px;
 }
+
+.maplibregl-popup {
+  max-width: 330px!important;
+}
--- a/modern/public/styles.css
+++ b/modern/public/styles.css
@@ -12,5 +12,5 @@ canvas {
 }
 
 .maplibregl-popup {
-  max-width: 330px!important;
+  max-width: 330px !important;
 }
--- a/modern/src/map/StatusView.js
+++ b/modern/src/map/StatusView.js
@@ -19,7 +19,6 @@ import {
 import { useAttributePreference } from '../common/preferences';
 import RemoveDialog from '../RemoveDialog';
 import t from '../common/localization';
-import { formatPosition } from '../common/formatter';
 import { getPosition } from '../common/selectors';
 
 const useStyles = makeStyles((theme) => ({
--- a/modern/src/map/StatusView.js
+++ b/modern/src/map/StatusView.js
@@ -25,15 +25,7 @@ const useStyles = makeStyles((theme) => ({
   paper: {
     width: '300px',
   },
-  green: {
-    color: theme.palette.common.green,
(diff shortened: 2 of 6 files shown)


modern/public/styles.css
```diff
@@ -10,3 +10,7 @@ canvas {
 .mapboxgl-ctrl {
   margin: 10px;
 }
+
+.maplibregl-popup {
```
tananaev, line 14: Does it not adjust automatically based on the content?

modern/src/map/StatusView.js
```diff
@@ -1,69 +1,141 @@
...
+  },
+  green: {
+    color: theme.palette.common.green,
+  },
+  red: {
+    color: theme.palette.common.red,
+  },
+  gray: {
+    color: theme.palette.common.gray,
+  },
```
tananaev, line 35: Should we extract this common part somewhere?

tananaev (commented):
Overall looks good, I think. Can you please also attach a screenshot.

> Can you please tell me how to pass the event(which we pass to onMarkerClick function), on clicking the list item to show the quick menu?

Maybe through props? Why do you need to pass the event?

tananaev:
Also don't forget to rebase. You have a merge conflict.

modern/public/styles.css
mail2bishnoi: No. There is an inline style with a max-width of 240px.

modern/src/map/StatusView.js
mail2bishnoi: Should I move these classes to global css file?

mail2bishnoi:
> 
> 
> Overall looks good, I think. Can you please also attach a screenshot.
> 
> > Can you please tell me how to pass the event(which we pass to onMarkerClick function), on clicking the list item to show the quick menu?
> 
> Maybe through props? Why do you need to pass the event?

To show the StatusView on clicking device in the list.

tananaev:
> To show the StatusView on clicking device in the list.

I see what you mean. I think the best option is to tell map that a device needs to selected. Then map can get coordinates to display popup.

modern/src/map/StatusView.js
tananaev: I think just a separate file. You can the include it in the object using deconstruction like `...colors`.

modern/public/styles.css
```diff
@@ -10,3 +10,7 @@ canvas {
...
+  max-width: 330px!important;
```
tananaev, line 15:
```suggestion
  max-width: 330px !important;
```

mail2bishnoi:
Hi @tananaev 

Please review now.

tananaev:
Merged, thank you.

PR #891, opened 2021-08-17
Replay Screen Initial Draft

mail2bishnoi:
Hi Anton,

This is an initial draft of Replay Screen. Please review.

Replay_desktop

Replay_Mobile


modern/src/reports/ReportFilter.js
```diff
@@ -2,17 +2,27 @@ import React, { useState } from 'react';
...
 } from '@material-ui/core';
-import { useSelector } from 'react-redux';
+import { useDispatch, useSelector } from 'react-redux';
 import moment from 'moment';
+import { devicesActions } from '../store';
 import t from '../common/localization';
 
-const ReportFilter = ({ children, handleSubmit, showOnly }) => {
+const ReportFilter = ({
+  children, handleSubmit, showOnly, replay,
```
tananaev, line 11: I don't think it's ideal to have feature specific parameter here. Maybe we can make the name more generic, like `fullScreen` or something like that.

modern/src/reports/ReportFilter.js
```diff
@@ -2,17 +2,27 @@ import React, { useState } from 'react';
...
+  const dispatch = useDispatch();
+
   const [deviceId, setDeviceId] = useState();
   const [period, setPeriod] = useState('today');
   const [from, setFrom] = useState(moment().subtract(1, 'hour'));
   const [to, setTo] = useState(moment());
 
+  const handleDeviceChange = (e) => {
+    setDeviceId(e.target.value);
+    dispatch(devicesActions.select({ id: e.target.value }));
```
tananaev, line 23: What is this used for?

modern/src/reports/ReplayPage.js
```diff
@@ -41,58 +62,136 @@ const TimeLabel = ({ children, open, value }) => (
...
+                  </Grid>
+                </Grid>
+              </Paper>
+            ) : (
+              <Paper elevation={3} className={classes.reportFilterContainer} square>
+                <ReportFilter handleSubmit={handleSubmit} replay showOnly>
+                  <Grid item xs={6}>
+                    <TextField
+                      fullWidth
+                      label="Playback per min"
```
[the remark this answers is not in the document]
mail2bishnoi: Localization done. I have no idea about it just implemented the design.

modern/src/reports/ReplayPage.js
```diff
@@ -41,58 +62,136 @@ const TimeLabel = ({ children, open, value }) => (
...
+                      value={playbackPerMin}
+                      onChange={(e) => setPlaybackPerMin(e.target.value)}
+                      variant="filled"
+                    />
+                  </Grid>
+                  <Grid item xs={6}>
+                    <FormControlLabel
+                      classes={{ root: classes.formControlLabel }}
+                      control={<Switch checked={autoPlay} onChange={(e) => setAutoPlay(e.target.checked)} name="autoPlay" color="primary" edge="start" />}
+                      label="AutoPlay"
```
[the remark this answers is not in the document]
mail2bishnoi: Done.

modern/src/reports/ReportFilter.js
mail2bishnoi: I too don't want to have this param but the layout of the Report Filter is completely different on Replay Screen.

modern/src/reports/ReportFilter.js
mail2bishnoi: To show the name of the selected device above the slider.

modern/src/reports/ReplayPage.js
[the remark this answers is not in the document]
tananaev: Well, you have to add new strings to en.json.

modern/src/reports/ReportFilter.js
tananaev: That's fine, but I want the variable name to be more generic because what if some other screen needs the same layout?

modern/src/reports/ReportFilter.js
tananaev: We need to find another way. You are reusing global state that's not designed for it. Also, this has to be more localized to this feature.

mail2bishnoi: Done.

mail2bishnoi:
Hi Anton,

Please review now.

Also need help from you and @dkyeremeh for implementing Autoplay and pause filter. How to implement that feature?

Thanks.

modern/src/reports/ReportFilter.js
tananaev: So, do we still need this code?

modern/src/theme/palette.js
```diff
@@ -29,4 +29,13 @@ export default {
       color: traccarGray,
     },
   },
+  typography: {
```
tananaev, line 32: Why is this in `palette` file?

modern/src/reports/ReportFilter.js
```diff
@@ -108,16 +118,32 @@ const ReportFilter = ({ children, handleSubmit, showOnly }) => {
...
+          <Button
+            onClick={() => handleClick(false, true)}
+            variant="outlined"
+            color="secondary"
+            fullWidth
+          >
+            {t('reportShow')}
+          </Button>
+        </Grid>
+        )}
```
tananaev, line 146: We need to remove duplication here.

modern/src/reports/ReportFilter.js
```diff
@@ -59,17 +69,17 @@ const ReportFilter = ({ children, handleSubmit, showOnly }) => {
...
-          <Select value={deviceId} onChange={(e) => setDeviceId(e.target.value)}>
+          <Select value={deviceId} onChange={handleDeviceChange}>
             {devices.map((device) => (
               <MenuItem value={device.id}>{device.name}</MenuItem>
             ))}
           </Select>
         </FormControl>
       </Grid>
-      <Grid item xs={12} sm={period === 'custom' ? 3 : 6}>
+      <Grid item xs={fullScreen ? 6 : 12} sm={!fullScreen && period === 'custom' ? 3 : 6}>
```
tananaev, line 82: Is there some way to check `fullScreen` on the parent instead instead of doing to every single item?

modern/src/reports/ReplayPage.js
```diff
@@ -41,58 +62,138 @@ const TimeLabel = ({ children, open, value }) => (
...
+                  </Grid>
+                </Grid>
+              </Paper>
+            ) : (
+              <Paper elevation={3} className={classes.reportFilterContainer} square>
+                <ReportFilter handleSubmit={handleSubmit} fullScreen showOnly>
+                  <Grid item xs={6}>
+                    <TextField
+                      fullWidth
+                      label={t('reportPlaybackPerMinute')}
```
tananaev, line 176: Is this available in the `en.json`? I don't see it in your PR.

Are there any other strings that are also missing?

modern/src/reports/ReplayPage.js
```diff
@@ -41,58 +62,138 @@ const TimeLabel = ({ children, open, value }) => (
...
+    if (selectedDeviceId) {
+      const device = state.devices.items[selectedDeviceId] || null;
+      if (device) {
+        return device.name;
+      }
+    }
+    return null;
+  });
+
+  const [playbackPerMin, setPlaybackPerMin] = useState('');
```
tananaev, line 81: Can we call this `playbackSpeed` or something?

modern/src/reports/ReplayPage.js
```diff
@@ -41,58 +62,138 @@ const TimeLabel = ({ children, open, value }) => (
...
+  const [autoPlay, setAutoPlay] = useState(false);
+  const [isPlaying, setIsPlaying] = useState(false);
```
tananaev, line 83: It looks like you replaced `expanded` with this, but I don't think it's conceptually correct. You can potentially have history playing and configuration expanded.

mail2bishnoi:
Hi Anton,

Please review now.

@dkyeremeh Can you please explain the logic for "Playback per Minute"?

Thanks

dkyeremeh:
@mail2bishnoi

The playback per minute is the number of instances that will be shown within a minute.

### Example
If a device has 500 position data today, clicking play will replay the positions one after the other (like an animation). 

If the playback speed is 120 per minute, then, 120 different positions will be shown in 1 minute. In other words, the user will see 2 every second.

The higher the playback speed (playback per minute), the more items will be shown every minute.

The user can press pause to stop the positions from automatically playing

tananaev:
I think we need some better logic because some devices might report once every few minute and others can report every second or so. We should probably calculate by time. For example, 1 hour -> 1 minute in replay mode.

dkyeremeh:
Unless we are assigning user-friendly names such as `fast`, `slow`, etc, I think the initial approach is still more user friendly. The user can adjust the playback speed to correspond with how fast they want the replay to be. This is useful whether the device reports frequently or less frequently

### Example 1: A device reports once every 2 minutes. 
- The user can set the playback speed to 60 per minute to show 2 hours worth of reported data in one minute. 
- They can also set it to 10 to show 20 mins of reported data in a minute. 
- Or even set it to 1 to show 2 mins of data every minute

### Example 2: A device reports every seconds. 
- The user can set the playback speed to 3600 per minute to show 1 hour worth of reported data in one minute. 
- Or even set it to 120 to show 2 mins of data every minute

Alternatively, we can use a logarithmic slider so users don't need bother themselves with numbers; or use keywords such as `very slow`, `slow`, `fast`, etch

tananaev:
@dkyeremeh what would happen if device reports at irregular intervals? It can be reporting by distance, for example. It would look very confusing in playback at regular intervals.

dkyeremeh:
@tananaev I understand your concern. The idea of the playback is to give a quick overview of the device's location over time. 

Using a playback rate that reflects time will reflect waits and stops; and also accommodate reporting at irregular  intervals. 

The downside is that the logic to compute that can be extremely tricky.
Using an event-loop such as `requestAnimationFrame` can be  performant. But even that can be very tricky to implement

tananaev:
OK, should we go ahead with some simple implementation then for now? I can explore something more complicated later.

dkyeremeh:
I agree with a simple solution for now. We can implement a time-based approach once we are able to come up with a good approach.

mail2bishnoi:
Hi Anton,

Please review and merge this pull request.

Can we implement PlaybackSpeed feature in a separate PR?

Thanks.

web/l10n/en.json
```diff
@@ -424,6 +424,8 @@
     "reportSpentFuel": "Spent Fuel",
     "reportStartOdometer": "Odometer Start",
     "reportEndOdometer": "Odometer End",
+	"reportPlaybackPerMinute": "Playback Per Minute",
+	"reportAutoPlay": "Auto Play",
```
tananaev, line 428: Please remove indentation. Also I think "Autoplay" can be a single word.

modern/src/theme/typography.js
```diff
@@ -0,0 +1,9 @@
+export default {
+  title: {
+    fontFamily: 'Roboto',
+    fontSize: 20,
+    fontWeight: 500,
+    lineHeight: 1.5,
+    color: '#777',
```
tananaev, line 7: Is it not some standard font? Why do we need a custom one?

modern/src/reports/ReportFilter.js
```diff
@@ -59,9 +61,23 @@ const ReportFilter = ({ children, handleSubmit, showOnly }) => {
...
+    <>
+      {fullScreen ? (
+        <Grid item xs={6} sm={6}>
+          {children}
+        </Grid>
+      ) : (
+        <Grid item xs={12} sm={period === 'custom' ? 3 : 6}>
+          {children}
+        </Grid>
+      )}
```
tananaev, line 74: Seems like the difference is only in attributes. Can we remove duplication?

modern/src/reports/ReplayPage.js
```diff
@@ -40,60 +62,149 @@ const TimeLabel = ({ children, open, value }) => (
...
   const [index, setIndex] = useState(0);
+  const [selectedDeviceId, setSelectedDeviceId] = useState();
+  const [playbackSpeed, setPlaybackSpeed] = useState('');
+  const [expanded, setExpanded] = useState(true);
+  const [autoPlay, setAutoPlay] = useState(false);
+  const [isPlaying, setIsPlaying] = useState(true);
+
+  const deviceName = useSelector((state) => {
+    if (selectedDeviceId) {
+      const device = state.devices.items[selectedDeviceId] || null;
```
tananaev, line 79: `|| null` is not needed.

modern/src/reports/ReplayPage.js
```diff
@@ -40,60 +62,149 @@ const TimeLabel = ({ children, open, value }) => (
...
+    setSelectedDeviceId(deviceId);
     const query = new URLSearchParams({ deviceId, from, to });
     const response = await fetch(`/api/positions?${query.toString()}`, { headers });
     if (response.ok) {
       setIndex(0);
       setPositions(await response.json());
       setExpanded(false);
+
+      if (autoPlay) {
+        setInterval(routePlayback, 500);
```
tananaev, line 102: How do we cancel playback when component is removed?

mail2bishnoi:
Hi Anton,

Please review now.

modern/src/debug.log
```diff
@@ -0,0 +1,2 @@
+[0908/162944.036:ERROR:registration_protocol_win.cc(103)] CreateFile: The system cannot find the file specified. (0x2)
```
tananaev, line 1: What is this?

web/l10n/en.json
```diff
@@ -424,6 +424,8 @@
     "reportSpentFuel": "Spent Fuel",
     "reportStartOdometer": "Odometer Start",
     "reportEndOdometer": "Odometer End",
+	"reportPlaybackPerMinute": "Playback Per Minute",
```
tananaev, line 427: Indentation is wrong.

tananaev (commented):
Can you please review your own PR first before asking us to review it to make sure it's cleaned up (e.g. no unrelated changes/files, formatting is right etc).

mail2bishnoi:
> 
> 
> Can you please review your own PR first before asking us to review it to make sure it's cleaned up (e.g. no unrelated changes/files, formatting is right etc).

Hi Anton,

I do review every time I push code to the server. Unfortunately, this time missed the unrelated log file generated "debug.log" and the Indentation in en.json file changed during the push.

Sorry for the inconvenience.

modern/src/reports/ReplayPage.js
```diff
@@ -40,14 +62,48 @@ const TimeLabel = ({ children, open, value }) => (
...
+    let timer;
+    if (autoPlay) {
+      timer = setInterval(routePlayback, 500);
+    }
+    return () => {
+      if (timer) {
+        clearInterval(timer);
+      }
+    };
+  }, [positions]);
```
tananaev, line 103:
```suggestion
  useEffect(() => {
    let timer;
    if (autoPlay) {
      const timer = setInterval(routePlayback, 500);
      return () => clearInterval(timer);
    };
  }, [positions]);
```

modern/src/reports/ReplayPage.js
```diff
@@ -40,14 +62,48 @@ const TimeLabel = ({ children, open, value }) => (
...
 
   const handleSubmit = async (deviceId, from, to, _, headers) => {
+    setSelectedDeviceId(deviceId);
```
tananaev, line 106: Why select device here instead of when you actually select it?

modern/src/reports/ReplayPage.js
```diff
@@ -59,41 +115,105 @@ const ReplayPage = () => {
...
+                      name="playbackpermin"
+                      value={playbackSpeed}
+                      onChange={(e) => setPlaybackSpeed(e.target.value)}
+                      variant="filled"
+                    />
+                  </Grid>
+                  <Grid item xs={6}>
+                    <FormControlLabel
+                      classes={{ root: classes.formControlLabel }}
+                      control={<Switch checked={autoPlay} onChange={(e) => setAutoPlay(e.target.checked)} name="autoPlay" color="primary" edge="start" />}
```
tananaev, line 206: Can you please split parameters one per line for more readability.

tananaev (commented):
Added a couple of minor comments. Can you also please include screenshot for both true and false values of `expanded`.

tananaev:
Closing as the code is outdated.

PR #900, opened 2021-09-13
Updated Spanish l10n

castillo92:
Changes in spelling and language, suspensory points, unnecessary capitalization, accents, etc.


tananaev:
Please use Transifex for localization:

https://www.transifex.com/traccar/traccar

castillo92:
OK Anton, I uploaded again all translation. I hope you could include in next release.

I only have a doubt with two important buttons in the first screen... "Login" button... Because in spanish location is "Iniciar Sesión" but have two words and I do not know if it will fit in the button.

The same for "Reset Pasword" because the translation is "Reiniciar contraseña" so... longer...

tananaev:
Buttons should automatically scale up to fit the text.

PR #908, opened 2021-12-03
Added device attribute allowForwarding

mivtdole:
With this change a user can select per device if the data should be forwarded or not.
In our situation we have a lot of trackers but only the data for a few needs to be forwarded.

### DIFF
--- a/web/l10n/en.json
+++ b/web/l10n/en.json
@@ -90,6 +90,7 @@
     "attributeDevicePassword": "Device Password",
     "attributeDeviceInactivityStart": "Device Inactivity Start",
     "attributeDeviceInactivityPeriod": "Device Inactivity Period",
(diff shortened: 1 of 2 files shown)


web/l10n/en.json
```diff
@@ -90,6 +90,7 @@
     "attributeDevicePassword": "Device Password",
     "attributeDeviceInactivityStart": "Device Inactivity Start",
     "attributeDeviceInactivityPeriod": "Device Inactivity Period",
+    "attributeDeviceAllowForward": "Allow Forward",
```
tananaev, line 93: This should probably be "Allow Forwarding" or something like this.

mivtdole: Renamed to "Allow Forwarding"

PR #912, opened 2022-01-21
Update fa.json l10n file

mrbayati2002:
Translate some phrases to Persian(fa).


tananaev:
For localization please use transfer:

mrbayati2002:
Thanks.

PR #916, opened 2022-02-08
Update devices list search

lucasleite01:
Implemented search for devices list:

https://user-images.githubusercontent.com/17521730/153055060-a707fdc9-d12b-4a67-9f26-6d53575a63bf.mov

### DIFF
--- a/modern/src/DevicesList.js
+++ b/modern/src/DevicesList.js
@@ -135,7 +135,7 @@ const DeviceView = ({ updateTimestamp, onMenuClick }) => {
 
   const filteredItems = useSelector(getFilteredDevices);
 
-  const getItems = () => filteredItems.length == 0 ? items : filteredItems;
+  const getItems = () => (filteredItems.length === 0 ? items : filteredItems);
 
   if (listInnerEl.current) {
     listInnerEl.current.className = classes.listInner;
--- a/modern/src/DevicesList.js
+++ b/modern/src/DevicesList.js
@@ -18,7 +18,7 @@ import { devicesActions } from './store';
 import EditCollectionView from './EditCollectionView';
 import { useEffectAsync } from './reactHelper';
 import { formatPosition } from './common/formatter';
-import { getDevices, getPosition } from './common/selectors';
+import { getDevices, getFilteredDevices, getPosition } from './common/selectors';
 import { useTranslation } from './LocalizationProvider';
 
 const useStyles = makeStyles((theme) => ({
@@ -133,6 +133,10 @@ const DeviceView = ({ updateTimestamp, onMenuClick }) => {
 
   const items = useSelector(getDevices);
 
+  const filteredItems = useSelector(getFilteredDevices);
+
+  const getItems = () => filteredItems.length == 0 ? items : filteredItems;
+
   if (listInnerEl.current) {
     listInnerEl.current.className = classes.listInner;
   }
@@ -151,8 +155,8 @@ const DeviceView = ({ updateTimestamp, onMenuClick }) => {
           <FixedSizeList
             width={width}
             height={height}
-            itemCount={items.length}
-            itemData={{ items, onMenuClick }}
+            itemCount={getItems().length}
+            itemData={{ items: getItems(), onMenuClick }}
             itemSize={72}
             overscanCount={10}
             innerRef={listInnerEl}
--- a/modern/src/DevicesList.js
+++ b/modern/src/DevicesList.js
@@ -1,4 +1,4 @@
-import React, { useRef } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { makeStyles } from '@material-ui/core/styles';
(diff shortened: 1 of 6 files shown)


modern/src/store/devices.js
```diff
@@ -20,6 +21,14 @@ const { reducer, actions } = createSlice({
     remove(state, action) {
       delete state.items[action.payload];
     },
+    filterByName(state, action) {
```
tananaev, line 24: Can filtering be done in the component instead of global?

modern/src/MainPage.js
```diff
@@ -105,6 +108,8 @@ const MainPage = () => {
     setCollapsed(!collapsed);
   };
 
+  useEffect(() => dispatch(devicesActions.filterByName(deviceName.trim())), [deviceName]);
```
tananaev, line 111: It would be nice to filter by other parameters as well. At least device id.

modern/src/store/devices.js
```diff
@@ -20,6 +21,14 @@ const { reducer, actions } = createSlice({
...
+      return {
+        ...state,
+        filteredItems: action.payload
+          ? Object.values(state.items).filter((item) => item.name?.toLowerCase().includes(action.payload?.toLowerCase()))
```
tananaev, line 28: What happens to this if device list is updated? How is this updated?

modern/src/store/devices.js
lucasleite01: I think that could be done that way, but it will require to pass params down to components and to store a filtered list of devices in the component state or to use Context API.

Which one do you think is best?

modern/src/MainPage.js
lucasleite01: Do you mean using the same input text?
image

Since it has only one text input, a way to do that would be comparing also de the device id. Something like this:
```
filterByNameOrDeviceId(state, action) {
      return {
        ...state,
        filteredItems: action.payload
          ? Object.values(state.items).filter((item) => item.name?.toLowerCase().includes(action.payload?.toLowerCase()) || item.deviceId === action.payload)
          : state.items,
      };
    },
```
It's OK?

tananaev: Yeah, search in both. Don't do equals.

modern/src/store/devices.js
lucasleite01: Every time the user changes the input text it uses the current device list to filter. That way, it will return the most updated filtered list possible.

However, if the user types "John" and stops for a few seconds and in that time new information about devices arrive, the list displayed will be outdated.

I don't think it worth the effort to make also this filtered list of devices updated at every new information arrived since it will be outdated generally for a few seconds, but it could be done if you think differently. Please, let me know what you think.

Note: if we change the filtering to the component state or to a context, this problem probably will disappear, since we will use the most updated `state.items` every time we filter.

modern/src/store/devices.js
lucasleite01: Ok, I'll work on it.

modern/src/store/devices.js
tananaev: If you type something in the field and devices don't get updated anymore, it's a big issue.

modern/src/store/devices.js
```diff
@@ -5,6 +5,7 @@ const { reducer, actions } = createSlice({
   initialState: {
     items: {},
     selectedId: null,
+    filteredItems: {},
```
tananaev, line 8: Why do we need this?

modern/src/MainPage.js
```diff
@@ -158,7 +158,7 @@ const MainPage = () => {
           </Toolbar>
         </Paper>
         <div className={classes.deviceList}>
-          <DevicesList />
+          <DevicesList filter={deviceName} />
```
tananaev, line 161: Can we rename `deviceName` variable to something better? Like `searchKeyword` or something.

modern/src/DevicesList.js
```diff
@@ -126,12 +126,21 @@ const DeviceRow = ({ data, index, style }) => {
...
   const dispatch = useDispatch();
   const listInnerEl = useRef(null);
 
   const items = useSelector(getDevices);
+  const [filteredItems, setFilteredItems] = useState(null);
+
+  useEffect(() => {
+    setFilteredItems(
+      filter.trim().length > 0
+        ? items.filter((item) => `${item.name} ${item.uniqueId} ${item.groupId}`.toLowerCase().includes(filter?.toLowerCase()))
```
tananaev, line 140: Doesn't make sense to use `item.groupId` here. It's an internal id.

modern/src/store/devices.js
lucasleite01: Forgot to remove.

modern/src/MainPage.js
lucasleite01: Sure.

tananaev:
Looks like there's a conflict. Can you please fix it. Otherwise looks good.

tananaev:
Merged, thank you.

PR #917, opened 2022-02-09
Add notification snackbar

lucasleite01:
Notification snackbar added when some event arrives from websocket:

snackbar

### DIFF
--- a/modern/src/SocketController.js
+++ b/modern/src/SocketController.js
@@ -1,34 +1,53 @@
-import { useRef } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { useDispatch, useSelector, connect } from 'react-redux';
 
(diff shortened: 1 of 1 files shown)


tananaev:
Why are we replacing this?

lucasleite01:
I couldn't see any notification on screen when events arrived even with the `displayNotifications` function. Perhaps some misconfiguration in my environment?

tananaev:
Have you enabled permissions?

lucasleite01:
Notification permissions on Traccar server? Yes.

tananaev:
Not sure why it doesn't work then. You should probably investigate it instead of re-implementing it.

lucasleite01:
Right. I'll do that and if I don't find anything I'll close this PR.

PR #919, opened 2022-02-11
Update notification to snackbar

lucasleite01:
Snackbar displaying device name and event.

image

### DIFF
--- a/modern/package.json
+++ b/modern/package.json
@@ -17,6 +17,7 @@
     "maplibre-gl": "^1.15.0",
     "material-ui-dropzone": "^3.5.0",
     "moment": "^2.29.1",
+    "notistack": "^1.0.10",
     "react": "^17.0.2",
     "react-container-dimensions": "^1.4.1",
     "react-dom": "^17.0.2",
--- a/modern/package.json
+++ b/modern/package.json
@@ -17,7 +17,6 @@
     "maplibre-gl": "^1.15.0",
     "material-ui-dropzone": "^3.5.0",
     "moment": "^2.29.1",
-    "notistack": "^1.0.10",
     "react": "^17.0.2",
     "react-container-dimensions": "^1.4.1",
     "react-dom": "^17.0.2",
--- a/modern/src/SocketController.js
+++ b/modern/src/SocketController.js
@@ -75,13 +75,13 @@ const SocketController = () => {
   }, [authenticated]);
 
   useEffect(() => {
-    setNotifications(events.map((event) => {
-      event.message = `${devices[event.deviceId]?.name}: ${t(prefixString('event', event.type))}`;
-      event.show = true;
-      setTimeout(() => event.show = false, 5000);
-      return event;
-    }));
-  }, [events]);
+    setNotifications(events.map((event) => ({
+      id: event.id,
+      message: `${devices[event.deviceId]?.name}: ${t(prefixString('event', event.type))}`,
+      show: true,
+      close: () => setEvents(events.filter((e) => e.id !== event.id)),
+    })));
+  }, [events, devices]);
 
   return (
     <>
@@ -94,6 +94,8 @@ const SocketController = () => {
           }}
           open={notification.show}
           message={notification.message}
+          autoHideDuration={5000}
+          onClose={notification.close}
         />
       ))}
     </>
--- a/modern/src/SocketController.js
+++ b/modern/src/SocketController.js
@@ -79,7 +79,6 @@ const SocketController = () => {
       id: event.id,
       message: `${devices[event.deviceId]?.name}: ${t(prefixString('event', event.type))}`,
       show: true,
-      close: () => setEvents(events.filter((e) => e.id !== event.id)),
     })));
   }, [events, devices]);
 
@@ -95,7 +94,7 @@ const SocketController = () => {
           open={notification.show}
           message={notification.message}
           autoHideDuration={5000}
-          onClose={notification.close}
+          onClose={() => setEvents(events.filter((e) => e.id !== notification.id))}
         />
       ))}
     </>
--- a/modern/src/SocketController.js
+++ b/modern/src/SocketController.js
@@ -1,6 +1,6 @@
-import { useRef } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { useDispatch, useSelector, connect } from 'react-redux';
-import { useSnackbar } from 'notistack';
+import { Snackbar } from '@material-ui/core';
 import { useHistory } from 'react-router-dom';
 
 import { positionsActions, devicesActions, sessionActions } from './store';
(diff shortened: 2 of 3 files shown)


modern/src/SocketController.js
```diff
@@ -83,7 +77,27 @@ const SocketController = () => {
     return null;
   }, [authenticated]);
 
-  return null;
+  useEffect(() => {
+    events.forEach((event) => {
+      setNotification({
+        message: `${devices[event.deviceId]?.name}: ${t(prefixString('event', `${event.type}`))}`,
```
tananaev, line 83: Can we do this?

```suggestion
        message: `${devices[event.deviceId]?.name}: ${t(prefixString('event', event.type))}`,
```

modern/src/SocketController.js
```diff
@@ -83,7 +77,27 @@ const SocketController = () => {
...
+  }, [events]);
+
+  return (
+    <Snackbar
+      anchorOrigin={{
+        vertical: 'bottom',
+        horizontal: 'right',
+      }}
+      open={notification.show}
+      autoHideDuration={5000}
```
tananaev, line 97: Why do we have both `autoHideDuration` and `setTimeout`? Can you please explain how it works.

modern/src/SocketController.js
```diff
@@ -83,7 +77,27 @@ const SocketController = () => {
     return null;
   }, [authenticated]);
 
-  return null;
+  useEffect(() => {
+    events.forEach((event) => {
+      setNotification({
```
tananaev, line 82: This doesn't seem to support multiple events.

modern/src/SocketController.js
```diff
@@ -1,35 +1,29 @@
...
 
   const authenticated = useSelector((state) => !!state.session.user);
+  const devices = useSelector((state) => state.devices.items);
 
   const socketRef = useRef();
 
+  const [events, setEvents] = useState([]);
+  const [notification, setNotification] = useState({
+    message: '',
+    show: false,
```
tananaev, line 24: Why do we need two variables? Can't we just use `message`?

modern/src/SocketController.js
```diff
@@ -1,35 +1,29 @@
-import { useRef } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
```
tananaev, line 1: Where is `React` used?

tananaev:
Can you also please rename issue to something better.

modern/package.json
```diff
@@ -17,6 +17,7 @@
     "maplibre-gl": "^1.15.0",
     "material-ui-dropzone": "^3.5.0",
     "moment": "^2.29.1",
+    "notistack": "^1.0.10",
```
tananaev, line 20: I don't like adding new library just for supporting multiple notifications. Can't we just stack default ones?

lucasleite01:
Anton, your question about multiple events pointed a big problem in the implementation, so I searched and found [notistack](https://iamhosseindhv.com/notistack) package. 

It uses the same Snackbar from MaterialUI and handle multiple Snackbars by enqueuing the messages and showing them later with a better control. I think this made the implementation more simple. 

Now, the notification appears without device name, though, because when the notification arrives we cannot guarantee that we already have loaded the devices:

image

tananaev:
Not showing device is not acceptable. It basically makes it pointless.

lucasleite01:
> Not showing device is not acceptable. It basically makes it pointless.

Right. We need to think in a better way to ensure that we cover multiple events notification and to have the devices when showing the notification.

tananaev:
I also don't like adding dependency. Is there no way to display multiple toast as is?

modern/package.json
lucasleite01: I think we can, but we basically would have to do that notistack does. There's a [discussion about stacking snackbars](https://github.com/mui/material-ui/issues/1824).

tananaev: I would say if it's too complicated or requires too much code, let's only support a single one.

lucasleite01: Agreed, this was my first approach. But since we cannot control when we will receive events from websocket we need to have some kind of list or queue for the notifications that arrive.

Otherwise, if we receive an event from websocket and just update the component state it could lead to too much re-rendering if many events arrive in short time, causing error.

tananaev: Why would there be an error?

lucasleite01: Because an event can arrive and change the component state, generating a re-rendering just in the same time the component is being rendered.

tananaev: That's fine. Why would cause an error?

lucasleite01: Sorry, I might be mistaken:
> “Too many re-renderers” is a React error that happens after you have reached an infinite render loop, typically caused by code that in a useEffect hook or the main body of the component itself unconditionally calls state setters.

I'll have to try something like the first approach but keeping in mind the multiple events concern.

modern/src/SocketController.js
```diff
@@ -83,7 +74,30 @@ const SocketController = () => {
     return null;
   }, [authenticated]);
 
-  return null;
+  useEffect(() => {
+    setNotifications(events.map((event) => {
```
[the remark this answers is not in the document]
lucasleite01: > Why do we need to save events in the state? Can't we just do this directly and call setNotifications?

I tested that way, but here we don't have the devices yet:
```
socket.onmessage = (event) => {
      const data = JSON.parse(event.data);
      if (data.devices) {
        dispatch(devicesActions.update(data.devices));
      }
      if (data.positions) {
        dispatch(positionsActions.update(data.positions));
      }
      if (data.events) {
        // setEvents(data.events);
        // if we use setNotifications here, we don't have devices yet to construct the message.
        // something like that will return undefined for devices name:
        setNotifications(events.map((event) => {
          event.message = `${devices[event.deviceId]?.name}: ${t(prefixString('event', event.type))}`;
          event.show = true;
          setTimeout(() => event.show = false, 5000);
          return event;
        }));
      }
    };
```

> Also, can we please not reuse event object. It gets confusing. Create a new object for notifications.

Do you mean to rename `events`?
```
const [events, setEvents] = useState([]);
```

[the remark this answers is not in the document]
tananaev: I see what you mean. In that case:

1. Don't reuse the same event object. Create a new object for notification.
2. Set devices as a dependency.

modern/src/SocketController.js
```diff
@@ -83,7 +74,30 @@ const SocketController = () => {
...
+      event.message = `${devices[event.deviceId]?.name}: ${t(prefixString('event', event.type))}`;
+      event.show = true;
+      setTimeout(() => event.show = false, 5000);
+      return event;
+    }));
+  }, [events]);
```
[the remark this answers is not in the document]
lucasleite01: This uses devices, but we don't want this to be triggered every time devices update because devices are updated frequently. If we add devices to the array the snackbars will be displayed every time new devices arrive even a new event wasn't arrived.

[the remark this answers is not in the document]
tananaev: No, wait.. you need to clear notifications once they are dismissed. Looks like that's another thing that's missing.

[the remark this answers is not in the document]
lucasleite01: That's right. I'll do that.

modern/src/SocketController.js
```diff
@@ -83,7 +74,32 @@ const SocketController = () => {
     return null;
   }, [authenticated]);
 
-  return null;
+  useEffect(() => {
+    setNotifications(events.map((event) => ({
+      id: event.id,
+      message: `${devices[event.deviceId]?.name}: ${t(prefixString('event', event.type))}`,
+      show: true,
+      close: () => setEvents(events.filter((e) => e.id !== event.id)),
```
tananaev, line 82: I don't like that there's a callback in a data object. Why do we need it here? We can just call it directly in `onClose`.

tananaev (commented):
Everything else looks good.

modern/src/SocketController.js
lucasleite01: Done.

tananaev:
Merged, thank you.

PR #920, opened 2022-02-12
improving italian translation + add missing key in en.json

arkanoid87:
syncing keys between italian and english json
matching capitalization strategy of en.json


tananaev:
The "missing key" is the one we already removed. As for localization, you shou do it via Transifex:

arkanoid87:
my fault! thanks for the link

arkanoid87:
Italian translation 100%

tananaev:
Thanks. It will be picked up with the next official release.

Issue #931, opened 2022-05-05
l10n new translation:  "mapClustering": "Clustering"


castillo92:
https://github.com/traccar/traccar-web/blob/644a598fb76d23a042a69a45806623ccb9fb79de/web/l10n/en.json#L276

What does it means? I ask to make the best translation in Transifex

tananaev:
Map clustering means combining closeby markers on a map to avoid clutter.

castillo92:
Thanks for the explanation Anton,

What do you think about change "Clustering" by "Markers clustering"

tananaev:
I don't mind changing. Feel free to send a pull request.

castillo92:
Created a pull request.

PR #933, opened 2022-05-17
Update en.json

castillo92:
Improving new label been more specific to allow itself explanation.

Reference:

https://github.com/traccar/traccar-web/issues/931#issuecomment-1118537042


tananaev:
Merged, thanks.

PR #934, opened 2022-05-18
Add OS Maps Data Hub integration

OutrightHosting:
Add Ordnance Survey Maps Data Hub API integration to modern interface. 

Detailed UK mapping free tier up to £1000 usage / month

Demand - https://www.traccar.org/forums/topic/custom-map-coordinate-conversion/

Documentation - https://osdatahub.os.uk/

### DIFF
--- a/modern/src/map/core/Map.js
+++ b/modern/src/map/core/Map.js
@@ -8,7 +8,7 @@ import { SwitcherControl } from '../switcher/switcher';
 import deviceCategories from '../../common/util/deviceCategories';
 import { prepareIcon, loadImage } from './mapUtil';
 import {
-  styleCarto, styleLocationIq, styleMapbox, styleMapTiler, styleOsm,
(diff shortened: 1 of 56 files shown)


tananaev:
I feel like you can just use this map as a custom map. It doesn't really need to be added, especially considering that it's not free and it's only for UK.

OutrightHosting:
@tananaev Ordnance Survey looks like it has crept in to traccar, if I update this to be current can we implement this to allow users to use their own key and select between the two map types?

Projection also appears to be off as this device should be on the south coast of UK

image

tananaev:
From what I understand, there are no limitations on the key, so we just used our own key. Maybe I'm wrong.

As for projection, this should fix it: https://github.com/traccar/traccar-web/commit/eeaa67a1fcafb2a95bea450035504726b9481838.

PR #936, opened 2022-05-22
fix: correct children key issue

BIhab:
Hi Anton,
Please accept my gratitude for all the effort you're doing to improve the code.
Concerning the child key warning, I found that is was only fixed for "RouteReport", for this specific report:
['StopReport','SummaryReport','TripReport']

The warning did not get fixed, the cause was that in the api result there is no "id" attribute for items, I choose the array index instead to make the key unique.

For the 'EventReport', in addition to TableCell, MenuItem also need a unique key.


BIhab:
The test job didn't accept the array index as a unique key, should I modify the API directly to send a unique id with each item?

tananaev:
Can you please rebase your PR.

tananaev:
Ah the issue is with the array. I don't want to add something to the API. Is there any other way?

BIhab:
What about using: startPositionId, I think that the positionId is unique for each item, if that's correct, I'll use "startPositionId" instead of array index

tananaev:
Sounds good to me.

modern/src/reports/SummaryReportPage.js
```diff
@@ -99,8 +98,8 @@ const SummaryReportPage = () => {
             </TableRow>
           </TableHead>
           <TableBody>
-            {items.map((item, index) => (
-              <TableRow key={index}>
+            {items.map((item) => (
+              <TableRow key={item.deviceId}>
```
tananaev, line 102: This probably won't work for daily report.

BIhab: I want to ask you to check that all the key I used are unique one, if deviceId not a unique key for the SummaryReport, I'm just using the demo generated by test-generator, and I got only one row as result on summary Report I couldn't find if deviceId is a unique key or not. there is also a console.log that I forgot on Statistics Report

tananaev: Can you combine device id + time as a key here?

BIhab: I combined as you suggested deviceId with startTime (ms) and separated them with ' _ '

modern/src/reports/SummaryReportPage.js
```diff
@@ -100,7 +102,7 @@ const SummaryReportPage = () => {
           </TableHead>
           <TableBody>
             {items.map((item) => (
-              <TableRow key={item.id}>
+              <TableRow key={generateKey(item)}>
```
tananaev, line 105: I would just inline it. Easier to maintain.

modern/src/reports/TripReportPage.js
```diff
@@ -79,7 +79,6 @@ const TripReportPage = () => {
         return item[key];
     }
   };
-
```
tananaev, line 82: revert this please

tananaev:
Merged, thanks for fixing this.

BIhab:
> Merged, thanks for fixing this.

It was a pleasure working with you

PR #966, opened 2022-06-20
feature: max zoom on map

BIhab:
Hi Anton, 
I hope you're fine, please review the request and feel free to suggest improvements.

### DIFF
--- a/modern/src/map/core/MapView.js
+++ b/modern/src/map/core/MapView.js
@@ -78,6 +78,11 @@ const MapView = ({ children }) => {
   const [activeMapStyles] = usePersistedState('activeMapStyles', ['locationIqStreets', 'osm', 'carto']);
   const [defaultMapStyle] = usePersistedState('selectedMapStyle', 'locationIqStreets');
   const mapboxAccessToken = useAttributePreference('mapboxAccessToken');
+  const maxZoom = useAttributePreference('web.maxZoom') || 16;
+
+  useEffect(() => {
+    map.setMaxZoom(maxZoom);
+  }, [maxZoom]);
 
   useEffect(() => {
     maplibregl.accessToken = mapboxAccessToken;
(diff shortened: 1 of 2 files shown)


BIhab:
I choose default value on 16 to avoid marker mispositioning due to zoom -  fix#964

modern/src/map/core/MapView.js
```diff
@@ -78,6 +78,11 @@ const MapView = ({ children }) => {
   const [activeMapStyles] = usePersistedState('activeMapStyles', ['locationIqStreets', 'osm', 'carto']);
   const [defaultMapStyle] = usePersistedState('selectedMapStyle', 'locationIqStreets');
   const mapboxAccessToken = useAttributePreference('mapboxAccessToken');
+  const maxZoom = useAttributePreference('web.maxZoom') || 16;
```
tananaev, line 81: The only part that I don't like is the hardcoded default value.

tananaev: I don't think there should be any. Or there might be some depending on the map layer.

modern/src/map/core/MapView.js
```diff
@@ -78,6 +78,11 @@ const MapView = ({ children }) => {
   const [activeMapStyles] = usePersistedState('activeMapStyles', ['locationIqStreets', 'osm', 'carto']);
   const [defaultMapStyle] = usePersistedState('selectedMapStyle', 'locationIqStreets');
   const mapboxAccessToken = useAttributePreference('mapboxAccessToken');
+  const maxZoom = useAttributePreference('web.maxZoom');
+
+  useEffect(() => {
+    map.setMaxZoom(maxZoom);
```
tananaev, line 84: Should we do a check here to make sure there's a value? Or it doesn't matter?

BIhab: Normaly we should, to avoid any bug in the future, another remark, when you set an attribute he's not persisting, as long as you're log in, the value stay in session but it's not stored on database, and as soon as you logout, the value disappear, can you try on your side just to know if I'm the only one facing that issue, if I'm not I'll open a ticket.

tananaev: What do you mean it's not persisted? Is it not saved in the user attributes?

tananaev:
Merged, thank you.

PR #969, opened 2022-06-21
Added option to disable map styles

WGrobler:
I want to disable certain map styles server wide, therefore I included attributes for each provider as well as an option to construct the styles array based on the attributes.  To limit the attributes, I created it on provider level; we can change it to individual level, to allow disabling of individual styles.

I would like to hear thoughts and possible improvements


tananaev (commented):
I don't think it makes sense to make an attribute for each map. I think one attribute with a list should do it. And also it should probably be server-only attribute.

WGrobler:
When you say "one attribute with a list", are you thinking of a string attribute with comma separated values or a list of maps linked to a bool attribute (like I did here)?

You are right that the attributes should be server-only, I'm still familiarising myself with the entire codebase, how do I set it server only? is there a `useAttributePreference` for server only or should I do something similar to [useFeatures.js#L3](https://github.com/traccar/traccar-web/blob/b418d001a498b5a5108bb13983b4f3aba6c60dbc/modern/src/common/util/useFeatures.js#L3)

tananaev:
Comma-separated list of disabled maps.. or maybe enabled.

You don't need to use `useAttributePreference`. You can just get server from redux state directly.

WGrobler:
I was also initially considering a list, thought bools would be easier since I can include it in the common attributes so that it can autofill, but since this would be a server setup it does not require autofill for each style. I agree with you and will change it to a list instead. 

I think the list needs to be a disabled list. Most instances will enable everything by default but then this is the option to disable. Otherwise we complicate existing instances.

Looking at the code for `useAttributePreference` I can see that it forces the server setting, therefore is it required to make this server only? It might not be used for users, but does it matter if its possible for them to set it since they can't override server?

WGrobler:
Is there an common/attributes for server only attributes or must it be included in [attributes/useCommonUserAttributes.js](https://github.com/traccar/traccar-web/blob/b418d001a498b5a5108bb13983b4f3aba6c60dbc/modern/src/common/attributes/useCommonUserAttributes.js)?

tananaev:
> therefore is it required to make this server only?

Yes. We don't want to pollute user configuration. And also "force server settings" is a flag that's disabled by default.

Common attributes are common for server and user. Instead you should create a separate one for server. Similar to how we have `useUserAttributes`.

WGrobler:
Made changes as discussed and create a new PR #970

PR #970, opened 2022-06-21
Disable styles

WGrobler:
Update to PR #969 

- "Disable Styles" now uses a comma separated string
- Added a common attributes for server only attributes 
- Added `disableStyles` to useCommonServerAttributes
- Updated EN l10n to include `attributeUiDisableStylesAttributes`


modern/src/common/attributes/useCommonServerAttributes.js
```diff
@@ -0,0 +1,8 @@
+import { useMemo } from 'react';
+
+export default (t) => useMemo(() => ({
```
tananaev, line 3: This file should be called `useServerAttributes` because it doesn't have any common attributes.

modern/src/map/core/useMapStyles.js
```diff
@@ -169,20 +171,27 @@ export default () => {
...
+    custom: {
       id: 'custom',
       title: t('mapCustom'),
       style: styleCustom([customMapUrl]),
       available: !!customMapUrl,
     },
-  ];
+  };
+
+  Object.keys(Styles).forEach((key) => {
```
tananaev, line 191: I think the right think is to keep the array as is and just filter it. Otherwise the order is not guaranteed and you kind of have duplication of the ids.

web/l10n/en.json
```diff
@@ -118,6 +118,7 @@
     "attributeUiDisableCalendars": "UI: Disable Calendars",
     "attributeUiDisableMaintenance": "UI: Disable Maintenance",
     "attributeUiHidePositionAttributes": "UI: Hide Position Attributes",
+    "attributeUiDisableStylesAttributes": "UI: Disable Styles",
```
tananaev, line 121: The key should be renamed to something like `attributeUiDisableMapLayers`. It definitely shouldn't have `Attributes` in the name. And also "styles" won't be as clear for people because it's a technical term specifically used by this map library. It's not a common term people would understand.

And the value should be "UI: Disable Map Layers".

modern/src/common/attributes/useCommonServerAttributes.js
WGrobler: Agreed, it probably wont be that common. Changed it in [60995ed](https://github.com/traccar/traccar-web/pull/970/commits/60995ed858af7da0493c4fd84f38f3c7709e2474)

web/l10n/en.json
WGrobler: Agreed. I don't know why I double added the word attribute(s). Also I changed the word styles to layers on the attributes side; within the useMapStyles script it still uses the word styles to align with the script. 

Changed was made in [22ef038](https://github.com/traccar/traccar-web/pull/970/commits/22ef038f139f0bc6be7f97f792f0988356593204)

modern/src/map/core/useMapStyles.js
WGrobler: Are you suggesting that we use an array for `Styles` (like it currently is in master) and delete styles from it if found in the `disabledStyles` list? Or do you suggest that we still loop over `Styles` and create `activeStyles`, but only changing `Styles` back to a list instead of an dict?

I thought about looping though the `disabledStyles` set, looking for the id in `Styles` and removing it if found. The problem is that it would require a search though the entire `Styles` array for every entry in `disabledStyles`. Since `disabledStyles` will always be smaller than `Styles`, I decided that it would be better to perform the "has key" check on the `disabledStyles` set.

Changing `Styles` back to an array and still using the same logic where its copied over to `activeStyles` if not in  `disabledStyles`, should solve your order and id duplication concerns.

WGrobler: > Changing `Styles` back to an array and still using the same logic where its copied over to `activeStyles` if not in `disabledStyles`, should solve your order and id duplication concerns.

I made this change with [d17e1b3](https://github.com/traccar/traccar-web/pull/970/commits/d17e1b3e2d6d539695e03c49c6c30b1481c06750)

modern/src/map/core/useMapStyles.js
```diff
@@ -185,4 +187,11 @@ export default () => {
       available: !!customMapUrl,
     },
   ];
+
+  for (let i = 0; i < Styles.length; i += 1) {
+    if (!disabledStyles.has(Styles[i].id)) {
+      activeStyles.push(Styles[i]);
+    }
+  }
+  return activeStyles;
```
tananaev, line 196: Just use `styles.filter(...)` instead of doing this loop. And no need for a separate variable in that case.

modern/src/settings/ServerPage.js
```diff
@@ -58,6 +59,7 @@ const ServerPage = () => {
   const t = useTranslation();
 
   const commonUserAttributes = useCommonUserAttributes(t);
+  const commonServerAttributes = useServerAttributes(t);
```
tananaev, line 62: You forgot to rename the variable.

tananaev (commented):
Added some small comments, but overall looks pretty good.

tananaev:
Any plans to update this or should we close it for now?

PR #973, opened 2022-06-28
update Thai

tomzt:


tananaev:

tomzt:
Sorry! I’ll join

tananaev:
Thank you for the contribution.

PR #977, opened 2022-06-30
fix: live route blank page issue

BIhab:
Hi Anton,
There is a bug concerning live route, if you tried it on Demo Traccar you'll get a blank page with error: route.findLast(...) is not a function, I think the findLast was set to get the last item of the route array, if yes, I simply change the function.

### DIFF
--- a/modern/src/map/main/MapLiveRoutes.js
+++ b/modern/src/map/main/MapLiveRoutes.js
@@ -60,7 +60,7 @@ const MapLiveRoutes = () => {
         setRoute([position]);
       }
     } else if (position) {
-      const last = route.findLast(() => true);
+      const last = route.at(-1);
       if (!last || (last.latitude !== position.latitude && last.longitude !== position.longitude)) {
         setRoute([...route.slice(-9), position]);
       }


tananaev:
What's the error and what browser?

BIhab:
> What's the error and what browser?

Concerning the error:

Live route when you select a device.

To reproduce:
- Enable Live Route on Setting
- Go back to the Device List
- Select a Device

Dev Console:
Uncaught TypeError: o.findLast is not a function

Browser: 
Firefox

tananaev:
`findLast` should be supported in all modern browsers. And definitely in Firefox:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/lastIndexOf#browser_compatibility

Would like to know why it fails.

BIhab:
> `findLast` should be supported in all modern browsers. And definitely in Firefox:
> 
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/lastIndexOf#browser_compatibility
> 
> Would like to know why it fails.

The link is for .lastIndexOf not .findLast, I did a quick research, and I found that:

https://caniuse.com/mdn-javascript_builtins_array_findlast

I think it's not supported yet on Firefox and on some Mobile browser.

tananaev:
Makes sense. Merged, thank you.

PR #984, opened 2022-07-08
Fix: Fab Button now shown on mobile devices

BIhab:
Hi Anton, 
I hope you're doing fine, the fix is about Fab button not shown on mobile Device, the cause was that the existing code was concatening string values, and the output for Bottom attribute css was `bottom: 5616` instead of `bottom:72`, I parsed the string to become Int.

### DIFF
--- a/modern/src/settings/components/CollectionFab.js
+++ b/modern/src/settings/components/CollectionFab.js
@@ -11,7 +11,7 @@ const useStyles = makeStyles((theme) => ({
     bottom: theme.spacing(2),
     right: theme.spacing(2),
     [theme.breakpoints.down('md')]: {
-      bottom: parseInt(theme.dimensions.bottomBarHeight, 10) + parseInt(theme.spacing(2), 10),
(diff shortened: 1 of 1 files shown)


modern/src/settings/components/CollectionFab.js
```diff
@@ -11,7 +11,7 @@ const useStyles = makeStyles((theme) => ({
     bottom: theme.spacing(2),
     right: theme.spacing(2),
     [theme.breakpoints.down('md')]: {
-      bottom: theme.dimensions.bottomBarHeight + theme.spacing(2),
+      bottom: parseInt(theme.dimensions.bottomBarHeight, 10) + parseInt(theme.spacing(2), 10),
```
tananaev, line 14: I commented on the other PR. You should use `calc` option here instead of parsing values.

tananaev (approved):

tananaev:
Merged, thank you.

PR #991, opened 2022-07-20
Translate Dropzone text and Device Image

DiSousaDev:
Translate to pt-BR and add dinamic translate to dropzone.


tananaev:
Localization should be done through Transifex:

And if we want to localize the file upload component, we should probably localize all of the strings and make it more generic.

PR #992, opened 2022-07-20
Localizing DropzoneArea

DiSousaDev:
Localizing drag and drop component.


web/l10n/en.json
```diff
@@ -517,5 +517,6 @@
     "categoryVan": "Van",
     "categoryScooter": "Scooter",
     "maintenanceStart": "Start",
-    "maintenancePeriod": "Period"
+    "maintenancePeriod": "Period",
+    "dropzoneText": "Drag and drop a file here or click"
```
tananaev, line 521: We use prefixes for string. This one can probably go into "shared" once because it's used in multiple places. So name it something like `sharedDropzoneText` and put next to other shared string.

modern/src/settings/CalendarPage.js
```diff
@@ -66,6 +66,7 @@ const CalendarPage = () => {
                 label={t('sharedName')}
               />
               <DropzoneArea
+                dropzoneText={t('dropzoneText')}
```
tananaev, line 69: What about other unlocalized text related to `DropzoneArea`? There are toasts and some other strings.

tananaev (commented):
What about other

web/l10n/en.json
DiSousaDev: Thx for this review! Solved.

modern/src/settings/DevicePage.js
```diff
@@ -150,10 +150,15 @@ const DevicePage = () => {
...
                 <DropzoneArea
+                  dropzoneText={t('sharedDropzoneText')}
                   acceptedFiles={['image/*']}
                   filesLimit={1}
                   onChange={handleFiles}
                   alertSnackbarProps={components.MuiSnackbar.defaultProps}
+                  getFileLimitExceedMessage={(filesLimit) => `${t('sharedMaxFileLimit')} ${t('sharedOnly')} ${filesLimit} ${t('sharedAllowed')}`}
+                  getFileAddedMessage={(fileName) => `${t('sharedFile')} ${fileName} ${t('sharedSuccessfullyAdded')}`}
+                  getFileRemovedMessage={(fileName) => `${t('sharedFile')} ${fileName} ${t('sharedRemoved')}`}
+                  getDropRejectMessage={(file) => `${t('sharedFile')} ${file.name} ${t('sharedDropRejectFile')} ${file.size} ${t('sharedMegabytes')}`}
```
tananaev, line 161: Is there a way to just disable these toasts?

Also, combining a string from different words like this doesn't really work for translation. Different languages might have different order etc.

DiSousaDev: Yesn Anton i can disable this.

modern/src/settings/CalendarPage.js
```diff
@@ -66,9 +66,11 @@ const CalendarPage = () => {
                 label={t('sharedName')}
               />
               <DropzoneArea
+                dropzoneText={t('sharedDropzoneText')}
                 filesLimit={1}
                 onChange={handleFiles}
                 alertSnackbarProps={components.MuiSnackbar.defaultProps}
```
tananaev, line 72: We probably don't need this if alerts are disabled.

tananaev (commented):
One last comment. Otherwise looks good.

modern/src/settings/CalendarPage.js
DiSousaDev: you are right! we really don't need it. done!

tananaev:
Merged, thank you for the contribution.

PR #994, opened 2022-07-21
Adds Pushover user attributes

lzkill:


tananaev:
You should only change `en.json`. All others are generated from Transifex.

lzkill:
Nice! Since I've already translated all files, is there anything I still need to do?

tananaev:
You need to remove all localization from the PR. It doesn't work like that. Those files will be overwritten.

lzkill:
> You need to remove all localization from the PR. It doesn't work like that. Those files will be overwritten.

Hi Anton. Do I have to do anything else here? This PR is supposed to be merged along with [traccar/pull/4908](https://github.com/traccar/traccar/pull/4908). The main idea is to allow Pushover notifications to be configured per user just like Telegram.

tananaev:
I still see all the localization files. You have to revert those changes.

PR #1001, opened 2022-07-30
Floating status card

lzkill:
This PR adds initial code in order to provide such a functionality. It can be largely customized as pointed by [react-draggable's documentation](https://www.npmjs.com/package/react-draggable).

### DIFF
--- a/modern/src/main/StatusCard.js
+++ b/modern/src/main/StatusCard.js
@@ -212,35 +212,30 @@ const StatusCard = ({ deviceId, onClose }) => {
               <IconButton
                 color="secondary"
                 onClick={(e) => setAnchorEl(e.currentTarget)}
-                onTouchStart={(e) => setAnchorEl(e.currentTarget)}
                 disabled={!position}
               >
                 <PendingIcon />
               </IconButton>
               <IconButton
                 onClick={() => navigate('/replay')}
-                onTouchStart={() => navigate('/replay')}
                 disabled={!position}
               >
                 <ReplayIcon />
               </IconButton>
               <IconButton
                 onClick={() => navigate(`/settings/command-send/${deviceId}`)}
-                onTouchStart={() => navigate(`/settings/command-send/${deviceId}`)}
                 disabled={readonly}
               >
                 <PublishIcon />
               </IconButton>
               <IconButton
                 onClick={() => navigate(`/settings/device/${deviceId}`)}
-                onTouchStart={() => navigate(`/settings/device/${deviceId}`)}
                 disabled={deviceReadonly}
               >
                 <EditIcon />
               </IconButton>
               <IconButton
                 onClick={() => setRemoving(true)}
-                onTouchStart={() => setRemoving(true)}
                 disabled={deviceReadonly}
                 className={classes.negative}
               >
--- a/modern/src/main/StatusCard.js
+++ b/modern/src/main/StatusCard.js
@@ -157,7 +157,7 @@ const StatusCard = ({ deviceId, onClose }) => {
     <>
       {device && (
         <Draggable
-          handle={deviceImage ? `.${classes.media}` : `.${classes.header}`}
+          handle={`.${classes.media}, .${classes.header}`}
         >
           <Card elevation={3} className={classes.card}>
             {deviceImage ? (
--- a/modern/src/main/StatusCard.js
+++ b/modern/src/main/StatusCard.js
@@ -157,7 +157,7 @@ const StatusCard = ({ deviceId, onClose }) => {
     <>
       {device && (
         <Draggable
-          handle={`.${classes.header}`}
+          handle={deviceImage ? `.${classes.media}` : `.${classes.header}`}
         >
           <Card elevation={3} className={classes.card}>
             {deviceImage ? (
--- a/modern/src/main/StatusCard.js
+++ b/modern/src/main/StatusCard.js
@@ -212,30 +212,35 @@ const StatusCard = ({ deviceId, onClose }) => {
               <IconButton
                 color="secondary"
                 onClick={(e) => setAnchorEl(e.currentTarget)}
+                onTouchStart={(e) => setAnchorEl(e.currentTarget)}
                 disabled={!position}
               >
                 <PendingIcon />
               </IconButton>
               <IconButton
                 onClick={() => navigate('/replay')}
+                onTouchStart={() => navigate('/replay')}
                 disabled={!position}
               >
                 <ReplayIcon />
               </IconButton>
               <IconButton
                 onClick={() => navigate(`/settings/command-send/${deviceId}`)}
+                onTouchStart={() => navigate(`/settings/command-send/${deviceId}`)}
                 disabled={readonly}
               >
                 <PublishIcon />
               </IconButton>
               <IconButton
                 onClick={() => navigate(`/settings/device/${deviceId}`)}
+                onTouchStart={() => navigate(`/settings/device/${deviceId}`)}
                 disabled={deviceReadonly}
               >
                 <EditIcon />
               </IconButton>
               <IconButton
                 onClick={() => setRemoving(true)}
+                onTouchStart={() => setRemoving(true)}
                 disabled={deviceReadonly}
                 className={classes.negative}
               >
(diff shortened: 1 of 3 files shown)


tananaev:
What do you mean by "initial code"? Is this fully functional or there are some issues?

lzkill:
> What do you mean by "initial code"? Is this fully functional or there are some issues?

It's fully functional. By initial I mean no customization has been made, it's the "hello world" of react-draggable. I didn't want to make any decisions by myself. The component's docs showcases many possibilities, like restricting where the draggable card can go.

tananaev:
Please fix the issue.

BIhab:
Hi @lzkill ,
After testing your solution, an issue will occur when trying to click on any element that have an onClick event - button, a ... - on mobile devices, a workaround is proposed in react-grid-layout/react-draggable#550

tananaev:
Wait, none of the elements are clickable? :)

BIhab:
> Wait, none of the elements are clickable? :)

In the desktop, all work fine, but in mobile devices - even in devtools with touch activated - none of the element inside the <Draggable></Draggable> component are clickable until your add "onTouchStart" event to the clickable elements to make them handle the click, for this PR, the issue will affect all clickable icon inside StatusCard.

BIhab:
> > Wait, none of the elements are clickable? :)
> 
> In the desktop, all work fine, but in mobile devices - even in devtools with touch activated - none of the element inside the Draggable component are clickable until your add "onTouchStart" event to the clickable elements to make them handle the click, for this PR, the issue will affect all clickable icon inside StatusCard.

lzkill:
> > Wait, none of the elements are clickable? :)
> 
> In the desktop, all work fine, but in mobile devices - even in devtools with touch activated - none of the element inside the Draggable component are clickable until you add "onTouchStart" event to the clickable elements to make them handle the click, for this PR, the issue will affect all clickable icon inside StatusCard.

Nice! I'll take a look!

lzkill:
I've added the proposed fix, it's working now on chrome devtools, at least.

tananaev:
It didn't work in chrome devtools before? Can you please test a real device and maybe a couple other browsers, so we don't break every with this change.

lzkill:
Sure, I'll do it later today.

lzkill:
Hi pals! I've tested the PR on Chrome, Firefox, Opera, Edge, Samsung Internet, Kiwi Browser, Brave Beta and Safari, all using real devices. In the Desktop I've tested Chrome and Firefox. In all cases the draggable feature worked like a charm.

modern/src/main/StatusCard.js
```diff
@@ -129,82 +130,97 @@ const StatusCard = ({ deviceId, onClose }) => {
...
+                    ))}
+                  </TableBody>
+                </Table>
+              </CardContent>
+            )}
+            <CardActions classes={{ root: classes.actions }} disableSpacing>
+              <IconButton
+                color="secondary"
+                onClick={(e) => setAnchorEl(e.currentTarget)}
+                onTouchStart={(e) => setAnchorEl(e.currentTarget)}
```
tananaev, line 187: Do we need both `onClick` and `onTouchStart`?

lzkill: Yes, that's the workaround found by @BIhab that makes it work on mobile devices.

tananaev: Yeah, but wouldn't it cause double-calls in some cases? It's also mentioned in that ticket.

lzkill: Where did you read about double calls? What I've found in that ticket is somebody talking about the [double click event](https://github.com/react-grid-layout/react-draggable/issues/597) not being fired on touch devices. There is [another](https://github.com/react-grid-layout/react-draggable/issues/597#issuecomment-1029947048) solution proprosed that adds only an handler to react-draggable. It doesn't duplicate code, but feels even more hacky. What do you prefer? 

```
const handleOnDragStop = e => {
    const { target, type } = e;
    if (type === 'touchend') target.click();
    return true;
  };

...

<Draggable
    ...
    onStop={handleOnDragStop}
    ...
>
        ...
</Draggable>
```

tananaev: I saw this in one of the snippets, so I'm assuming it's to avoid duplicate calls:
```
  const eventHandler = cb => {
    return event => {
      if (isMobile && event.type === 'touchstart') {
        cb(event);
      } else if (!isMobile && event.type === 'click') {
        cb(event);
      }
    };
  };
```

BIhab: > I saw this in one of the snippets, so I'm assuming it's to avoid duplicate calls:
> ```
>   const eventHandler = cb => {
>     return event => {
>       if (isMobile && event.type === 'touchstart') {
>         cb(event);
>       } else if (!isMobile && event.type === 'click') {
>         cb(event);
>       }
>     };
>   };
> ```

Assuming that the called function is the same, why not using an || instead of else if? 
``` 
if ( (isMobile && event.type === 'touchstart') || (!isMobile && event.type === 'click') ) {
          cb(event);
}
```

tananaev: My point is mostly about safety of not doing the check, like it's implemented currently.

tananaev: By the way, what about clicks on the attributes? For example, does "Show Address" link work? Or accumulators.

lzkill: Yes, these links work as well.

tananaev: How come those links work without a need for `onTouchStart`?

lzkill: I'm sorry, they work only on the desktop just like the others.

tananaev: OK, then we have a big issue. I suggest we wait for a proper fix or find some other solution for dragging.

lzkill: For the address filed the `onTouchStart` hack works too, but for the accumulator I think it will get tricky. Guess it's better to hold on as you suggest.

tananaev: What if we set a "handle", so you can only drag by the header? Would it solve the problem with clickable content?

lzkill: We would still need to fix the close button. But that's a good idea, I'll try it.

lzkill: It worked. In fact the header handle is even more intuitive.

tananaev:
Can you please rebase on master to resolve the conflict.

lzkill:
> Can you please rebase on master to resolve the conflict.

Done.

modern/src/main/StatusCard.js
```diff
@@ -155,82 +156,94 @@ const StatusCard = ({ deviceId, onClose }) => {
...
-        <Card elevation={3} className={classes.card}>
-          {deviceImage ? (
-            <CardMedia
-              className={classes.media}
-              image={`/api/media/${device.uniqueId}/${deviceImage}`}
-            >
-              <IconButton size="small" onClick={onClose}>
-                <CloseIcon fontSize="small" className={classes.mediaButton} />
+        <Draggable
+          handle={`.${classes.header}`}
```
tananaev, line 160: What if you have an image as the header? It looks like you are not handling that case.

tananaev:
Also looks like there are some issues with package files.

modern/src/main/StatusCard.js
lzkill: I'll take a look.

modern/src/main/StatusCard.js
```diff
@@ -157,7 +157,7 @@ const StatusCard = ({ deviceId, onClose }) => {
     <>
       {device && (
         <Draggable
-          handle={`.${classes.header}`}
+          handle={deviceImage ? `.${classes.media}` : `.${classes.header}`}
```
tananaev, line 160: Does this work instead?

`.${classes.media}, .${classes.header}`

lzkill: It does, I'm changing it.

tananaev:
Merged, thank you.

lzkill:
> Merged, thank you.

Thank you for the things you taught me here.

PR #1003, opened 2022-08-01
BugFix: Form type field in AddAttributeDialog gets disabled when selecting a valid attribute option

Sam-damn:


tananaev:
If you select a pre-defined attribute, you shouldn't be able to change the type. What am I missing here?

Sam-damn:
@tananaev  now that you mention it i believe i made a mistake, i thought this was an unintended behavior, apologies.

PR #1031, opened 2022-09-26
Fixed Localization & Linter

tahirmanzoor120:


tananaev:
Doesn't look like a fix to me. Seems like you just duplicated the files.

PR #1035, opened 2022-10-09
Standardization API legend for each maps

jonathandreyer:


tananaev:
I would prefer not to do this because it will reset all the localization that has been done for this string.

jonathandreyer:
OK I see. I have update the French translation with this standardization.

PR #1041, opened 2022-10-22
Feature request, Parking timer on traccar-web

Zenyte-zz:
This is my very first feature request on Github, sorry if it is on wrong place?

Is it possible to show the Parking Time directly on the detailscreen?
There is a parameter "motion" behind the green button, maybe this can be used to calculate the parking minutes/hours?

e.g. No Motion since: "xx:xx:xx" minutes

My intention for this request: 
I´, delivering on a route from 9 am to 3 pm with round about 10 Stops a day, every 30-45 minutes to my customers with exactly defind appointments on time, sometimes the appointment to be at the customer is not possible because of traffic or other circumstances so the next customer knows we are little late.


tananaev:
Seems like you created a pull request instead. Create a feature request here:

https://github.com/traccar/traccar-web/issues/new/choose

PR #1077, opened 2023-02-01
Update SplitButton.js

joystickjockey:
Fix warnings when building in IntelliJ IDEA


tananaev:
How come I'm not getting any warnings? Are you using different lint settings?

tananaev:
Then why do you have this warning?

tananaev:
The lint check here is successful, so I think it's something on your end:

https://github.com/traccar/traccar-web/actions

tananaev:
Actually I tried to run the lint directly and I see it as a warning. Not sure why I'm still able to build everything without any issues.

PR #1080, opened 2023-02-08
Fix invalid lat/lon usePreference

WesleyKoech:
Fix out-of-range lat/lon error when user sets values outside of (-180, 180) and (-90, 90) respectively. The error is as shown:
Screenshot from 2023-02-08 12-53-35

### DIFF
--- a/modern/package-lock.json
+++ b/modern/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "traccar",
-  "version": "5.5.0",
+  "version": "5.6.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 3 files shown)


tananaev:
I think the right solution is to implement validation at the time you enter those values.

WesleyKoech:
It makes sense to add validations to both the endpoint for creating a user on the server-side and the user-add form in the web app.

tananaev:
Agree. Do you want to update this PR or create a separate one?

WesleyKoech:
Cool. I will update the current PR.

tananaev:
Recent versions of Traccar already have server side validation.

PR #1086, opened 2023-03-14
Improvement of Korean translation

sayzard:
Add and improve Korean translation


tananaev:
For localization please use Transifex:

https://transifex.com/traccar/

PR #1092, opened 2023-04-02
Frontend changes for OIDC

dan-r:
Main description in https://github.com/traccar/traccar/pull/5057


modern/src/resources/l10n/en.json
```diff
@@ -177,6 +177,8 @@
     "loginReset": "Reset Password",
     "loginRegister": "Register",
     "loginLogin": "Login",
+    "loginLoginOIDC": "Login with SSO",
```
tananaev, line 180: No all caps please. Also maybe name this something like "Login with OpenID"? I think it would be more clear for the end user.

modern/src/resources/l10n/en.json
```diff
@@ -177,6 +177,8 @@
...
+    "loginOIDCRedirect": "Authenticating with SSO...",
```
tananaev, line 181: Maybe just use a progress bar or something to indicate redirect? We try to avoid adding new string when possible.

modern/src/login/LoginPage.js
```diff
@@ -140,70 +148,85 @@ const LoginPage = () => {
...
-          disabled={!email || !password}
-        >
-          {t('loginLogin')}
-        </Button>
-        <div className={classes.extraContainer}>
+      {oidcForced ? (
+        <div className={classes.container}>
+          {t('loginOIDCRedirect')}
+        </div>
+      ) : (
```
tananaev, line 155: Maybe do a full screen loading in this case like we do in other places?

modern/src/login/LoginPage.js
```diff
@@ -60,6 +60,8 @@ const LoginPage = () => {
   const registrationEnabled = useSelector((state) => state.session.server.registration);
   const languageEnabled = useSelector((state) => !state.session.server.attributes['ui.disableLoginLanguage']);
   const emailEnabled = useSelector((state) => state.session.server.emailEnabled);
+  const oidcEnabled = useSelector((state) => state.session.server.oidcEnabled);
+  const oidcForced = useSelector((state) => state.session.server.oidcEnabled && state.session.server.oidcForce);
```
tananaev, line 64: Also need to rename these to `openId` instead of `oidc`. Related to backend comment.

modern/src/login/LoginPage.js
```diff
@@ -129,6 +131,12 @@ const LoginPage = () => {
     return () => handleLoginTokenListeners.delete(listener);
   }, []);
 
+  useEffect(() => {
+    if (oidcForced) {
+      document.location = '/api/session/openid/auth';
```
tananaev, line 136: Maybe extract this into a function so it can be reused here and in the button click callback?

modern/src/login/LoginPage.js
```diff
@@ -54,12 +54,16 @@ const LoginPage = () => {
 
   const [failed, setFailed] = useState(false);
 
+  const [loading, setLoading] = useState(false);
```
tananaev, line 57: It looks like you're not using loading state anywhere.

dan-r:
Right, those changes should match up with the backend ones now. I've done a full page loading for when force is set, and taken out the additional string - saves one more translation!

tananaev:
This looks great, thanks. I'll merge once we finish with the backed PR.

tananaev:
@dan-r looks like there a problem with push notifications and login token. Basically OIDC login bypasses all the logic we have in the web app for a normal login. Would you be able to work on a fix?

PR #1094, opened 2023-04-03
Hide password field when OpenID forced

dan-r:
As discussed on traccar/traccar PR.

Is it worth doing this for LDAP too? I didn't add it now because it doesnt look like the /api/server endpoint reports it.


tananaev:
Thanks again.

PR #1116, opened 2023-05-05
Teltonika IO30 Fault count

K-J-Dod24:
Addition of Teltonika decode FMBXXX IO30 fault count, updated the language files and usePositionAttribute.js


tananaev:
See my comment here: https://github.com/traccar/traccar/pull/5077#issuecomment-1536288466

Also, for the future, the localization is done via Transifex. In the code you should only add english version. All other files are auto-generated.

PR #1121, opened 2023-05-09
Teltonika io30 fault count

K-J-Dod24:
Appropriate code changes for web side of io30 Fault count


tananaev:
This is not something we can add here for now, but we can merge the backend part.

tananaev:
We only add the attributes that are in `Position.java`.

PR #1123, opened 2023-05-10
Teltonika io32 coolant temp/io115 engine temp

K-J-Dod24:
Coolant temperature sent from teltonika devices comes in as the io32 but is currently written in code as io115 which is engine temperature as sent from device. This is the web side of the code to correctly define and separate the io's.


tananaev:
As I already mentioned previously, you should not include any translations in the pull request. Instead the localization should be done through Transifex.

K-J-Dod24:
This isn't for language, we have included the English so that the output displayed is written out correctly, as well as this the usePositionAttributes.js, formatter.js and positionvalue.js have been edited to correctly display °C correctly and allow for the inputs to be displayed on the popup aswell as to be charted.

tananaev:
As I explained several times, you should never include non-English strings in your PRs. This is the first thing you have to revert.

Secondly, we only add attributes that in the `Position.java`. We can keep `coolantTemp`, but `engineTemp` has to go, unless you can add it to at least one other protocol.

modern/src/common/util/formatter.js
```diff
@@ -17,6 +17,8 @@ export const formatNumber = (value, precision = 1) => Number(value.toFixed(preci
 
 export const formatPercentage = (value) => `${value}%`;
 
+export const formatCelsius = (value) => `${value}°C`;
```
tananaev, line 20: I think we should rename this to `formatTemperature`.

modern/src/resources/l10n/af.json
```diff
@@ -231,6 +231,8 @@
     "linkGoogleMaps": "Google Maps",
     "linkAppleMaps": "Apple Maps",
     "linkStreetView": "Street View",
+    "positionEngineTemp": "Enjinsemperatuur",
```
tananaev, line 234: Please remove all the localization except English.

modern/src/resources/l10n/en.json
```diff
@@ -231,6 +231,7 @@
     "linkGoogleMaps": "Google Maps",
     "linkAppleMaps": "Apple Maps",
     "linkStreetView": "Street View",
+    "positionCoolantTemp": "Coolant Temperature",
```
tananaev, line 234: Can you please move this down to the list, ideally in the same order it is present in the Position.java.

tananaev (commented):
Looks like there's also a bunch of compilation errors.

tananaev:
Merged, thank you.

PR #1130, opened 2023-05-30
change roboto font package

jjreset:
PR from https://www.traccar.org/forums/topic/serviceworker-downloads-unnecessary-fonts/
Changing typeface-roboto to @fontsource/roboto
font

### DIFF
--- a/modern/src/index.js
+++ b/modern/src/index.js
@@ -1,5 +1,6 @@
 import '@fontsource/roboto/latin-400.css';
 import '@fontsource/roboto/latin-500.css';
+import '@fontsource/roboto/latin-700.css';
 import React from 'react';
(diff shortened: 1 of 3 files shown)


modern/src/index.js
```diff
@@ -1,4 +1,5 @@
-import 'typeface-roboto';
+import '@fontsource/roboto/latin-400.css';
+import '@fontsource/roboto/latin-500.css';
```
tananaev, line 2: What about 300 and 700?

jjreset: It seemed that only 400 and 500 were used in traccar, but it turns out that 400, 500, and 700 are used. However 300 doesn't seem to be used, is it?

PR #1131, opened 2023-05-30
To display litres on fuel

K-J-Dod24:
Correction to display fuel in Litres instead of no unit


K-J-Dod24:
My apologies whilst making some changes to are test environment I pulled to the wrong master branch(The main traccar instead of our master) as some of these are local changes to us or changes still in testing.

PR #1133, opened 2023-06-13
RouteReportPage.js added table pagination

kriistiina:
[fix] Browser slowdown after executing a Route report with large number of records; 
The 4000 row preview slice limit is removed.

Forum [Discussion](https://www.traccar.org/forums/topic/question-about-adding-pagination-for-report-visualization/)
Library used: [Material-UI table pagination](https://mui.com/material-ui/react-table/#sticky-header)

This is an initial version using the MUI example for table pagination. 
Additional features/behavior may be added after discussion and approval.

### DIFF
--- a/modern/package.json
+++ b/modern/package.json
@@ -66,6 +66,7 @@
     "eslint-config-airbnb": "^19.0.4",
     "eslint-plugin-import": "^2.27.5",
     "eslint-plugin-react": "^7.32.2",
-    "eslint-plugin-react-hooks": "^4.6.0"
+    "eslint-plugin-react-hooks": "^4.6.0",
+    "react-virtuoso": "^4.3.10"
   }
 }
--- a/modern/package.json
+++ b/modern/package.json
@@ -66,7 +66,6 @@
     "eslint-config-airbnb": "^19.0.4",
     "eslint-plugin-import": "^2.27.5",
     "eslint-plugin-react": "^7.32.2",
-    "eslint-plugin-react-hooks": "^4.6.0",
-    "react-virtuoso": "^4.3.10"
+    "eslint-plugin-react-hooks": "^4.6.0"
   }
 }
--- a/modern/src/reports/RouteReportPage.js
+++ b/modern/src/reports/RouteReportPage.js
@@ -176,15 +176,18 @@ const RouteReportPage = () => {
               )) : (<TableShimmer columns={columns.length + 2} startAction />)}
             </TableBody>
           </Table>
-          <TablePagination
-            rowsPerPageOptions={[100, 150, 200]}
-            component="div"
-            count={items.length}
-            rowsPerPage={rowsPerPage}
-            page={page}
-            onPageChange={handleChangePage}
-            onRowsPerPageChange={handleChangeRowsPerPage}
-          />
+          {items.length > 100 && (
+            <TablePagination
+              sx={{ position: 'sticky', bottom: 0, left: 0, backgroundColor: 'Background' }}
+              rowsPerPageOptions={[100, 150, 200]}
+              component="div"
+              count={items.length}
+              rowsPerPage={rowsPerPage}
+              page={page}
+              onPageChange={handleChangePage}
+              onRowsPerPageChange={handleChangeRowsPerPage}
+            />
+          )}
         </div>
       </div>
     </PageLayout>
--- a/modern/src/reports/RouteReportPage.js
+++ b/modern/src/reports/RouteReportPage.js
@@ -38,19 +38,10 @@ const RouteReportPage = () => {
   const [items, setItems] = useState([]);
   const [loading, setLoading] = useState(false);
   const [selectedItem, setSelectedItem] = useState(null);
-  //
+
   const [page, setPage] = useState(0);
-  const [rowsPerPage, setRowsPerPage] = useState(100);
+  const [itemsPerPage, setItemsPerPage] = useState(50);
 
-  const handleChangePage = (event, newPage) => {
-    setPage(newPage);
-  };
-
-  const handleChangeRowsPerPage = (event) => {
-    setRowsPerPage(+event.target.value);
-    setPage(0);
-  };
-  //
   const onMapPointClick = useCallback((positionId) => {
     setSelectedItem(items.find((it) => it.id === positionId));
   }, [items, setSelectedItem]);
@@ -149,7 +140,7 @@ const RouteReportPage = () => {
               </TableRow>
             </TableHead>
             <TableBody>
-              {!loading ? items.slice(page * rowsPerPage, page * rowsPerPage + rowsPerPage).map((item) => (
+              {!loading ? items.slice(page * itemsPerPage, page * itemsPerPage + itemsPerPage).map((item) => (
                 <TableRow key={item.id}>
                   <TableCell className={classes.columnAction} padding="none">
                     {selectedItem === item ? (
@@ -176,16 +167,21 @@ const RouteReportPage = () => {
               )) : (<TableShimmer columns={columns.length + 2} startAction />)}
             </TableBody>
           </Table>
-          {items.length > 100 && (
+          {items.length > 1000 && (
             <TablePagination
-              sx={{ position: 'sticky', bottom: 0, left: 0, backgroundColor: 'Background' }}
-              rowsPerPageOptions={[100, 150, 200]}
+              className={classes.pagination}
+              rowsPerPageOptions={[50, 100, 200, 500]}
               component="div"
               count={items.length}
-              rowsPerPage={rowsPerPage}
+              rowsPerPage={itemsPerPage}
               page={page}
-              onPageChange={handleChangePage}
-              onRowsPerPageChange={handleChangeRowsPerPage}
+              onPageChange={(event, page) => setPage(page)}
+              onRowsPerPageChange={
+                (event) => {
+                  setItemsPerPage(Number(event.target.value));
+                  setPage(0);
+                }
+              }
             />
           )}
         </div>
--- a/modern/src/reports/RouteReportPage.js
+++ b/modern/src/reports/RouteReportPage.js
@@ -2,13 +2,14 @@ import React, { Fragment, useCallback, useState } from 'react';
 import { useSelector } from 'react-redux';
 import { useNavigate } from 'react-router-dom';
 import {
-  IconButton, Table, TableBody, TableCell, TableHead, TableRow,
+  IconButton, Skeleton,
 } from '@mui/material';
 import GpsFixedIcon from '@mui/icons-material/GpsFixed';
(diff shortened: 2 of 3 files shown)


tananaev:
I think we discussed making it optional if there number exceeds some limit.

Can you also provide screenshots of what it looks like in desktop and mobile views.

kriistiina:
Yes, with some customization, the pagination component can be shown or hidden on a condition. 
With the changes I just added, the initial views of Route report look like this for [desktop](https://github.com/traccar/traccar-web/assets/70015610/f936b731-c5a5-4ec2-b7de-35f5da189ed2) and [mobile](https://github.com/traccar/traccar-web/assets/70015610/e73bc44f-2133-4907-8e7d-b5a1ffb9d011).

After execution of Route report with large number of records, the pagination component is shown and looks like this for [desktop](https://github.com/traccar/traccar-web/assets/70015610/a6f8a03a-4c85-40ba-b283-b852ee7cbb6b) and [mobile](https://github.com/traccar/traccar-web/assets/70015610/dee80c48-c446-47f8-ae03-02c15e74d91e).

On a following report execution, the pagination component is shown or hidden, based on the number of records in the new report. Reports with less records than the first pagination step do not contain pagination and look like this for [desktop](https://github.com/traccar/traccar-web/assets/70015610/2fdbbc19-76dc-41a5-8bb6-f0c9ceb18876) and [mobile](https://github.com/traccar/traccar-web/assets/70015610/f5c5c694-ab94-4c16-9928-c49bfe5a92de).

modern/src/reports/RouteReportPage.js
```diff
@@ -164,6 +176,18 @@ const RouteReportPage = () => {
               )) : (<TableShimmer columns={columns.length + 2} startAction />)}
             </TableBody>
           </Table>
+          {items.length > 100 && (
+            <TablePagination
+              sx={{ position: 'sticky', bottom: 0, left: 0, backgroundColor: 'Background' }}
```
tananaev, line 181: I don't like the style here. I think the pagination should be embedded in the layout instead of using "sticky". And ideally it should be in the middle of page at the bottom with proper padding. And you should use class styling.

Also, what is the "Background" color?

modern/src/reports/RouteReportPage.js
```diff
@@ -164,6 +176,18 @@ const RouteReportPage = () => {
...
+              rowsPerPageOptions={[100, 150, 200]}
```
tananaev, line 182: I think we should have something like 50, 100, 200, 500.

modern/src/reports/RouteReportPage.js
```diff
@@ -38,7 +38,19 @@ const RouteReportPage = () => {
   const [items, setItems] = useState([]);
   const [loading, setLoading] = useState(false);
   const [selectedItem, setSelectedItem] = useState(null);
+  //
```
tananaev, line 41: Remove comments

modern/src/reports/RouteReportPage.js
```diff
@@ -38,7 +38,19 @@ const RouteReportPage = () => {
...
+  const [page, setPage] = useState(0);
+  const [rowsPerPage, setRowsPerPage] = useState(100);
 
+  const handleChangePage = (event, newPage) => {
+    setPage(newPage);
+  };
```
tananaev, line 47: This can be inlined.

modern/src/reports/RouteReportPage.js
```diff
@@ -38,7 +38,19 @@ const RouteReportPage = () => {
   const [items, setItems] = useState([]);
   const [loading, setLoading] = useState(false);
   const [selectedItem, setSelectedItem] = useState(null);
+  //
+  const [page, setPage] = useState(0);
+  const [rowsPerPage, setRowsPerPage] = useState(100);
```
tananaev, line 43: I would rename to `itemsPerPage`.

modern/src/reports/RouteReportPage.js
```diff
@@ -38,7 +38,19 @@ const RouteReportPage = () => {
...
 
+  const handleChangePage = (event, newPage) => {
+    setPage(newPage);
+  };
+
+  const handleChangeRowsPerPage = (event) => {
+    setRowsPerPage(+event.target.value);
```
tananaev, line 50: Use proper explicit conversion and only if needed. I would also inline it.

tananaev (commented):
When I said exceeding some limit, I didn't really mean one page. I was thinking something like 4k limit that we had there or at least 1-2k.

modern/src/reports/RouteReportPage.js
kriistiina: I cannot seem to understand what "embedded in the layout" exactly means here. Can you please clarify?

tananaev: `position` => `static`

modern/src/reports/common/useReportStyles.js
```diff
@@ -46,4 +46,14 @@ export default makeStyles((theme) => ({
...
+  pagination: {
+    position: 'static',
+    bottom: 0,
+    '& .MuiTablePagination-spacer': {
+      display: 'none',
+    },
+    '& .MuiTablePagination-toolbar': {
+      justifyContent: 'center',
+    },
+  },
```
tananaev, line 58: What is this for?

modern/src/reports/RouteReportPage.js
```diff
@@ -137,7 +140,7 @@ const RouteReportPage = () => {
               </TableRow>
             </TableHead>
             <TableBody>
-              {!loading ? items.slice(0, 4000).map((item) => (
+              {!loading ? items.slice(page * itemsPerPage, page * itemsPerPage + itemsPerPage).map((item) => (
```
tananaev, line 143: Have you tested that it shows 1000 if less than 1000?

LiGuru: What you think, is it tested? 

`              {!loading ? items.slice(page * itemsPerPage, items.length > 1000 ? page * itemsPerPage + itemsPerPage : items.length).map((item, id) => (
`
😁

tananaev: I think it's too complicated and not very readable. Also we don't need to do `slice` if the array fits.

mohammadraufzahed: Hey there,

I wanted to suggest adding virtualization, like react-window, to improve the table's performance. It can optimize rendering, especially for large datasets, by only rendering visible rows. This can enhance the user experience by reducing memory usage and providing smoother scrolling.

Consider exploring the integration of react-window for better efficiency. It's a popular library for efficient virtualization in React applications.

Thanks for considering this suggestion!

Best regards,
Mohammad Raufzahed

kriistiina:
@tananaev 
Sorry about the delay, I was on a business trip last few days. I fixed the issue regarding the visualization of all items when their count is less than 1000 and made the .slice conditional only if the items count exceeds 1000.

tananaev:
What do you think about the comment of using the "react-window" or something similar? The same way we do on the main screen for devices.

kriistiina:
I used [react-virtuoso](https://github.com/petyosi/react-virtuoso) to load all report items in a Table component, based on this [example](https://mui.com/material-ui/react-table/#virtualized-table). This way, the behavior stays close to the one before with infinite scroll, but without the browser slowdown.

modern/package.json
```diff
@@ -66,6 +66,7 @@
     "eslint-config-airbnb": "^19.0.4",
     "eslint-plugin-import": "^2.27.5",
     "eslint-plugin-react": "^7.32.2",
-    "eslint-plugin-react-hooks": "^4.6.0"
+    "eslint-plugin-react-hooks": "^4.6.0",
+    "react-virtuoso": "^4.3.10"
```
tananaev, line 70: Why do we need this here in dev dependencies?

modern/src/reports/common/useReportStyles.js
```diff
@@ -17,6 +17,7 @@ export default makeStyles((theme) => ({
     position: 'sticky',
     left: 0,
     display: 'flex',
+    height: '100%',
```
tananaev, line 20: Please explain this.

modern/src/reports/RouteReportPage.js
```diff
@@ -128,43 +184,15 @@ const RouteReportPage = () => {
...
-                      />
-                    </TableCell>
-                  ))}
-                </TableRow>
-              )) : (<TableShimmer columns={columns.length + 2} startAction />)}
-            </TableBody>
-          </Table>
         </div>
+        {!loading && (
+          <TableVirtuoso
```
tananaev, line 189: We already have `react-window`. Why are we adding another library that basically does the same thing?

tananaev: Obviously I didn't mean that it's ok to just add more libraries that do the same thing. If you want to replace the virtualization library, you have to provide justification why the old one is bad and the new one is better. Then migrate all existing use cases to the new library.

LiGuru: Ok, let's clarify.
Kristina should implement rendering with react-window.

Kristina should not use additional libraries without clarification by you.

What about adding filters and sorting?

kriistiina:
I managed to visualize the Route report by using the [FixedSizeGrid](https://react-window.vercel.app/#/api/FixedSizeGrid) component from [react-window](https://react-window.vercel.app/#/examples/grid/fixed-size) and removed the additional [react-virtuoso](https://virtuoso.dev/hello-table/).

modern/src/reports/RouteReportPage.js
```diff
@@ -128,42 +135,94 @@ const RouteReportPage = () => {
...
+            <AutoSizer>
+              {({ height, width }) => (
+                <FixedSizeGrid
+                  height={height}
+                  width={width}
+                  columnCount={columns.length + 2}
+                  columnWidth={(columns.length + 2) * (width * (phone ? 0.26 : desktop ? 0.1 : 0.18)) >= width ?
+                    width * (phone ? 0.26 : desktop ? 0.1 : 0.18) : width / (columns.length + 2)}
+                  rowCount={items.length > 0 ? items.length : 5}
+                  rowHeight={52}
```
tananaev, line 148: This does not look right at all. Why are there a million hardcoded numbers and some crazy calculations?

modern/src/reports/RouteReportPage.js
```diff
@@ -128,42 +135,94 @@ const RouteReportPage = () => {
...
+                  overscanRowCount={20}
+                >
+                  {({ columnIndex, rowIndex, style }) => {
+                    const item = items[rowIndex];
+                    const columnKey = columns[columnIndex - 2];
+                    return (
```
tananaev, line 154: The content also doesn't look good. Very hard to understand what's going on.

PR #1135, opened 2023-06-16
Sync old Position Attribute formatters to Modern

jinzo:
Hi,

this synces how the PositionAttribute formatting is done on the Modern interface, so that it matches the old interface. 
I extended the attributes which get formatted with speed and distance formatter, added voltage, driver unique id, consumption and volume.

### DIFF
--- a/modern/src/common/util/formatter.js
+++ b/modern/src/common/util/formatter.js
@@ -21,6 +21,8 @@ export const formatTemperature = (value) => `${value}°C`;
 
 export const formatVoltage = (value, t) => `${value} ${t('sharedVoltAbbreviation')}`;
 
+export const formatConsumption = (value, t) => `${value} ${t('sharedLiterPerHourAbbreviation')}`;
+
 export const formatTime = (value, format, hours12) => {
   if (value) {
     const m = moment(value);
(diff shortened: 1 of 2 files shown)


modern/src/common/util/formatter.js
```diff
@@ -64,6 +69,16 @@ export const formatNumericHours = (value, t) => {
   return `${hours} ${t('sharedHourAbbreviation')} ${minutes} ${t('sharedMinuteAbbreviation')}`;
 };
 
+export const formatDriverUniqueId = (value) => {
```
tananaev, line 72: I don't know about this one. We only do light formatting in these functions. A lookup like this should probably be in a different place.

jinzo: Was thinking the same initially, but then saw that it's implemented like that in the old interface and just decided to duplicate it. But I'm open to discuss how we can better represent driver(s)/mapping and just remove it from this PR/commits.

jinzo: Will tackle this at some later time then, removed it for now. Thanks for the guidance.

tananaev:
Merged, thanks.

PR #1144, opened 2023-06-29
Update fa.json

mehdi-song:


tananaev:
The localization should be done through Transifex:

https://explore.transifex.com/traccar/traccar/

PR #1155, opened 2023-07-29
StatusCard on mobile screens hides the device on the map

jcardus:
before:
traccar fleetmap io_(iPhone SE) (1)

after:
localhost_3000_(iPhone SE)

### DIFF
--- a/modern/src/common/components/StatusCard.js
+++ b/modern/src/common/components/StatusCard.js
@@ -57,7 +57,7 @@ const useStyles = makeStyles((theme) => ({
   content: {
     paddingTop: theme.spacing(1),
     paddingBottom: theme.spacing(1),
-    'max-height': '40vh',
+    maxHeight: theme.dimensions.cardContentMaxHeight,
     overflow: 'scroll',
   },
   negative: {
--- a/modern/src/common/components/StatusCard.js
+++ b/modern/src/common/components/StatusCard.js
@@ -58,7 +58,7 @@ const useStyles = makeStyles((theme) => ({
     paddingTop: theme.spacing(1),
     paddingBottom: theme.spacing(1),
     'max-height': '40vh',
-    overflow: 'hidden',
+    overflow: 'scroll',
   },
   negative: {
     color: theme.palette.colors.negative,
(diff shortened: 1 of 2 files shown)


modern/src/common/components/StatusCard.js
```diff
@@ -57,6 +57,8 @@ const useStyles = makeStyles((theme) => ({
   content: {
     paddingTop: theme.spacing(1),
     paddingBottom: theme.spacing(1),
+    'max-height': '40vh',
+    overflow: 'hidden',
```
tananaev, line 61: Why are we hiding overflow?

jcardus: because it overlaps with the actions

jcardus: without overflow: hidden

localhost_3000_(iPhone SE) (3)

tananaev: Screenshot please. And also wouldn't it mean that it's not the content is not accessible?

jcardus: yes, the content overflow content will not accesible

tananaev: That's not an acceptable solution then. It should become scrollable if you want to limit the height.

jcardus: Changed to overflow: scroll
localhost_3000_(iPhone SE) (4)

modern/src/common/components/StatusCard.js
```diff
@@ -57,6 +57,8 @@ const useStyles = makeStyles((theme) => ({
   content: {
     paddingTop: theme.spacing(1),
     paddingBottom: theme.spacing(1),
+    'max-height': '40vh',
```
tananaev, line 60: You have to use camelcase name here. And also the constant should be in the same place where we put other similar constants. It's part of the theme.

jcardus:
please check

tananaev:
Merged, thanks.

tananaev:
@jcardus is looks like the card is broken by this change. Tested on Chrome on Mac. Can you please fix.

Screenshot 2023-08-19 at 2 05 04 PM

jcardus:
Sorry @tananaev, I just saw this now. Is this still an issue?

tananaev:
I forgot. Maybe I already fixed it?

jcardus:
I can't reproduce it.

Issue #1156, opened 2023-07-29
Selected device inside cluster


jcardus:
When we select a device inside cluster it's not shown.

How should we fix this?

https://github.com/traccar/traccar-web/assets/31902485/84c6a447-6c88-4d40-878e-874d77253955

tananaev:
We probably should not cluster the selected device.

jcardus:
ok, I'll make a PR

PR #1157, opened 2023-07-29
distinct map source for selected device

jcardus:
https://github.com/traccar/traccar-web/issues/1156

I tried to use the same source and filter the selected device but it doesn't work. When the source is clustered the feature is never rendered when it's inside a cluster. So I ended up making two sources.

https://github.com/traccar/traccar-web/assets/31902485/5ad3b044-b2cb-477e-a2b1-345eab636202

### DIFF
--- a/modern/src/map/MapPositions.js
+++ b/modern/src/map/MapPositions.js
@@ -101,11 +101,11 @@ const MapPositions = ({ positions, onClick, showStatus, selectedPosition, titleF
         features: [],
       },
     });
-    [id, selected].forEach((id) => {
+    [id, selected].forEach((source) => {
       map.addLayer({
-        id,
+        id: source,
         type: 'symbol',
-        source: id,
+        source,
         filter: ['!has', 'point_count'],
         layout: {
           'icon-image': '{category}-{color}',
@@ -124,9 +124,9 @@ const MapPositions = ({ positions, onClick, showStatus, selectedPosition, titleF
         },
       });
       map.addLayer({
-        id: `direction-${id}`,
+        id: `direction-${source}`,
         type: 'symbol',
-        source: id,
+        source,
         filter: [
           'all',
           ['!has', 'point_count'],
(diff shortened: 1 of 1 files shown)


modern/src/map/MapPositions.js
```diff
@@ -11,13 +11,14 @@ import { useAttributePreference, usePreference } from '../common/util/preference
 const MapPositions = ({ positions, onClick, showStatus, selectedPosition, titleField }) => {
   const id = useId();
   const clusters = `${id}-clusters`;
-  const direction = `${id}-direction`;
```
tananaev, line 14: Why are we removing this?

modern/src/map/MapPositions.js
```diff
@@ -93,27 +94,53 @@ const MapPositions = ({ positions, onClick, showStatus, selectedPosition, titleF
...
+          'text-font': findFonts(map),
+          'text-size': 12,
+        },
+        paint: {
+          'text-halo-color': 'white',
+          'text-halo-width': 1,
+        },
+      });
+      map.addLayer({
+        id: `direction-${id}`,
```
jcardus, line 127: we need two direction layers, I create the id here

modern/src/map/MapPositions.js
```diff
@@ -162,34 +172,41 @@ const MapPositions = ({ positions, onClick, showStatus, selectedPosition, titleF
...
         map.removeLayer(clusters);
       }
-      if (map.getLayer(direction)) {
-        map.removeLayer(direction);
-      }
-      if (map.getSource(id)) {
-        map.removeSource(id);
-      }
+
+      [id, selected].forEach((id) => {
```
tananaev, line 179: Actually the `id` in the lambda is very confusing. It clashes with the component id.

tananaev:
Merged, thank you.

tananaev:
Looks like this change broke click callbacks for selected devices. Fixed here: https://github.com/traccar/traccar-web/commit/f87566f0bdc7837cf6ac33eadecced1160d019bf.

Also for context:

https://www.traccar.org/forums/topic/route-replay-from-device-popup-window/

PR #1158, opened 2023-07-30
Route replay with events

jcardus:
I'm checking if I'm going in the right direction.

The idea is to show events on the route replay:

https://github.com/traccar/traccar-web/assets/31902485/950340fb-f99a-4cfd-be78-a5037b695f15


tananaev:
The idea is to use combined report to show events on a route.

jcardus:
ok, I'll check it out, thanks.

PR #1162, opened 2023-08-16
Include tooltips for menu actions

jguthrie100:
I've added tooltips to some of the action buttons in the various settings page, which were not always clear about the expected action

[Screencast from 2023-08-16 19:07:38.webm](https://github.com/traccar/traccar-web/assets/4670502/4358a78d-6942-4e24-8c15-2a81e07b57cf)


modern/src/settings/GroupsPage.js
```diff
@@ -45,13 +45,19 @@ const GroupsPage = () => {
   const actionCommand = {
     key: 'command',
     title: t('deviceCommand'),
+    tooltip: t('commandSend')
+      .concat(' ')
+      .concat(t('reportGroup'))
+      .concat(' ')
```
tananaev, line 51: Concatenating strings is not the right solution.

tananaev (commented):
The comment applies to all of the concats.

modern/src/settings/GroupsPage.js
jguthrie100: How should they be resolved?

Adding new strings to the translation file? There are a lot of dynamically strings that will have to be added..
"Remove Group", "Remove Driver", "Remove Geofence" etc

tananaev: Well, I don't think you need to duplicate information. If it's a command button, it can just say "command" or "send command". We're already on the group screen in this case, so you don't need to add that information.

jguthrie100: I've updated to remove the type of item from the tooltips (i.e. "Remove" instead of "Remove Group").

I've left in a couple of more specific custom translations

modern/src/common/components/LocalizationProvider.js
```diff
@@ -178,7 +178,7 @@ export const useLocalization = () => useContext(LocalizationContext);
 export const useTranslation = () => {
   const context = useContext(LocalizationContext);
   const { data } = context.languages[context.language];
-  return useMemo(() => (key) => data[key], [data]);
+  return useMemo(() => (key) => (data[key] ? data[key] : context.languages.en.data[key]), [data]);
```
tananaev, line 181: Why do you need the fallback here?

modern/src/other/GeofencesPage.js
```diff
@@ -117,7 +118,9 @@ const GeofencesPage = () => {
             <label htmlFor="upload-gpx">
               <input accept=".gpx" id="upload-gpx" type="file" className={classes.fileInput} onChange={handleFile} />
               <IconButton edge="end" component="span" onClick={() => {}}>
-                <UploadFileIcon />
+                <Tooltip title={t('sharedUploadGeofence')} enterDelay={500} enterNextDelay={500}>
```
tananaev, line 121: Why do you have hardcoded delays everywhere?

modern/src/settings/GroupsPage.js
```diff
@@ -45,6 +45,7 @@ const GroupsPage = () => {
   const actionCommand = {
     key: 'command',
     title: t('deviceCommand'),
+    tooltip: t('commandSendTitle'),
```
tananaev, line 48: Why can't we just use title? Then we don't need extra strings either.

modern/src/common/components/LocalizationProvider.js
jguthrie100: So that the frontend doesn't fail when run in another language.

It was more relevant when the translations were being concat-ed, but even now, unless I add translations for every language for "Upload geofence" for example, then nothing will be displayed.

The above change means that the English option will always be displayed, even if there is no local translation.

tananaev: I guess it makes sense, but you should do the same thing as for selected language. Pass it as a variable.

modern/src/other/GeofencesPage.js
jguthrie100: I can remove them.. the default delay is pretty short, and I figured that expert users might get sick of tooltips constantly popping up as they move their mouse around the screen.

Whereas beginners and people who don't know what a button does, are more likely to hover over a button for longer, when the tooltip will actually help them.

modern/src/settings/GroupsPage.js
jguthrie100: I wanted to use the specific "Send Command" instruction, but I could just use the title and leave it as "Command"

modern/src/common/components/LocalizationProvider.js
jguthrie100: Please could you elaborate? I don't understand what you mean..

modern/src/other/GeofencesPage.js
tananaev: If you think it's too long, we should change it globally, not per call.

modern/src/settings/GroupsPage.js
tananaev: Let's leave it as "command".

modern/src/common/components/LocalizationProvider.js
tananaev: It needs to be passed to `useMemo`.

modern/.env
```diff
@@ -1,3 +1,4 @@
 ESLINT_NO_DEV_ERRORS=false
 REACT_APP_URL_NAME='localhost:8082'
 REACT_APP_VERSION=$npm_package_version
+REACT_APP_TOOLTIP_DELAY=750
```
tananaev, line 4: No no no. That's not the right way. You have to define override the default in the theme.

modern/src/common/components/LocalizationProvider.js
jguthrie100: I still don't understand what you mean here. Maybe I should just remove it - it only affects one non-breaking translation

tananaev: Do you see how the selected language is used? You have to do absolutely the same thing with english.

modern/src/common/components/DelayedTooltip.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import Tooltip from '@mui/material/Tooltip';
+
+const DelayedTooltip = ({
```
tananaev, line 4: This is not what I asked. I'm talking about something like this:

https://github.com/traccar/traccar-web/blob/master/modern/src/common/theme/components.js

modern/.env
```diff
@@ -1,3 +1,3 @@
 ESLINT_NO_DEV_ERRORS=false
 REACT_APP_URL_NAME='localhost:8082'
-REACT_APP_VERSION=$npm_package_version
+REACT_APP_VERSION=$npm_package_version
```
tananaev, line 3: Please revert this change

modern/src/resources/l10n/en.json
```diff
@@ -35,6 +35,7 @@
     "sharedGeofence": "Geofence",
     "sharedGeofences": "Geofences",
     "sharedCreateGeofence": "Create Geofence",
+    "sharedUploadGeofence": "Upload Geofence",
```
tananaev, line 38: By the way, I would recommend making this more generic. Maybe just "upload". It might be useful in other places in the future. And the button is on the geofences, page, so it's pretty clear already that it's for geofences.

modern/src/common/components/LocalizationProvider.js
jguthrie100: I still don't understand what you mean tbh, so I've just removed it. Its only relevant for the "Upload" tooltip which can just remain blank for non-English languages

tananaev:
Merged, thank you.

PR #1164, opened 2023-08-19
Add country flags to language select

jguthrie100:
I thought it might be nice to include country flags within the language select dropdown..

image


tananaev:
1. How do we deal with languages that are used in multiple countries?
2. Can we use Unicode symbols instead of including another library?

jguthrie100:
Sure, I can change it to use emojis/unicode.

For multiple countries, I just used the main country (apart from English, where I used the UK, but could make it USA)

I was thinking of even listing multiple flags for certain languages, but it might look a bit cluttered!

tananaev:
Yeah, let's do emojis and set the English to the USA flag as the more common.

jguthrie100:
Apparently Windows doesn't have support for flag emojis? I don't have windows so I can't check..

https://answers.microsoft.com/en-us/windows/forum/all/flag-emoji/85b163bc-786a-4918-9042-763ccf4b6c05

tananaev:
It's not working on windows even with a custom font, like roboto?

jguthrie100:
My understanding that Microsoft have left all unicode flags out of Windows in order to avoid issues around supporting various countries (i.e. whether to include Palestine/Israel flags etc)

tananaev:
Can you update the pr and I'll try.

modern/src/common/components/LocalizationProvider.js
```diff
@@ -64,62 +64,62 @@ import zhTW from '../../resources/l10n/zh_TW.json';
...
-  sv: { data: sv, name: 'Svenska' },
-  ta: { data: ta, name: 'தமிழ்' },
-  th: { data: th, name: 'ไทย' },
-  tr: { data: tr, name: 'Türkçe' },
-  uk: { data: uk, name: 'Українська' },
-  uz: { data: uz, name: 'Oʻzbekcha' },
-  vi: { data: vi, name: 'Tiếng Việt' },
-  zh: { data: zh, name: '中文' },
-  zhTW: { data: zhTW, name: '中文 (Taiwan)' },
+  af: { data: af, countryCode: 'ZA', name: 'Afrikaans' },
```
tananaev: Let's rename `countryCode` to just `country`.

modern/src/common/components/LocalizationProvider.jsx
```diff
@@ -64,62 +64,62 @@ import zhTW from '../../resources/l10n/zh_TW.json';
...
+  km: { data: km, countryCode: 'KH', name: 'ភាសាខ្មែរ' },
+  lo: { data: lo, countryCode: 'LA', name: 'ລາວ' },
+  lt: { data: lt, countryCode: 'LT', name: 'Lietuvių' },
+  lv: { data: lv, countryCode: 'LV', name: 'Latviešu' },
+  ml: { data: ml, countryCode: 'IN', name: 'മലയാളം' },
+  mn: { data: mn, countryCode: 'MN', name: 'Монгол хэл' },
+  ms: { data: ms, countryCode: 'MY', name: 'بهاس ملايو' },
+  ne: { data: ne, countryCode: 'NP', name: 'नेपाली' },
+  nl: { data: nl, countryCode: 'NL', name: 'Nederlands' },
+  nb: { data: nb, countryCode: 'NO', name: 'Norsk bokmål' },
```
tananaev, line 102: Please keep the order as before.

modern/src/common/components/LocalizationProvider.jsx
```diff
@@ -186,3 +186,11 @@ export const useTranslationKeys = (predicate) => {
   const { data } = context.languages[context.language];
   return Object.keys(data).filter(predicate);
 };
+
+export const getFlagEmoji = (countryCode) => {
+  const codePoints = countryCode
+    .toUpperCase()
```
tananaev, line 192: I think this is not needed.

jguthrie100:
Have you tested it on Windows? I just heard back from someone that it only shows the unicode letters..

tananaev:
Doesn't work on Windows, but I've added this check: https://github.com/traccar/traccar-web/commit/c36d1bed4a8781b2307fe2bfcdd1c6d4ddd00d64.

jguthrie100:
Is it not worth just including the tiny library so the majority Windows users can see flags too?

tananaev:
OK, let's do it.

PR #1165, opened 2023-08-19
Use country flags in language select

jguthrie100:


tananaev:
This version doesn't work on Windows either. You need to use SVG: https://github.com/traccar/traccar-web/commit/a5fbacbe37184c845731b6a6d69eb4dc4ac1b2e9.

PR #1166, opened 2023-08-20
Use SVG flag icons only on Windows

jguthrie100:
Made a small change using @tananaev old code to determine whether to use emojis or svg icons


jguthrie100:
Maybe a good change? Feel free to close without merging if its overkill!

tananaev:
I think it's fine to use SVGs for all platforms. I actually like how they look more. And also more consistency.

PR #1178, opened 2023-09-18
Update StatusCard.jsx - translate key name for custom attributes

SINECANET:


tananaev:
This change doesn't make sense to me.

PR #1179, opened 2023-09-20
OBDSpeed expected as km/h not knots.

geduxas:
OBD Speed value is saved as km/h, but front side expect it as knots, and convert's it.

### DIFF
--- a/modern/src/common/util/converter.js
+++ b/modern/src/common/util/converter.js
@@ -10,18 +10,6 @@ const speedConverter = (unit) => {
   }
 };
 
-const speedConverterKph = (unit) => {
-  switch (unit) {
-    case 'kmh':
-      return 1;
-    case 'mph':
-      return 0.62137;
-    case 'kn':
-    default:
-      return 0.539957;
-  }
-};
-
 export const speedUnitString = (unit, t) => {
   switch (unit) {
     case 'kmh':
@@ -35,8 +23,6 @@ export const speedUnitString = (unit, t) => {
 };
 
 export const speedFromKnots = (value, unit) => value * speedConverter(unit);
-
-export const speedFromKph = (value, unit) => value * speedConverterKph(unit);
 
 export const speedToKnots = (value, unit) => value / speedConverter(unit);
 
--- a/modern/src/common/util/converter.js
+++ b/modern/src/common/util/converter.js
@@ -10,6 +10,18 @@ const speedConverter = (unit) => {
   }
 };
 
+const speedConverterKph = (unit) => {
+  switch (unit) {
+    case 'kmh':
+      return 1;
+    case 'mph':
+      return 0.62137;
+    case 'kn':
+    default:
+      return 0.539957;
+  }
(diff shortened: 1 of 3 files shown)


modern/src/common/util/converter.js
```diff
@@ -10,6 +10,18 @@ const speedConverter = (unit) => {
   }
 };
 
+const speedConverterKph = (unit) => {
```
tananaev, line 13: I don't think we need a whole new conversion function. You should be able to just do `value / speedConverter('kmh').

modern/src/common/util/formatter.js
```diff
@@ -64,6 +65,8 @@ export const formatAltitude = (value, unit, t) => `${altitudeFromMeters(value, u
 
 export const formatSpeed = (value, unit, t) => `${speedFromKnots(value, unit).toFixed(2)} ${speedUnitString(unit, t)}`;
 
+export const formatSpeedKph = (value, unit, t) => `${speedFromKph(value, unit).toFixed(2)} ${speedUnitString(unit, t)}`;
```
tananaev, line 68: Same here. You should just convert the value and call `formatSpeed` instead of duplicating the code.

geduxas: Truly I don't like that ide either.. all speed values should be treated same, don't think it shoud be different speed types at all.. and why speed value was chosen to be in knots? And only obd speed in km/h? I would like to stickto idea that all values are stored in same format.. also in traccar repository i find couple obd_speed storied as knot... other's stored asis..

tananaev: I think you misunderstood something. I'm not suggesting that display units are limited. I'm telling you that you can reuse the existing code instead of duplicating it.

geduxas: @tananaev will it be better now?

tananaev: No. You removed what you shouldn't remove and didn't remove what you should.

geduxas: Sorry i am giving up :D something i didn't understand when.. 
So i just need to convert saved kmh value to knots, and then use same function to convert back to kmh?

tananaev: You just need to convert value to knots and pass it to the existing function.

geduxas: Yes, but it will add error convert and round operations.. saved value of 100km/h => (56.996) 56.0 kn=>  (100.008) 100.01km/h 

So why's my solution is so bad?

tananaev: The rounding shouldn't be a problem. We pass the value with very high accuracy, so rounded value should always be the same as the original.

Your solution is bad because it adds code that we don't need.

geduxas: Ok @tananaev made changes as you suggested, if it's ok please confirm PR.

tananaev:
Merged, thank you. Looks much better.

Issue #1169, opened 2023-08-28
Feature Request: PWA Install button/prompt


jinzo:
Hi,

I know that there are wrappers for iOS and Android, but as I'm trying to avoid deploying apps to the both app stores - I was thinking if we would add a menu item or a prompt that would allow the users to install the (new) PWA. 
I'm willing to implement this PR, but I'm not 100% sure where to put the button or if we would raither go with a floating prompt? I personally prefer a menu button, and in the new interface the only feasible position would be under Account alongside Log Out. It does not fit 100%, but it's the best position I can come up on? Any other ideas?

Ofcourse, the button would only show if it's possible to install the PWA.

See as a sample: [https://dev.to/woile/simplest-react-hook-component-for-pwa-install-button-2die](https://dev.to/woile/simplest-react-hook-component-for-pwa-install-button-2die)

tananaev:
I guess if the change is small, I don't mind adding it, but only with a server settings flag. I don't want it to be the default.

As for the location, maybe we can put it in the settings screen somewhere?

jinzo:
Yes the change is small, and I can put it in the settings screen yes. 
I started prototyping this and from what I see this would be better suited in .env setting, not server side config? As I did not find any similar server side setting that affects only the frontend/web display? Or did I miss any?
On the other hand, I see that .env nowadays doesen't actually contain any settings (it did before afaik? or are they just not in the default .env?)

tananaev:
It has to be user configurable, not developer configurable. You should add a server attribute for it.

jinzo:
After some proof of concept work and tests I noticed that unfortunately this is only kinda-sorta supported on Chrome. See https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeinstallprompt_event#browser_compatibility

So it's not really worth it currently, I will keep an eye on this and reopen if if becomes feasible.

PR #1188, opened 2023-11-05
PWA regularly check for service worker updates and automaticially reload the page

jinzo:
Hi,

I was working on the https://github.com/traccar/traccar-web/issues/1169 issue and when deploying to staging I noticed that my PWA (on phone or open in the browser) does not actually update automatically like I would presume with the ```registerType: 'autoUpdate'``` config option set. 
After a bit of digging I actually found two things missing:
1. The ServiceWorker itself did not check for an update when running. So I checked the documentation of VitePWA and adapted the code found there: https://vite-pwa-org.netlify.app/guide/periodic-sw-updates
2. The actual webpage reload (after the ServiceWorker autoUpdates with step 1.) was not applied, therefore I set the immediate parameter to true as seen in the warning here: https://vite-pwa-org.netlify.app/guide/auto-update.html#automatic-reload

I tested this and now it behaves like I would expect it to behave. The Traccar PWA now checks every hour, if there is a new version and if found updates the ServiceWorker and reloads the webpage automaticially. 

On a somewhat related note, just to be sure, so you do prefer automatic reload instead of the update-popup-and-user-clicks-for-reload?


tananaev:
I thought they have a standard way to detect it automatically? And have a prompt option. How is `registerType: 'autoUpdate'` supposed to work?

jinzo:
I was under that impression too, but apparently not - it did update itself/check (with autoUpdate option) on a new page load (either open a new tab or reload the current page). And that's how I understand this option alone works (it silently updates the ServiceWorker, not the webapp) on new page load (as this is actually not guaranteed with ServiceWorkers, there could be old ServiceWorkers open and somewhere in another tab and they would not get updated if a new tab loads a new version of a ServiceWorker). More info on updating ServiceWorkers: https://web.dev/articles/service-worker-lifecycle#updates
Yes there is a prompt option, if you change registerType to 'prompt' instead of 'autoUpdate' (or omit it, 'prompt' is default) then it won't autoUpdate the ServiceWorker, but it will do it once the user clicks on the button (so it will update the ServiceWorker AND the pwa/web). 
But you still have to check every X ms, if you don't want to wait for (re)load of the webpage (which I'm not even sure how the customer can do if it 'installs' the PWA on their phone for example). 
I personally would prefer the prompt option, so the PWA would just check every X time if there is an update, and use the VitePWA prompt to let the user update if there is one.

tananaev:
It seems like the browser should automatically check for updates:

https://web.dev/articles/service-worker-lifecycle#updates

Maybe I misunderstand something?

jinzo:
The browser is checking for an update in this cases:

- A navigation to an in-scope page.
- A functional events such as push and sync, unless there's been an update check within the previous 24 hours.
- Calling .register() only if the service worker URL has changed. However, you should [avoid changing the worker URL](https://web.dev/articles/service-worker-lifecycle#avoid-url-change).

The last two are not applicable, as we are not pushing or syncing anything to the ServiceWorker and we are not chaging the ServiceWorker URL.
Which means that only the first check is applicable, that happens when a user opens the webpage (either a new load, or an reload of the existing page). So in Traccars React App this means, that if a user has just one tab with it open, and uses it normally (and doesen't close the tab itself) he won't get the updated serviceworker/code until he opens a new one (or reloads the webpage). As a React App doesen't really navigate to a new page and therefore trigger a update. And in the case of a "Installed" PWA on a phone, this probably means force closing the "app" and opening it again.

tananaev:
Why do we need the `immediate: true`? Would having prompt, as we discussed, fix it?

jinzo:
Because of the warning here: https://vite-pwa-org.netlify.app/guide/auto-update.html#automatic-reload
Quote from the yellow WARNING "Automatic reload is not automatic page reload, you will need to use the following code in your application entry point if you want automatic page reload:" My note: Automatic reload == 'autoUpdate'

Yes, having a prompt removes the need for ```immediate: true```

tananaev:
- OK, let's change to prompt
- Let's make auto-update configurable
- Let's keep this copied code in a separate file and provide a link to the original

jinzo:
Ok, just to be clear, so user configurable as a server setting? And do we make it just on/off or can the update check interval be configured too (eg. then I would still go with one setting, but if it's 0 then don't check, if it's more than 0 then use that as an update check interval)?

tananaev:
We can make the interval configurable, if that's easy to implement. And zero means no updates. I think we should include it in the server attributes.

tananaev:
I think if it works well, let's make it as a default, so some reasonable default value.

jinzo:
So this is just a work in progress for now - but have to finish for today and comited to get some preliminary feedback.

I moved everything to a React component and adapted it to use plain Snackbar like other similar functionality in Traccar. While tweaking this I tested it plenty, and it works very well. What remains is figuring out why "useAttributePreference" is erroring out on me and do basic cleanup around that. I'm thinking 1h as the default value (as this is also the 'default' in Vite PWA documentation).

P.S: There is a big ol red warning that switching between 'autoUpdate' and 'prompt' while in production can be a pain, but from what I can see 'autoUpdate' did not make it into 5.9 so this won't be a problem: https://vite-pwa-org.netlify.app/guide/auto-update.html

jinzo:
After stepping back from the monitor for a few minutes it dawned on me, that as the system is not initialized, I cannot access the user attributes. Moved it to a better location and it is ready for a review.

modern/src/ReloadPrompt.tsx
```diff
@@ -0,0 +1,56 @@
+import { Snackbar, Button } from '@mui/material';
+import React from 'react'
+import { useTranslation } from './common/components/LocalizationProvider';
+import { useAttributePreference } from './common/util/preferences';
+import { useRegisterSW } from 'virtual:pwa-register/react'
+
+// Based on https://vite-pwa-org.netlify.app/frameworks/react.html
+function ReloadPrompt() {
```
tananaev, line 8: We need a better name because it's not just prompt. It actually checks.

modern/src/ReloadPrompt.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const t = useTranslation();
+
+  const serviceWorkerUpdateInterval = useAttributePreference('serviceWorkerUpdateInterval', 3600000);
+
+  const {
+    offlineReady: [offlineReady, setOfflineReady],
```
tananaev, line 14: Why do we need this?

modern/src/ReloadPrompt.tsx
```diff
@@ -0,0 +1,56 @@
...
+    needRefresh: [needRefresh, setNeedRefresh],
```
tananaev, line 15: And `setNeedRefresh`?

modern/src/ReloadPrompt.tsx
```diff
@@ -0,0 +1,56 @@
...
+import React from 'react'
+import { useTranslation } from './common/components/LocalizationProvider';
+import { useAttributePreference } from './common/util/preferences';
+import { useRegisterSW } from 'virtual:pwa-register/react'
+
+// Based on https://vite-pwa-org.netlify.app/frameworks/react.html
+function ReloadPrompt() {
+  const t = useTranslation();
+
+  const serviceWorkerUpdateInterval = useAttributePreference('serviceWorkerUpdateInterval', 3600000);
```
tananaev, line 11: This should be server only. Also we usually have values in seconds or milliseconds. This seems to be in hours?

modern/src/ReloadPrompt.tsx
```diff
@@ -0,0 +1,56 @@
...
+          cache: 'no-store',
+          headers: {
+            'cache': 'no-store',
+            'cache-control': 'no-cache',
+          },
+        });
+  
+        if (resp?.status === 200) {
+          await r.update();
+        }
```
tananaev, line 38: Can we give proper names to variables please.

modern/src/ReloadPrompt.tsx
```diff
@@ -0,0 +1,56 @@
...
+      action={(
+        <Button onClick={() => updateServiceWorker(true)}>
+          {t('sharedReload')}
+        </Button>
+      )}
+      />
+  );
+}
+
+export default ReloadPrompt;
```
tananaev, line 56: Missing new line

modern/src/resources/l10n/en.json
```diff
@@ -99,6 +99,8 @@
     "sharedImport": "Import",
     "sharedColumns": "Columns",
     "sharedDropzoneText": "Drag and drop a file here or click",
+    "sharedReload": "Reload",
+    "updateAvailable": "New update available, click on the reload button to update.",
```
tananaev, line 103: This should have a prefix. Maybe it should all be under "settings" prefix?

tananaev: Also I think the wording can be simplified. If there's a clear reload button, there's no need to tell that you have to reload.

tananaev: Please include a screenshot.

modern/src/ReloadPrompt.tsx
jinzo: UpdateCheckPrompt ? or even UpdateCheckAndPrompt?

modern/src/ReloadPrompt.tsx
jinzo: Sorry this remained from the original snippet, will remove it.

modern/src/ReloadPrompt.tsx
jinzo: Sorry this remained from the original snippet, will remove it.

modern/src/ReloadPrompt.tsx
jinzo: I'm still not 100% sure I understand different settings and possibilities in Traccar. Are we talking about here checking only ```state.session.server.attributes.serviceWorkerUpdateInterval``` right? or are we talking about traccar.conf on the server side?
Also value is in milliseconds 60 * 60 * 1000 = 3600000

modern/src/ReloadPrompt.tsx
jinzo: I mostly left this part with minimum changes from the original snippet, but will rename.

modern/src/resources/l10n/en.json
jinzo: Simplified the wording, see attached screenshot of a mobile view.
Screenshot from 2023-11-06 09-47-12

modern/src/ReloadPrompt.tsx
jinzo: Went with UpdateCheckPrompt for now, after looking around UpdateController could be used too. What do you think?

modern/src/resources/l10n/en.json
jinzo: There is also a possibility of outlined button
Screenshot from 2023-11-06 13-47-12

modern/src/UpdateCheckPrompt.tsx
```diff
@@ -0,0 +1,57 @@
...
+  });
+
+  return (
+    <Snackbar 
+      open={needRefresh} 
+      message={t('settingsUpdateAvailable')}
+      action={(
+        <Button onClick={() => updateServiceWorker(true)}>
+          {t('settingsReload')}
+        </Button>
```
tananaev, line 51: Can we just use an icon button? Like this one:

https://mui.com/material-ui/material-icons/?query=ref&selected=Refresh

It looks better and also we can remove one of the strings.

modern/src/ReloadPrompt.tsx
tananaev: We're talking about `state.session.server.attributes.serviceWorkerUpdateInterval`.

modern/vite-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="vite-plugin-pwa/react" />
```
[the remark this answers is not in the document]
tananaev: What is it complaining about? I suggest we remove this. We're not planning to make any changes to that file anyway.

modern/src/UpdateCheckPrompt.tsx
jinzo: Good idea, did not even think about this.
image

modern/vite-env.d.ts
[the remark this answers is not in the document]
jinzo: image
Removed it.

modern/src/common/util/preferences.js
```diff
@@ -39,3 +39,10 @@ export const useAttributePreference = (key, defaultValue) => useSelector((state)
   }
   return defaultValue;
 });
+
+export const useServerAttributePreference = (key, defaultValue) => useSelector((state) => {
```
tananaev, line 43: It doesn't make sense to add this function. You can simply write:
```js
state.session.server.attributes.key || defaultValue
```

jinzo: Thank you, did not know that.

tananaev:
Merged, thank you.

jinzo:
Thank you very much for your patience.

PR #1195, opened 2023-11-23
Add camper device category

TheZoker:
This PR adds a new device category called `Camper`. 

Screenshot of a tracker with that category:
image


tananaev (commented):
1. What is the source of the image?
2. Please don't include string translations. You should only include English.

TheZoker:
The source is here:
https://www.iconpacks.net/free-icon/camper-van-13375.html

It does not require credit or similar.

I removed the translations

tananaev:
Can you please provide a screen shot of the map with another category next to is, so we can compare.

TheZoker:
Here are three in a row:

image

tananaev:
It doesn't look very consistent with the style. Some very small details. And wheels are in different style.

TheZoker:
Alright I will look for a different file. Where do the original icons come from? Maybe that will help to find something. 

It needs to be a icon without credit, right?

tananaev:
Yes, it needs to be without credit. I believe the original icons are from Icons8.

TheZoker:
Alright so this should be the car file:
https://icons8.de/icon/set/car/material-rounded

And this would be the camper:
https://icons8.de/icon/set/camper/material-rounded

Would that work for you?

tananaev:
I don't think those are links to specific image.

TheZoker:
Ah sorry, here is the correct link:
https://icons8.de/icon/93858/wohnwagen-campingplatz

https://icons8.de/icon/85104/auto

TheZoker:
Here is the new preview:

image

Next to the camper I also added a caravan.

tananaev:
Wait. Are we adding two icons now? I think that's a bit of overkill.

TheZoker:
A friend that also uses the software for his caravan/trailer asked me if I can add this icon as well. But I can also remove it again

tananaev:
I think one icon is enough.

TheZoker:
Done, second icon is removed again

tananaev:
Merged, thanks.

PR #1199, opened 2023-12-05
Details drawer

arg3ni5:
feat: Add DetailsDrawer for showing device position details

Adds a new component, DetailsDrawer, which displays position details for a selected device. This drawer can be opened from the StatusCard component, providing additional contextual information about the device's state. Users can view attributes for the selected device. The addition of the DetailsDrawer enhances the user experience by offering a convenient way to explore and analyze position data.  When enabled in the preferences, show DetailsDrawer, when is disabled redirect to normal route.

image

feat: Add Quick Details Button in StatusCard component

This commit introduces a new feature to the StatusCard component by adding a Quick Details button. When enabled in the preferences, the button appears below the table in the card. Clicking on the button triggers the onDetailsClick function and displays additional information. This enhancement provides users with a convenient way to quickly access more details about the device.

image

### DIFF
--- a/modern/src/common/components/StatusCard.jsx
+++ b/modern/src/common/components/StatusCard.jsx
@@ -17,6 +17,7 @@ import {
   MenuItem,
   CardMedia,
   Snackbar,
+  Button,
 } from '@mui/material';
 import makeStyles from '@mui/styles/makeStyles';
 import CloseIcon from '@mui/icons-material/Close';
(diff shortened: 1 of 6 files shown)


tananaev:
Were there any feature requests for this? I'm not sure I understand the value over what's already available.

arg3ni5:
I think there is no request for this. 
The value is that by opening it in a drawer it is easier and faster to see the details of the different devices. 

https://github.com/traccar/traccar-web/assets/12776890/0036ef6a-792f-48e9-81e7-df93f3d42c7f

The features are option, so you can enable or disables from preferences.

https://github.com/traccar/traccar-web/assets/12776890/4281c26a-e18e-47fc-97f3-fcf5d14330e4

tananaev:
I think we need to wait and see if there's an interest for it. This basically duplicates existing functionality.

TheZoker:
I really like this change. It allows to do/see most stuff from the main page without leaving it

ligenzasantiago:
I find it a very interesting function, especially to be able to see the map and a status panel at the same time as in the legacy version.
Something that I think would add a lot is that every time the device changes position it is updated in this new panel.

Turbovix:
I suggest replacing this version proposed here with the existing one.

PR #1205, opened 2024-01-05
Add MapBox Dark to useMapStyle

juniormelo26:
Adding the dark map for those who have a MapBox key.


juniormelo26:
@tananaev

modern/src/resources/l10n/en.json
```diff
@@ -1,594 +1,595 @@
 {
-    "sharedLoading": "Loading...",
```
tananaev, line 2: Something happened with the localization file.

juniormelo26:
Added line 361 "mapMapboxStreetsDark": "Mapbox Streets Dark"

tananaev:
It looks like you reformatted the whole file. You have to fix that.

modern/src/resources/l10n/en.json
juniormelo26: I didn't understand why it was formatted like that, I flew or corrected it.

juniormelo26:
@tananaev  I made the correction. Please check when you can.

Issue #802, opened 2020-12-02
Device Category "0"


neilromig:
Using the Modern interface it is possible to set the device category to "0". This will then cause the old interface to fail in displaying devices.

To reproduce, edit the device in Modern interface and select the blank entry at top of the dropdown list of categories.

It should not be possible to set a value for device category in the Modern interface that is not handled by the "old" interface.

PR #1212, opened 2024-01-19
Replace Select with Autocomplete with single Selects

jinzo:
As described in https://github.com/traccar/traccar-web/pull/1189#issuecomment-1900520452 this is a new PR where I'm working on replacing all the single Select(s) with Autocomplete. 
Currently in draft mode, as I'm not particularly happy how the SelectField change is implemented and I probably missed some Select(s)


modern/src/common/components/SelectField.jsx
```diff
@@ -33,20 +34,47 @@ const SelectField = ({
...
+                                      if (typeof option != 'object') {
+                                        if (keyField) {
+                                          option = items.find(obj => obj[keyField] === option)
+                                        } else {
+                                          option = items.find(obj => obj === option)
+                                        }
+                                      }
+                                      const title = option ? titleGetter(option) : ''
+                                      return title ? title : ''
+                                      }
```
tananaev, line 66: You have to fix formatting here

modern/src/reports/components/ReportFilter.jsx
```diff
@@ -93,17 +93,39 @@ const ReportFilter = ({ children, handleSubmit, handleSchedule, showOnly, ignore
...
+                  </MenuItem>
+                )}
+                value={devices[deviceId] || null}
+                onChange={(event, newValue) => {
+                  dispatch(devicesActions.selectId(newValue?.id ?? null))
+                }}
+                renderInput={(params) => <TextField {...params} label={t('reportDevice')} />}
+              />
+            </>
+          )}
```
tananaev, line 128: Can we use `SelectField` here?

jinzo: We probably could yeah, the plan for the next commits is to go thorough all of the Select(s) and check which ones makes sense to convert to `SelectField` and/or `Autocomplete` and then finalize the `SelectField` changes that I implemented. This was the first change before I encountered SelectField and will now go re-check in the next sweep as there are some candidates in `MaintainancePage`, `ChartReportPage`,...

modern/src/common/components/SelectField.jsx
jinzo: The plan is to go thorough all the usecases and check if I need to add any functionality to the SelectField and then clean this up a bit. I also don't like the onChange stuff, I'll clean that up too.

jinzo:
So I reworked the option label, changed some more code to use SelectField.
Please check. Also - we are currently doing a little bit 'hacky' solution to support the onChange e.target.value - I would prefer changing this and passing the value directly, but on the other hand I will understand if you prefer that it stays like it currently is with "emulated" e.target.value.

modern/src/settings/UserPage.jsx
```diff
@@ -267,10 +267,11 @@ const UserPage = () => {
                 </Select>
               </FormControl>
               <SelectField
-                value={(item.attributes && item.attributes.timezone) || ''}
-                emptyValue=""
+                value={(item.attributes && item.attributes.timezone) || null}
+                emptyValue={null}
                 onChange={(e) => setItem({ ...item, attributes: { ...item.attributes, timezone: e.target.value } })}
                 endpoint="/api/server/timezones"
+                keyField={false}
```
tananaev, line 274: Why do we need `keyField` in addition to `keyGetter`?

jinzo: Because sometimes we have to find the default/selected option even if only the key value is passed. For example:   https://github.com/traccar/traccar-web/blob/2bd044ecf9fa456e168057acc513bd3b4582d0c6/modern/src/settings/PreferencesPage.jsx#L268

tananaev: Unfortunately it does not answer my question.

jinzo: Ok, let me try to explain it in a different way `value` in the above linked case is either a full Object, from which we can get the label via `titleGetter` or a string `'name'`. We need to have a way to get a label from the value `'name'` and we cannot pass a string to `titleGetter`. The current solution searches all the possible options (so an array of Objects) for the value `'name'` and it needs to know in which property of the Object to look for. `keyGetter` gets the key value from the Object and cannot help us to search for the appropriate Object in the array of Objects.

The second option would be to change the code so that the value passed is always a full object.

jinzo: Oh, and the original example you commented it flips it around, as when we add the above logic, we need a way to handle non Object values (so just an array of strings). So indeed, this can all be simplified if we change the few instances where the default value passed is of a different type (so a string instead of Object) as all the options.

tananaev: I all for simplification if there's an opportunity for it.

jinzo: So apparently, after reading the above mentioned issue and this one (https://github.com/mui/material-ui/issues/23708)  I found out that my initial explanation was a bit to narrow. Because we are always working with **keys/IDs** (eg: supplied `value` is always a key/id, no matter if it's a default value or one from the backend, and we are always returning key/ids in onChange) we need to convert that to full Object(s) if we wish to use `titleGetter` - contrary to what I wrote before, that this is not only needed for default values but we need it for stored/selected values too.
Because the flow (for a selected or default) goes like that:
`value` -> `getOptionLabel` -> `titleGetter` for which we need the Object. So there are basically 2 options:
1. We convert the `value` key/id to an Object and feed that to `titleGetter` - like we do now, and where we need the `keyField` to search the array for the Object
2. We always work with Objects instead of keys/ids -> so `value` would contain Object(s) and `onChange` would return Objects and the Redux would keep objects instead of keys/ids - **I think this would be way too much work**.

Unfortunately this works differently than `Select` and causes us problems. Also doesen't seem that the MUI developers will address this, as people have been complaining about this for quite some time with no proper solution in sight. We also can't easly rework this to use something akin like `objectGetter` because it won't work for `endpoint` enabled `SelectField`. 
TL;DR: I have not found any other proper solutions for this and this implementation - even tho it's not so nice, is the best one we can do without major refactoring of either the frontend or the backend.

tananaev: There must be a way to wrap `Autocomplete` with something to make it work similar to normal select field.

jinzo: I mean, that's exactly what `SelectField` now does, but if we don't have a uniform property in all of the Objects for where the `value` is (eg. `id`) we need to specify where our solution/wrapper searches for Object(s). Similarly why we need `keyGetter`, `titleGetter` because Objects do not have uniform properties (eg. value in `id`, title/label in `name`) we need a way to specify this. 
On the other hand, I did just figure out I am a little bit of an idiot and I can probably reuse `keyGetter` now with all the info I have and the refactoring.

tananaev: That was my original point. Having key and title getters should be enough.

jinzo: Yeah I over complicated and did not re-think completely. This latest commit is a lot simpler and works good.

modern/src/settings/GeofencePage.jsx
```diff
@@ -76,7 +76,7 @@ const GeofencePage = () => {
                 label={t('sharedDescription')}
               />
               <SelectField
-                value={item.calendarId || 0}
+                value={item.calendarId || null}
```
tananaev, line 79: `|| null` is probably not needed here and in other places

modern/src/settings/PreferencesPage.jsx
```diff
@@ -272,8 +272,8 @@ const PreferencesPage = () => {
                   label={t('devicePrimaryInfo')}
                 />
                 <SelectField
-                  emptyValue=""
-                  value={attributes.deviceSecondary || ''}
+                  emptyValue={null}
```
tananaev, line 275: Shouldn't `null` be the default already?

modern/src/settings/GeofencePage.jsx
jinzo: It'll probably generate warnings if item.calendarId (and similar) will be `false`. But I will check this.

modern/src/settings/PreferencesPage.jsx
jinzo: Yeah generally speaking I think we can change the default value from "" to `null`, will double check if there are any differences between `Select` and `Autocomplete` in this regard - but I mostly changed it to `null` as otherwise `Autocomplete` generated a Warning (but still worked without a hitch).

tananaev: I meant that we don't need to explicitly set it to null in every place.

modern/src/settings/GeofencePage.jsx
tananaev: It would never be false though, right? It will either be not set or an id.

modern/src/settings/PreferencesPage.jsx
jinzo: So, emptyValue difference between `0` and `null` apparently stems from Issue #802 ? And there is difference on how this is set in the old interface?

modern/src/settings/GeofencePage.jsx
jinzo: Removing this actually leads to a MUI/React error:

> MUI: A component is changing the uncontrolled value state of Autocomplete to be controlled.
> Elements should not switch from uncontrolled to controlled (or vice versa).
> Decide between using a controlled or uncontrolled Autocomplete element for the lifetime of the component.
> The nature of the state is determined during the first render. It's considered controlled if the value is not `undefined`.
> More info: https://fb.me/react-controlled-components

modern/src/settings/PreferencesPage.jsx
jinzo: But I guess we can default to `null` as now most of initial values (if no value stored) are defaulting to `null` instead of 0.

modern/src/settings/GeofencePage.jsx
jinzo: After the last commit we are now defaulting to null and it works as expected across the board.

modern/src/settings/PreferencesPage.jsx
jinzo: We are now defaulting to `null` and it works ok.

modern/src/common/components/SelectField.jsx
```diff
@@ -8,9 +8,9 @@ const SelectField = ({
   label,
   fullWidth,
   multiple,
-  value,
-  emptyValue = 0,
-  emptyTitle = '\u00a0',
```
tananaev, line 13: Are you sure this works with the regular select? It doesn't have a problem with item height?

jinzo: Currently emptyTitle is not used with the regular `Select` as it was used only in single `Select` and we are using it currently only for `multiple`
`{!multiple && emptyValue !== null && (` <-- so this is now always false for `Select` and that's why I removed it from the regular `Select`. And because of how the `Autocomplete` works, the `emptyTitle` is never actually displayed like this previous code fixed (as you have a clear selection X that 'replaces' this `Select` functionality). The only other way (eg. not an initial empty value) it could display (in list, like `Select` does) is if there is somehow an invalid `option` which is not a correct state anyway (eg. `titleGetter`/`keyGetter` are wrongly configured or one `option` is different Object than the others in the list)

modern/src/common/components/SelectField.jsx
```diff
@@ -33,20 +40,36 @@ const SelectField = ({
...
+              value={value}
+              onChange={onChange}
+            >
+              {items.map((item) => (
+                <MenuItem key={keyGetter(item)} value={keyGetter(item)}>{titleGetter(item)}</MenuItem>
+              ))}
+            </Select>
+          </>
+        ) : (
+          <>
```
tananaev, line 58: Why do we need a fragment here?

modern/src/common/components/SelectField.jsx
```diff
@@ -33,20 +40,36 @@ const SelectField = ({
...
+            <Autocomplete
+              size="small"
+              options={items}
+              getOptionLabel={getOptionLabel}
+              renderOption={(props, option) => (
+                <MenuItem {...props} key={keyGetter(option)} value={keyGetter(option)}>{titleGetter(option)}</MenuItem>
+              )}
+              isOptionEqualToValue={(option, value) => keyGetter(option) == value}
+              value={value}
+              onChange={(e, nV) => {onChange({target:{value:nV ? keyGetter(nV) : emptyValue}})}}
```
tananaev, line 68: Can you please rename `nV` to something else. Maybe just `value`.

Also please add spacings to the object. Maybe splitting object into multiple lines is better to make it more readable.

jinzo: What do you think about reworking this and always returning just the value? And of-course fixing all the callbacks. From what I can see, absolutely all just do `e.target.value` and nothing else, so I think it would be a lot cleaner to just return the value and be done with it. Ofcourse I would wrap the `Select` callback too, for consistency.

tananaev: I kind of like the fact that we can replace back with the standard select if needed. It also consistent with all other input fields.

jinzo: Ok, so I removed the fragment and for now just added spaces and it looks readable to me just ok - please check and tell me if I need to break it in more lines.

modern/src/common/components/SelectField.jsx
```diff
@@ -33,20 +40,34 @@ const SelectField = ({
...
+          <Autocomplete
+            size="small"
+            options={items}
+            getOptionLabel={getOptionLabel}
+            renderOption={(props, option) => (
+              <MenuItem {...props} key={keyGetter(option)} value={keyGetter(option)}>{titleGetter(option)}</MenuItem>
+            )}
+            isOptionEqualToValue={(option, value) => keyGetter(option) == value}
+            value={value}
+            onChange={(_, value) => {onChange({ target: { value: value ? keyGetter(value) : emptyValue }})}}
```
tananaev, line 67: I think this should work also

```suggestion
            onChange={(_, value) => onChange({ target: { value: value ? keyGetter(value) : emptyValue }})}
```

jinzo:
Any chance that this gets merged? Or what would need to be done to get it merged?

tananaev:
Is it ready for review?

jinzo:
I was under the impression that all of the review questions were addressed - I just synced and fixed (the new?) linting.
But yes, it is ready for further review - thanks!

modern/src/common/components/SelectField.jsx
```diff
@@ -33,20 +40,34 @@ const SelectField = ({
...
-          onChange={onChange}
-        >
-          {!multiple && emptyValue !== null && (
-            <MenuItem value={emptyValue}>{emptyTitle}</MenuItem>
-          )}
-          {items.map((item) => (
-            <MenuItem key={keyGetter(item)} value={keyGetter(item)}>{titleGetter(item)}</MenuItem>
-          ))}
-        </Select>
+        {multiple ? (
```
tananaev, line 43: I think overall looks good. But I forgot now why are we not using `Autocomplete` for multi-select? It seems like it would be the most valuable there actually.

jinzo: Yes I agree, but I did not find an easy way to make it work/look 'properly' - that's why I split it up in two and first implemented single select.
The problem is that the MUI Autocomplete component when using multiple values 'adds' another row in the Input field and that does not look that good - see this comment for the image: https://github.com/traccar/traccar-web/pull/1189#issuecomment-1806079577
The idea was to move the actual 'search input' into the drop down list instead of growing the original input to 2 lines (so the original input would only contain selected values), like select2 does for example - but I did not find a sane/easy way of doing that at the time. So the decision was made to first do it for single select, get that merged so it makes my life easier with my big lists, then tackle the multi usecase later.

tananaev:
OK, let's merge this for now. Thanks

PR #1225, opened 2024-03-25
reset password: pressing enter does not submit form

gramakri:
it reloads the page because of missing submit button


tananaev:
Merged, thanks.

PR #1228, opened 2024-04-04
PWA fix and improvements

jinzo:
After finally finding some time to debug why the PWA is not handling errors (like no internet connection) good - only white blank page was displayed I finally figured out what's happening. 

Because the index.js size grew beyond 2MB, it did not get pre-cached in the Service Worker of the PWA. This warning was nicely displayed in every build, but I just missed it before:
```
warnings
  assets/index-T6yVBWG0.js is 3.82 MB, and won't be precached. Configure maximumFileSizeToCacheInBytes to change this limit.
``` 
This then meant that the html and css was pre-cached in the Service Worker, but the main JS not - which resulted in a blank web page when the main JS was not loadable over HTTP for any reason. 

While working on this, I added all the needed assets to the Service Worker precache (fonts, alarm sound, favicon, apple touch icon) so that the whole PWA is now precached and managed by the Service Worker (like it should from the start). This leads to managed (and displayed) errors and (probably?) improved performance, as everything is precached and managed by the Service Worker (also avoids possibility of mismatched versions of assets).

I also updated the vite-plugin-pwa to the latest version, as there seems to be quite some bugfixing going around there too. This changes were tested on my staging and work great.

### DIFF
--- a/modern/vite.config.js
+++ b/modern/vite.config.js
@@ -19,8 +19,11 @@ export default defineConfig(() => ({
     svgr(),
     react(),
     VitePWA({
+      includeAssets: ['favicon.ico', 'apple-touch-icon-180x180.png'],
       workbox: {
         navigateFallbackDenylist: [/^\/api/],
+        maximumFileSizeToCacheInBytes: 4000000,
+        globPatterns: ['**/*.{js,css,html,woff,woff2,mp3}'],
       },
       manifest: {
         short_name: '${title}',
--- a/modern/package-lock.json
+++ b/modern/package-lock.json
@@ -38,7 +38,7 @@
         "recharts": "^2.10.4",
         "redux": "^5.0.1",
         "vite": "^5.0.13",
-        "vite-plugin-pwa": "^0.17.4",
+        "vite-plugin-pwa": "^0.19.7",
         "vite-plugin-svgr": "^4.2.0",
         "wellknown": "^0.5.0"
       },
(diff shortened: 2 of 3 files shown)


modern/vite.config.js
```diff
@@ -19,8 +19,11 @@ export default defineConfig(() => ({
     svgr(),
     react(),
     VitePWA({
+      includeAssets: ['favicon.ico', 'apple-touch-icon-180x180.png'],
       workbox: {
         navigateFallbackDenylist: [/^\/api/],
+        maximumFileSizeToCacheInBytes: 4000000,
+        globPatterns: ['**/*.{js,css,html,woff,woff2,mp3}'],
```
tananaev, line 26: Can you please explain each of the change.

jinzo: By default, Vite PWA generates a Service Worker that caches every `.{js,css,html}` file in `assets/` folder that is smaller than 2MB. 
`globPatterns` change here, extends the extensions to woff font files and mp3 for the alarm sound. 
`maximumFileSizeToCacheInBytes` increases the size limit to 4MB, because Traccars main generated JS file (index-hash.js) is now bigger than 2MB. 
`includeAssets` includes the favicon.ico and apple-touch-icon static files that are not in `assets/` folder.

All this 3 changes combined, make sure that all the static assets that we need for a working PWA are cached by the Service Worker. This then means that the Service Worker serves them, even before a HTTP browser cache is involved. 
image
Source: https://web.dev/articles/service-worker-caching-and-http-caching

Before this changes (and in recent builds) Service Worker only cached `index.html` and some internal .js files that it needed for work - which caused a problem if you loaded the PWA at least once (the Service Worker got installed and it cached the `index.html`), then you opened the PWA (or refreshed the page) without internet connection. The `index.html` got served from the Service Worker, but the `index-hash.js` HTTP request failed -> and this resulted in a blank white webpage with no indicator that anything is wrong. So with ensuring that all the assets are precached with the Service Worker version together, we avoid problems like this (or partial upgrades, where Service Worker serves cached `index.html` which is referencing now deleted `index-hash.js` file, that might not be in browser cache anymore - before the new Service Worker with new `index-hash.js` reference is installed/updated).

This also prepares the groundwork for possible offline functionality - as now the Service Worker (after first load) caches all the assets (html,css,fonts,...) we need for the PWA to work. And maybe we can, down the line, show something more than the "there is no network" error that is shown after this PR (instead of a blank page).

tananaev: This is awesome. That's for looking into this. What about other files outside of the `assets/` folder? Like all of the PWA images.

jinzo: Per the Vite PWA documentation those are included automatically:
`You don't need to configure PWA Manifest icons on includeAssets option`
Source: https://vite-pwa-org.netlify.app/guide/static-assets.html
The same goes for the base `styles.css`, it's not mentioned in the docs, but I can confirm they get included automatically too:
image

So after another pass, I think Service Worker now caches all static assets it can.

tananaev:
Merged, thanks again.

PR #1233, opened 2024-04-25
Add the possibility to add multiple map overlays

pieternaber:
Map overlays are a powerfull function in Traccar. But to only allow one overlay can be not enough.

That's why I added the possibility to select multiple layers on the preference page and use the overlays on your map.

### DIFF
--- a/src/map/overlay/MapOverlay.js
+++ b/src/map/overlay/MapOverlay.js
@@ -1,37 +1,41 @@
-import { useId, useEffect } from 'react';
+import { useEffect } from 'react';
 import { useAttributePreference } from '../../common/util/preferences';
 import { map } from '../core/MapView';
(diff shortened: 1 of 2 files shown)


camende:
That would be a perfect addition.

src/map/overlay/MapOverlay.js
```diff
@@ -1,37 +1,41 @@
...
+          if (map.getLayer(mapOverlay.id)) {
+            map.removeLayer(mapOverlay.id);
+          }
+          if (map.getSource(mapOverlay.id)) {
+            map.removeSource(mapOverlay.id);
+          }
+        });
       }
     };
-  }, [id, activeOverlay]);
```
tananaev, line 34: Why are we removing this?

tananaev:
Just implemented something similar.

PR #1240, opened 2024-06-09
Return status in handleReboot

geelongmicrosoldering:
Return response.statusText instead response.text in handleReboot.
Prevents displaying html source of error pages in user feedback

Discussed in [PR #5332](https://github.com/traccar/traccar/pull/5332)


tananaev:
Thanks.

PR #1243, opened 2024-06-16
draw route with speed based color

jcardus:
Draw the route with intepolated color.
Changed MapRoutePoints arrow to text instead of icon so that we can change the color.
I tried to use the palette instead of the hard coded yellow and red colors, but it didn't look good.

Captura de ecrã 2024-06-16, às 20 00 39

### DIFF
--- a/src/common/util/colors.js
+++ b/src/common/util/colors.js
@@ -4,7 +4,7 @@ export const interpolateColor = (color1, color2, factor) => {
   if (factor > 1) factor = 1;
   if (factor < 0) factor = 0;
 
-  const parseRgb = (rgb) => rgb.match(/\d+/g).map(Number);
+  const parseRgb = (rgb) => rgb.split(',').map(Number);
 
   const c1 = parseRgb(hexToRgb(color1));
   const c2 = parseRgb(hexToRgb(color2));
--- a/src/common/util/colors.js
+++ b/src/common/util/colors.js
@@ -1,17 +1,17 @@
-import { hexToRgb, rgbToHex, decomposeColor } from '@mui/material';
+import { decomposeColor } from '@mui/material';
 
 export const interpolateColor = (color1, color2, factor) => {
   if (factor > 1) factor = 1;
   if (factor < 0) factor = 0;
 
-  const c1 = decomposeColor(hexToRgb(color1)).values;
-  const c2 = decomposeColor(hexToRgb(color2)).values;
+  const c1 = decomposeColor(color1).values;
+  const c2 = decomposeColor(color2).values;
 
   const r = Math.round(c1[0] + factor * (c2[0] - c1[0]));
   const g = Math.round(c1[1] + factor * (c2[1] - c1[1]));
   const b = Math.round(c1[2] + factor * (c2[2] - c1[2]));
 
-  return rgbToHex(`rgb(${r}, ${g}, ${b})`);
+  return `rgb(${r}, ${g}, ${b})`;
 };
 
 export const getSpeedColor = (speed, max) => interpolateColor('#FFFF00', '#FF0000', speed / max);
--- a/src/common/util/colors.js
+++ b/src/common/util/colors.js
@@ -14,4 +14,4 @@ export const interpolateColor = (color1, color2, factor) => {
   return `rgb(${r}, ${g}, ${b})`;
 };
 
-export const getSpeedColor = (speed, max) => interpolateColor('#FFFF00', '#FF0000', speed / max);
+export const getSpeedColor = (color1, color2, speed, max) => interpolateColor(color1, color2, speed / max);
--- a/src/common/util/colors.js
+++ b/src/common/util/colors.js
@@ -4,10 +4,8 @@ export const interpolateColor = (color1, color2, factor) => {
   if (factor > 1) factor = 1;
   if (factor < 0) factor = 0;
 
-  const parseRgb = (rgb) => rgb.split(',').map(Number);
-
-  const c1 = parseRgb(hexToRgb(color1));
-  const c2 = parseRgb(hexToRgb(color2));
+  const c1 = hexToRgb(color1).split(',').map(Number);
+  const c2 = hexToRgb(color2).split(',').map(Number);
 
   const r = Math.round(c1[0] + factor * (c2[0] - c1[0]));
   const g = Math.round(c1[1] + factor * (c2[1] - c1[1]));
--- a/src/common/util/colors.js
+++ b/src/common/util/colors.js
@@ -1,11 +1,11 @@
-import { hexToRgb, rgbToHex } from '@mui/material';
+import { hexToRgb, rgbToHex, decomposeColor } from '@mui/material';
 
 export const interpolateColor = (color1, color2, factor) => {
   if (factor > 1) factor = 1;
   if (factor < 0) factor = 0;
 
-  const c1 = hexToRgb(color1).split(',').map(Number);
-  const c2 = hexToRgb(color2).split(',').map(Number);
+  const c1 = decomposeColor(hexToRgb(color1)).values;
+  const c2 = decomposeColor(hexToRgb(color2)).values;
 
   const r = Math.round(c1[0] + factor * (c2[0] - c1[0]));
   const g = Math.round(c1[1] + factor * (c2[1] - c1[1]));
--- a/src/common/util/colors.js
+++ b/src/common/util/colors.js
@@ -14,4 +14,10 @@ export const interpolateColor = (color1, color2, factor) => {
   return `rgb(${r}, ${g}, ${b})`;
 };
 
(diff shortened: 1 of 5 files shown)


src/common/util/colors.js
```diff
@@ -0,0 +1,19 @@
+import { hexToRgb, rgbToHex } from '@mui/material';
+
+export const interpolateColor = (color1, color2, factor) => {
+  if (factor > 1) factor = 1;
+  if (factor < 0) factor = 0;
+
+  const parseRgb = (rgb) => rgb.match(/\d+/g).map(Number);
```
tananaev, line 7: Why do we need this?

src/map/MapRoutePath.js
```diff
@@ -79,17 +81,18 @@ const MapRoutePath = ({ name, positions, coordinates }) => {
...
-        coordinates,
-      },
-      properties: {
-        name,
-        color: reportColor,
-      },
-    });
+    const maxSpeed = positions.map((p) => p.speed).reduce((a, b) => Math.max(a, b), -Infinity);
+    const features = [];
+    for (let i = 0; i < positions.length - 1; i += 1) {
```
tananaev, line 86: What happens if there's a single point?

src/map/MapRoutePath.js
```diff
@@ -79,17 +81,18 @@ const MapRoutePath = ({ name, positions, coordinates }) => {
...
+      const p1 = positions[i];
+      const p2 = positions[i + 1];
+      features.push(lineString(
```
tananaev, line 89: What does `lineString` function do?

src/map/MapRoutePoints.js
```diff
@@ -27,11 +28,13 @@ const MapRoutePoints = ({ positions, onClick }) => {
...
       source: id,
+      paint: {
+        'text-color': ['get', 'color'],
+      },
       layout: {
-        'icon-image': 'arrow',
-        'icon-allow-overlap': true,
-        'icon-rotate': ['get', 'rotation'],
-        'icon-rotation-alignment': 'map',
+        'text-field': '▲',
```
tananaev, line 35: If we don't use the arrow anymore, we should probably remove it.

src/common/util/colors.js
jcardus: hexToRgb returns a string, we can change it to this:

const parseRgb = (rgb) => rgb.split(',').map(Number);

src/map/MapRoutePath.js
jcardus: nothings is drawn

src/map/MapRoutePath.js
jcardus: it's a helper from turf, creates a lineString feature

src/map/MapRoutePoints.js
jcardus: ok

src/common/util/colors.js
jcardus: I removed it, please check

src/map/MapRoutePoints.js
jcardus: I removed the arrow

src/map/MapRoutePath.js
tananaev: But there should be one point, right?

src/map/MapRoutePath.js
tananaev: I thought it's a trivial thing to create a line. What does the result look like?

src/map/MapRoutePath.js
jcardus: It will be in the other layer (the arrow)

tananaev: I see. That makes sense then.

src/map/MapRoutePath.js
jcardus: It is, it's just a helper, it creates the whole structure, including the properties (like it was done before manually):

{
      type: 'Feature',
      geometry: {
        type: 'LineString',
        coordinates: [[....]]
      },
      properties: {
        color: '...',
      },
};

tananaev: I prefer to keep it as JSON instead of using helper.

jcardus: ok, lineString is gone.

src/common/util/colors.js
```diff
@@ -0,0 +1,17 @@
...
+  const c2 = decomposeColor(color2).values;
+
+  const r = Math.round(c1[0] + factor * (c2[0] - c1[0]));
+  const g = Math.round(c1[1] + factor * (c2[1] - c1[1]));
+  const b = Math.round(c1[2] + factor * (c2[2] - c1[2]));
+
+  return `rgb(${r}, ${g}, ${b})`;
+};
+
+export const getSpeedColor = (speed, max) => interpolateColor('#FFFF00', '#FF0000', speed / max);
```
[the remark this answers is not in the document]
jcardus: Before I was receiving the colors in the getSpeedColor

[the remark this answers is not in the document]
tananaev: It doesn't answer the question though.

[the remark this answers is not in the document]
jcardus: You didn't comment on this before. 
Yes, it answers the question, the getSpeedColor is called inside the components, the theme is available there.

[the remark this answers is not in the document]
tananaev: OK, so can we update it then?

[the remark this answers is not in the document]
jcardus: I tried a lot of combinations from the theme color, the one it's less bad is warning to success. But the colors are very "near" from each other so you don't get the same effect.

I didn't find any option in the palette to replace the yellow. 

Here's an example:

Screenshot 2024-06-17 at 01 13 36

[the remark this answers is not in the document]
tananaev: Shouldn't the success be green?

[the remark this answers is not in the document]
jcardus: Sorry, I meant warning to error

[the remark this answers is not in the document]
tananaev: What does it look like with success to error?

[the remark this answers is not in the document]
jcardus: Captura de ecrã 2024-06-17, às 04 46 20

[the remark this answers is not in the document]
tananaev: Maybe we can mix 3 colors? Success, warning and error.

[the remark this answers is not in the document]
jcardus: ok, and maybe we can reverse it. In traffic, usually green is moving and red is stoped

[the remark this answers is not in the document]
tananaev: Good point. I guess it depends on what most people use it for.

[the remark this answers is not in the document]
jcardus: error -> warning -> success
Screenshot 2024-06-17 at 18 36 28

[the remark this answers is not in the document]
tananaev: Looks good, but I wonder if it makes sense to reverse colors or not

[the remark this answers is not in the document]
jcardus: success -> warrning -> error
image

ivanfmartinez:
Very nice to see speed differences, but interpolating with current max speeds can have different colors when looking for different data. 

Maybe this can be a option to use max speed of view or a fixed configurable number per device because 80 km/h can be very good to see in red for a small truck, but not for a standard car that can run at 120 km/h on roads.

Or if still using the max speed at least show it to give to the user a notion about how the colors are distributed.

ivanfmartinez:
personally I like the success -> warrning -> error version 

green on low speed, red on higher speed

src/map/MapRoutePath.js
```diff
@@ -1,7 +1,9 @@
 import { useTheme } from '@mui/styles';
 import { useId, useEffect } from 'react';
 import { useSelector } from 'react-redux';
+import { featureCollection } from '@turf/turf';
```
tananaev, line 4: Can we expand `featureCollection` as well? I'm sure it's also a very simple json.

tananaev:
Merged, thank you.

Kalabint:
Here is another suggestion: Turbo Color Ramp.

image

It would look something like this (Range: 0-200km/h, divided into 100 Color Steps):

From Cooler Colors for slower movement to Green in the Middle (about 80km/h) to Yellow/Orange (120km/h) with Red for high speeds.

chrome_WeGyuyZHhT
chrome_J4A4CSvSvD

Kalabint:
The Information density with the MUI Color Scheme is not so great, because the second half is the mapping from Orange to Red, which are very similar colors.

Using the Turbo Color Ramp it would look something like this in Traccar:

image

jcardus:
@Kalabint I agree, but I think what's needed here now it's a legend. @tananaev what do you think?

tananaev:
Don't have a very strong opinion. I'm more worried about quality and reliability of the solution.

PR #1244, opened 2024-06-19
Add mapGoogleKey and mapGoogleTraffic in es.json

jorgeart81:
Captura de pantalla 2024-06-18 224028

### DIFF
--- a/src/resources/l10n/es.json
+++ b/src/resources/l10n/es.json
@@ -359,6 +359,8 @@
     "mapGoogleRoad": "Google Carreteras",
     "mapGoogleHybrid": "Google Híbrido",
     "mapGoogleSatellite": "Google Satélite",
+    "mapGoogleTraffic": "Google Tráfico",
+    "mapGoogleKey": "Google API Key",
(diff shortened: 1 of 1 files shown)


tananaev:

https://www.transifex.com/traccar

jorgeart81:
Hello Anton, thanks for your response. The change was to find the [mapGoogleKey](https://github.com/traccar/traccar-web/blob/c42bb2baac7e66f1fb330ffbf941c1ce09b19737/src/common/attributes/useCommonUserAttributes.js#L53) value for I10n translation in Spanish. 
Currently mapGoogleKey is only found in [en.json (L363)](https://github.com/traccar/traccar-web/blob/c42bb2baac7e66f1fb330ffbf941c1ce09b19737/src/resources/l10n/en.json#L363)
This does not affect whether Google Maps can be added, it just makes it easier to know the name of the attribute that is being entered.

Regards.

tananaev:
Make sure you read my previous comment. Transifex is the source of truth for localization. We don't make any manual changes to translation because they will be overwritten by Transifex.

PR #1247, opened 2024-06-22
Chart Report - Allow multiple X Axis for different time indexes

geelongmicrosoldering:
I noticed an issue displaying some data in Chart Reports.
For devices that periodically disable GPS, 'fixTime' can be the same across multiple updates, while other data changes.

So displaying 'Battery Level' over time for example, results in stacked ticks, because 'fixTime' is used for the X axis:
image

This is probably ideal behavior for some data, but not for others. Proposed solution is allowing user to choose which time to use by adding an extra select filter on chart reports:
image

I'm not convinced its the best approach, but its the simplest I could think of.

Criticism is welcome 😄

### DIFF
--- a/src/reports/ChartReportPage.jsx
+++ b/src/reports/ChartReportPage.jsx
@@ -142,7 +142,6 @@ const ChartReportPage = () => {
             >
               <XAxis
                 dataKey={timeType}
-                interval="preserveStartEnd"
                 type="number"
                 tickFormatter={(value) => formatTime(value, 'time')}
                 domain={['dataMin', 'dataMax']}
--- a/src/reports/ChartReportPage.jsx
+++ b/src/reports/ChartReportPage.jsx
@@ -33,6 +33,7 @@ const ChartReportPage = () => {
   const [items, setItems] = useState([]);
   const [types, setTypes] = useState(['speed']);
   const [type, setType] = useState('speed');
+  const [timeIndex, setTimeIndex] = useState('fixTime');
 
   const values = items.map((it) => it[type]);
   const minValue = Math.min(...values);
@@ -52,6 +53,8 @@ const ChartReportPage = () => {
         const data = { ...position, ...position.attributes };
         const formatted = {};
         formatted.fixTime = dayjs(position.fixTime).valueOf();
+        formatted.deviceTime = dayjs(position.deviceTime).valueOf();
+        formatted.serverTime = dayjs(position.serverTime).valueOf();
         Object.keys(data).filter((key) => !['id', 'deviceId'].includes(key)).forEach((key) => {
           const value = data[key];
           if (typeof value === 'number') {
@@ -112,6 +115,21 @@ const ChartReportPage = () => {
             </Select>
           </FormControl>
         </div>
+        <div className={classes.filterItem}>
+          <FormControl fullWidth>
+            <InputLabel>{t('commandIndex')}</InputLabel>
+            <Select
+              label={t('commandIndex')}
+              value={timeIndex}
+              onChange={(e) => setTimeIndex(e.target.value)}
+              disabled={!items.length}
+            >
+              <MenuItem value="fixTime">{t('positionFixTime')}</MenuItem>
+              <MenuItem value="deviceTime">{t('positionDeviceTime')}</MenuItem>
+              <MenuItem value="serverTime">{t('positionServerTime')}</MenuItem>
+            </Select>
+          </FormControl>
+        </div>
       </ReportFilter>
       {items.length > 0 && (
         <div className={classes.chart}>
@@ -123,7 +141,8 @@ const ChartReportPage = () => {
               }}
             >
               <XAxis
-                dataKey="fixTime"
+                dataKey={timeIndex}
+                interval="preserveStartEnd"
                 type="number"
                 tickFormatter={(value) => formatTime(value, 'time')}
                 domain={['dataMin', 'dataMax']}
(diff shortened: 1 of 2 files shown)


src/reports/ChartReportPage.jsx
```diff
@@ -112,6 +115,21 @@ const ChartReportPage = () => {
             </Select>
           </FormControl>
         </div>
+        <div className={classes.filterItem}>
+          <FormControl fullWidth>
+            <InputLabel>{t('commandIndex')}</InputLabel>
```
tananaev, line 120: I don't think this is the right label. We should probably add a new string. Maybe it should just say "time".

src/reports/ChartReportPage.jsx
```diff
@@ -123,7 +141,8 @@ const ChartReportPage = () => {
               }}
             >
               <XAxis
-                dataKey="fixTime"
+                dataKey={timeIndex}
+                interval="preserveStartEnd"
```
tananaev, line 145: What does this do?

src/reports/ChartReportPage.jsx
```diff
@@ -33,6 +33,7 @@ const ChartReportPage = () => {
   const [items, setItems] = useState([]);
   const [types, setTypes] = useState(['speed']);
   const [type, setType] = useState('speed');
+  const [timeIndex, setTimeIndex] = useState('fixTime');
```
tananaev, line 36: I don't think `timeIndex` is a good name. Maybe `timeType` or something.

src/reports/ChartReportPage.jsx
geelongmicrosoldering: You caught me. You're right, it definitely should say "Time". I just couldn't find it in the existing strings and wasn't sure how to go about adding a new one (in multiple languages).

src/reports/ChartReportPage.jsx
geelongmicrosoldering: Agreed. it was actually timeType before i changed it to match the name in the UI 😆

I'll change it shortly

src/reports/ChartReportPage.jsx
geelongmicrosoldering: I had an issue with labels overlapping, but im struggling to reproduce it now. Here it is on a tab i still had open:
image
```
interval="preserveStartEnd"
```
Causes overlapping labels to be hidden/grouped, when the default behavior is to show them all. `preserveStartEnd` just makes sure that the label at the start/end of the axis are not hidden, so that the timestamp where the graph starts and ends is preserved

geelongmicrosoldering: Oh, i can reproduce it now. Had to clear browser cache. This is what it looks like without preserving start/end
image

src/reports/ChartReportPage.jsx
geelongmicrosoldering: Is there a trick to generating a string in all of the languages or should i go ahead and do it manually?

tananaev: You should only add English. Translations are handled through Transifex.

geelongmicrosoldering: > You should only add English. Translations are handled through Transifex.

Amazing, I avoided it for no reason 😆 
So I just add it to `src/resources/l10n/en.json` ?

geelongmicrosoldering: I think I did that right..

src/reports/ChartReportPage.jsx
tananaev: Can you provide a comparison? I'm not sure I understand what it actually does.

geelongmicrosoldering: Yep, but I actually think I should remove this line and address it in a separate issue, because I think it needs further investigation. It's also not directly related to this PR, so I might remove this line and go over it in a new issue for someone to look into further later.

[interval](https://recharts.org/en-US/api/XAxis#interval) has a few options, but the documentation is really lacking. It can be a number, but its not clear what the number represents (i assume every X tick is used as a label). Setting it to 0, will apparently show all ticks on the X axis. So if you have 1000 ticks, you have 1000 labels, and they will all overlap and be illegible.

A value of `"preserveStart"`, `"preserveEnd"` or `"preserveStartEnd"`, is supposed to result in an automatic calculation of which labels to show, so that none of them overlap. Basically if you have 1000 labels, it will hide most of them, with conditions.

The default value is "preserveEnd". This should hide all the overlapping labels, but prioritize keeping the last one
image

"preserveStart" should hide all the overlapping labels, but prioritize keeping the first one
image

"preserveStartEnd" is just a combination of both. it makes sure that the labels at either end of the axis are displayed, and any that are hidden are somewhere in between. Except in testing, it doesn't always, and often you get the same problem of labels overlapping
image

Initially when I tested this, I had the following:
image

And then after changing the interval to preserveStartEnd, i had this:
image

So I thought i had rectified the issue and moved on. Basically a "this isnt related but i'll fix that bug while im here".
But in further testing, the actual result seems to be unpredictable. The value likely should be preserveStartEnd, and that should be the [recharts ](https://recharts.org/) default, but it isn't, and changing it doesn't actually fix the issue.

More likely than not, this is actually a bug in recharts. A lot of people seem to be having issues with these labels. Interval should honor the value of [minTickGap](https://recharts.org/en-US/api/XAxis#minTickGap), which by default is supposed to ensure a minimum gap between the labels, but in practice it just doesn't work the way the docs say it does. It seems like there is some kind of disconnect between `interval`, `minTickGap`, and `tickFormatter`. Other people have come up with [convoluted fixes for this issue](https://github.com/recharts/recharts/issues/3305#issuecomment-1427133689), but really it comes down to recharts behavior.

Anyway, happy to remove this line and investigate it separately. It doesn't help, so its really pointless being here

tananaev:
Thank you.

Issue #1248, opened 2024-06-25
Error after comment #1243


joystickjockey:
**Describe the bug**
After including comment #1243 I get below error when running a combined report.
This is on latest master with some customisation but I compared the copy that worked with the one flagging the error and the only changes is from comment #1243 
```
TypeError: Cannot read properties of undefined (reading 'map')
    at index-DDKx5vYz.js:871:302118
    at uC (index-DDKx5vYz.js:40:24296)
    at Yg (index-DDKx5vYz.js:40:42451)
    at index-DDKx5vYz.js:40:40763
    at E (index-DDKx5vYz.js:25:1659)
    at MessagePort.de (index-DDKx5vYz.js:25:2039)
```
**To Reproduce**
1. On latest master select any device in combined reports and select show.

tananaev:
@jcardus

jcardus:
hello @joystickjockey, should be ok in the latest master. Can you check?

tananaev:
Reverted the recent changes because they broke other things as well, like the name.

PR #1249, opened 2024-06-25
Speed based color

jcardus:
https://github.com/traccar/traccar-web/issues/1248

### DIFF
--- a/src/map/MapRoutePath.js
+++ b/src/map/MapRoutePath.js
@@ -96,11 +96,10 @@ const MapRoutePath = ({ name, positions }) => {
         });
       }
 
-      map.getSource(id)
-        ?.setData({
-          type: 'FeatureCollection',
-          features,
-        });
+      map.getSource(id)?.setData({
+        type: 'FeatureCollection',
+        features,
+      });
     }
   }, [theme, positions, reportColor]);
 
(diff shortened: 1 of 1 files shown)


src/map/MapRoutePath.js
```diff
@@ -4,7 +4,7 @@ import { useSelector } from 'react-redux';
 import { map } from './core/MapView';
 import { getSpeedColor } from '../common/util/colors';
 
-const MapRoutePath = ({ name, positions, coordinates }) => {
```
tananaev, line 7: Why are we removing `coordinates` here?

jcardus: it's not being used anymore

src/map/MapRoutePath.js
```diff
@@ -77,31 +77,32 @@ const MapRoutePath = ({ name, positions, coordinates }) => {
...
+        });
+      }
 
-    map.getSource(id)?.setData({
-      type: 'FeatureCollection',
-      features,
-    });
-  }, [theme, positions, coordinates, reportColor]);
+      map.getSource(id)
+        ?.setData({
```
tananaev, line 100: Why is this reformatted?

jcardus: probably intellij automatically

src/map/MapRoutePath.js
tananaev: @jcardus Looks like this is not true. You broke combined report. I appreciate contributions, but please make sure you do proper due diligence before sending pull requests.

PR #1250, opened 2024-06-28
keep old behaviour in RoutePath when we just have coordinates (and no…

jcardus:
… positions)

include name property in both cases.

### DIFF
--- a/src/map/MapRouteCoordinates.js
+++ b/src/map/MapRouteCoordinates.js
@@ -43,21 +43,19 @@ const MapRouteCoordinates = ({ name, coordinates, deviceId }) => {
         'line-width': 2,
       },
     });
-    if (name) {
-      map.addLayer({
-        source: id,
-        id: `${id}-title`,
-        type: 'symbol',
-        layout: {
-          'text-field': '{name}',
-          'text-size': 12,
-        },
-        paint: {
-          'text-halo-color': 'white',
-          'text-halo-width': 1,
-        },
-      });
-    }
+    map.addLayer({
+      source: id,
+      id: `${id}-title`,
+      type: 'symbol',
+      layout: {
+        'text-field': '{name}',
+        'text-size': 12,
+      },
+      paint: {
+        'text-halo-color': 'white',
+        'text-halo-width': 1,
+      },
+    });
 
     return () => {
       if (map.getLayer(`${id}-title`)) {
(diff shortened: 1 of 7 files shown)


jcardus:
any comments?

tananaev:
I think we should probably split `MapRoutePath` to component that handles positions and the one that handles just coordinates.

jcardus:
can you check?

src/map/MapRoutePath.js
```diff
@@ -1,28 +1,13 @@
 import { useTheme } from '@mui/styles';
 import { useId, useEffect } from 'react';
-import { useSelector } from 'react-redux';
 import { map } from './core/MapView';
 import { getSpeedColor } from '../common/util/colors';
 
-const MapRoutePath = ({ name, positions, coordinates }) => {
+const MapRoutePath = ({ name, positions }) => {
```
tananaev, line 6: It looks like `name` is never used. We can probably remove it from here.

src/map/MapRouteCoordinates.js
```diff
@@ -0,0 +1,92 @@
...
+      layout: {
+        'line-join': 'round',
+        'line-cap': 'round',
+      },
+      paint: {
+        'line-color': ['get', 'color'],
+        'line-width': 2,
+      },
+    });
+    if (name) {
```
tananaev, line 46: For this one the name is always provided, so we can probably make it mandatory.

src/map/MapRoutePath.js
```diff
@@ -1,28 +1,13 @@
 import { useTheme } from '@mui/styles';
 import { useId, useEffect } from 'react';
-import { useSelector } from 'react-redux';
 import { map } from './core/MapView';
```
tananaev, line 3: Can you please check the warnings and errors.

jcardus: I don't have any warnings

src/map/MapRoutePath.js
```diff
@@ -1,27 +1,13 @@
...
+import { getSpeedColor } from '../common/util/colors';
 
-const MapRoutePath = ({ name, positions, coordinates }) => {
+const MapRoutePath = ({ positions }) => {
   const id = useId();
 
   const theme = useTheme();
 
-  const reportColor = useSelector((state) => {
```
tananaev, line 11: Why are we removing this?

jcardus: the color is based on the position speed

tananaev: Yeah, but it means you're ignoring the settings.

PR #1251, opened 2024-06-28
Update the itemsCoordinates state every time the length of itemsRoute…

jorgeart81:
Hello, I hope this PR is helpful to you.

##
Update the itemsCoordinates state every time the length of itemsRoutes changes.

## Before -> the map was returning to its initial state when the state was updated with the socket
https://github.com/traccar/traccar-web/assets/92123185/2c3e0279-eb36-471f-8787-ea079ccfbde7

## After 
https://github.com/traccar/traccar-web/assets/92123185/9bc325df-fc87-4fee-8170-9286ccaf5102

### DIFF
--- a/src/reports/CombinedReportPage.jsx
+++ b/src/reports/CombinedReportPage.jsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import { useSelector } from 'react-redux';
 import {
   Table, TableBody, TableCell, TableHead, TableRow,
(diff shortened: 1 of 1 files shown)


tananaev:
Why does it update everything on the socket update?

jorgeart81:
El socket actualiza el estado del store. Por ejemplo en [devices ](https://github.com/traccar/traccar-web/blob/759605342a573437a1b294a627d8e252ea0ebca9/src/reports/CombinedReportPage.jsx#L25)en el caso del CombinedReportPage.jsx 

It seems to me that [PageLayout ](https://github.com/traccar/traccar-web/blob/759605342a573437a1b294a627d8e252ea0ebca9/src/reports/CombinedReportPage.jsx#L56)should be the parent of reports, so it will not be affected.

tananaev:
Sounds like using `useMemo` would be the right approach here.

PR #1252, opened 2024-06-29
Fix/prevent re-render reports menu

jorgeart81:
Create a ReportLayout so that the ReportsMenu is not affected by the status update.
Ref -> [PR/1251](https://github.com/traccar/traccar-web/pull/1251)
##
## Video 
https://github.com/traccar/traccar-web/assets/92123185/60e22d0d-b604-4668-bf6e-e74f90b3b709

### DIFF
--- a/src/Navigation.jsx
+++ b/src/Navigation.jsx
@@ -59,6 +59,7 @@ import SharePage from './settings/SharePage';
 import AnnouncementPage from './settings/AnnouncementPage';
 import EmulatorPage from './other/EmulatorPage';
 import ReportsLayout from './reports/components/ReportsLayout';
+import SettingsLayout from './settings/components/SettingsLayout';
 
 const Navigation = () => {
   const navigate = useNavigate();
@@ -113,7 +114,7 @@ const Navigation = () => {
         <Route path="geofences" element={<GeofencesPage />} />
         <Route path="emulator" element={<EmulatorPage />} />
 
-        <Route path="settings">
+        <Route path="settings" element={<SettingsLayout />}>
           <Route path="accumulators/:deviceId" element={<AccumulatorsPage />} />
           <Route path="announcement" element={<AnnouncementPage />} />
           <Route path="calendars" element={<CalendarsPage />} />
--- a/src/Navigation.jsx
+++ b/src/Navigation.jsx
@@ -58,6 +58,7 @@ import LogsPage from './reports/LogsPage';
 import SharePage from './settings/SharePage';
 import AnnouncementPage from './settings/AnnouncementPage';
 import EmulatorPage from './other/EmulatorPage';
+import ReportsLayout from './reports/components/ReportsLayout';
 
 const Navigation = () => {
   const navigate = useNavigate();
@@ -154,7 +155,7 @@ const Navigation = () => {
           <Route path="user" element={<UserPage />} />
         </Route>
 
-        <Route path="reports">
+        <Route path="reports" element={<ReportsLayout />}>
           <Route path="combined" element={<CombinedReportPage />} />
           <Route path="chart" element={<ChartReportPage />} />
           <Route path="event" element={<EventReportPage />} />
--- a/src/reports/components/ReportsLayout.jsx
+++ b/src/reports/components/ReportsLayout.jsx
@@ -1,15 +1,32 @@
 import React from 'react';
-import { Outlet } from 'react-router-dom';
+import { Outlet, useLocation } from 'react-router-dom';
 import PageLayout from '../../common/components/PageLayout';
 import ReportsMenu from './ReportsMenu';
 
-const ReportsLayout = () => (
-  <PageLayout
-    menu={<ReportsMenu />}
-    breadcrumbs={['reportTitle', 'reportCombined']}
-  >
-    <Outlet />
-  </PageLayout>
-);
+const reportsPaths = {
+  '/reports/combined': 'reportCombined',
+  '/reports/route': 'reportRoute',
+  '/reports/event': 'reportEvents',
+  '/reports/trip': 'reportTrips',
+  '/reports/stop': 'reportStops',
+  '/reports/summary': 'reportSummary',
+  '/reports/chart': 'reportChart',
+  '/reports/logs': 'statisticsTitle',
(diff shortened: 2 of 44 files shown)


src/reports/components/ReportsLayout.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { Outlet } from 'react-router-dom';
+import PageLayout from '../../common/components/PageLayout';
+import ReportsMenu from './ReportsMenu';
+
+const ReportsLayout = () => (
+  <PageLayout
+    menu={<ReportsMenu />}
+    breadcrumbs={['reportTitle', 'reportCombined']}
```
tananaev, line 9: The breadcrumbs is different per report.

src/Navigation.jsx
```diff
@@ -154,7 +155,7 @@ const Navigation = () => {
           <Route path="user" element={<UserPage />} />
         </Route>
 
-        <Route path="reports">
+        <Route path="reports" element={<ReportsLayout />}>
```
tananaev, line 158: Should we do the same thing for settings?

tananaev:
Can you please clarify how this fixes the problem.

src/reports/components/ReportsLayout.jsx
jorgeart81: It's true, I hadn't noticed it.
I made these changes to fix it.

- 17ade9707e14bd04e071a59a610cd34c4e3780a2
- 8651b8dc0d29cb9c0709c1092eb01e6e2126332a

src/Navigation.jsx
jorgeart81: It seems to me that yes, in the components that have the same design.
It would be necessary to analyze it because I do not see that the state update generated by the socket is rendering the settings menu.

**I apologize for any translation errors, my language is Spanish and I am using a translator.**

jorgeart81:
Response to comment https://github.com/traccar/traccar-web/pull/1252#issuecomment-2197794442
##
Child components are affected by the rendering of the parent or container component.
In this case, when converting <PageLayout menu={<ReportsMenu />} Breadcrumbs={['reportTitle', 'reportCombined']}> to the parent container, the change of state of the children only affects their content, therefore the component main (Layout) is not affected.

src/Navigation.jsx
tananaev: Can you please update settings as well, so we have consistent codebase.

jorgeart81: Of course, let me see what I can do.

src/reports/components/ReportsLayout.jsx
```diff
@@ -0,0 +1,29 @@
...
+  '/reports/combined': ['reportTitle', 'reportCombined'],
+  '/reports/route': ['reportTitle', 'reportRoute'],
+  '/reports/event': ['reportTitle', 'reportEvents'],
+  '/reports/trip': ['reportTitle', 'reportTrips'],
+  '/reports/stop': ['reportTitle', 'reportStops'],
+  '/reports/summary': ['reportTitle', 'reportSummary'],
+  '/reports/chart': ['reportTitle', 'reportChart'],
+  '/reports/logs': ['reportTitle', 'statisticsTitle'],
+  '/reports/scheduled': ['settingsTitle', 'reportScheduled'],
+  '/reports/statistics': ['reportTitle', 'statisticsTitle'],
```
tananaev, line 16: You probably don't need to duplicate `'reportTitle'` everywhere. Just the second part is different.

jorgeart81: I made this change, what do you think? 395ada36c39d5a003b6e0b81242039cd4e62d571

jorgeart81: I found that being part was different->  '/reports/scheduled': ['settingsTitle', 'reportScheduled'],

tananaev: This is probably a mistake. All of them should be under reports.

src/settings/components/SettingsLayout.jsx
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
+import { Outlet, useLocation } from 'react-router-dom';
+import PageLayout from '../../common/components/PageLayout';
+import SettingsMenu from './SettingsMenu';
+
+const commondBreadcrumb = 'settingsTitle';
+
+const routes = {
+  settings: {
```
tananaev, line 9: Same here

src/settings/components/SettingsLayout.jsx
```diff
@@ -0,0 +1,82 @@
...
+    notifications: 'sharedNotifications',
+    notification: 'sharedNotification',
+    preferences: 'sharedPreferences',
+    server: 'settingsServer',
+    users: 'settingsUsers',
+    user: 'settingsUser',
+  },
+};
+
+const compoundRoutes = {
```
tananaev, line 35: This looks really complicated. Why do we need all of these? I think you should be able to split it by path segements.

jorgeart81: What do you think of this alternative?
https://gist.github.com/jorgeart81/6140d1afeab31ae690b11eedf8dca106

tananaev: Seems like it's still overcomplicated. Why can't we just append to an array? Something like:
```
let breadcrumbs = ['settingsTitle'];
if (length >= 2) {
  breadcrumbs.add(routes[pathSegmets[1]);
}
if (length >= 3) {
  switch ...
}
```

jorgeart81: I think that although it works for the current context, using [push ](https://developer.mozilla.org/es/docs/Web/JavaScript/Reference/Global_Objects/Array/push) would not be correct because push is used to add elements to an array, every time it is called.

I propose another solution so as not to affect the structure of your code too much. PR -> https://github.com/traccar/traccar-web/pull/1253

tananaev: The other PR looks promising, if it works as expected. Commented there.

tananaev:
We already have a solution for this. We only re-render when id or name changes.

PR #1253, opened 2024-07-02
fix: useMemo in menus

jorgeart81:
Prevent re-rendering of report and settings menus. 
This solution would maintain the structure of your project and would not be as complex as @ #1252

### DIFF
--- a/src/common/components/PageLayout.jsx
+++ b/src/common/components/PageLayout.jsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useMemo, useState } from 'react';
 import {
   AppBar,
   Breadcrumbs,
@@ -74,6 +74,8 @@ const PageLayout = ({ menu, breadcrumbs, children }) => {
 
   const [openDrawer, setOpenDrawer] = useState(false);
 
+  const menuMemo = useMemo(() => menu, []);
+
   return desktop ? (
     <div className={classes.desktopRoot}>
       <Drawer
@@ -88,7 +90,7 @@ const PageLayout = ({ menu, breadcrumbs, children }) => {
           <PageTitle breadcrumbs={breadcrumbs} />
         </Toolbar>
         <Divider />
-        {menu}
+        {menuMemo}
       </Drawer>
       <div className={classes.content}>{children}</div>
     </div>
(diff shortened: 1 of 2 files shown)


src/common/components/PageLayout.jsx
```diff
@@ -88,7 +90,7 @@ const PageLayout = ({ menu, breadcrumbs, children }) => {
           <PageTitle breadcrumbs={breadcrumbs} />
         </Toolbar>
         <Divider />
-        {menu}
+        {menuMemo}
```
tananaev, line 93: Shouldn't the same thing be done in mobile layout?

tananaev: Also, how does it actually prevent re-rendering? Because the menu is still passed as a parameter every time. Have you verified that it actually works?

jorgeart81: It does not render again because in this case the useMemo has an empty dependency [ ].
I share the video of the tests.

## Video
https://github.com/traccar/traccar-web/assets/92123185/bc492c2b-c44e-495d-892a-a7d52fdd9fb3

tananaev: The menu won't be re-rendered, but the page will, no?

I'm not sure I understand what I'm supposed to see on the video, but I see re-rendering around 4th second.

tananaev: But I don't understand what I'm supposed to see in the video.

jorgeart81: Previously all items in the report navigation menu were rendered with each socket update.   With useMemo it no longer happens.

tananaev: I thought the problem with re-rendering the map, not the menu. Why would re-rendering the menu a problem?

jorgeart81: I thought that was what you were referring to in the comment https://github.com/traccar/traccar-web/pull/1251#issuecomment-2197750734

I agree with you, menu rendering is not a problem.


Also thank you for your time, because these days I am understanding your platform better.

tananaev:
Rendering menu is cheap, so I don't think we need to over-optimize it.

PR #1254, opened 2024-07-03
Add option to view one attribute in the sidebar

Masaneth:
Add option to view speed of the device in the sidebar

With the option to activate it or not to each user's taste.

image

image

### DIFF
--- a/src/main/DeviceRow.jsx
+++ b/src/main/DeviceRow.jsx
@@ -63,7 +63,7 @@ const DeviceRow = ({ data, index, style }) => {
 
   const devicePrimary = useAttributePreference('devicePrimary', 'name');
   const deviceSecondary = useAttributePreference('deviceSecondary', '');
-  const deviceAttributeToShow = useAttributePreference('deviceShowAttribute', '');
+  const deviceShowAttribute = useAttributePreference('deviceShowAttribute', '');
 
   const secondaryText = () => {
     let status;
@@ -76,13 +76,13 @@ const DeviceRow = ({ data, index, style }) => {
       <>
         {deviceSecondary && item[deviceSecondary] && `${item[deviceSecondary]} • `}
         <span className={classes[getStatusColor(item.status)]}>{status}</span>
-        {position && position.hasOwnProperty(deviceAttributeToShow) && (
+        {position && position.hasOwnProperty(deviceShowAttribute) && (
           <>
             <span>{' • '}</span>
             <PositionValue
               position={position}
-              property={position.hasOwnProperty(deviceAttributeToShow) ? deviceAttributeToShow : null}
-              attribute={position.hasOwnProperty(deviceAttributeToShow) ? null : deviceAttributeToShow}
+              property={position.hasOwnProperty(deviceShowAttribute) ? deviceShowAttribute : null}
+              attribute={position.hasOwnProperty(deviceShowAttribute) ? null : deviceShowAttribute}
             />
           </>
         )}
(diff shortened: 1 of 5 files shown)


tananaev:
I don't think it makes sense to have a dedicated checkbox just for the speed. We should probably support any position attribute.

Masaneth:
I had not thought about the possibility to see any attribute, it is better more customizable, it is implemented in the second commit.

Now you can choose which attribute to display or not display any

image

image

src/resources/l10n/es.json
```diff
@@ -121,6 +121,7 @@
     "calendarFriday": "Viernes",
     "calendarSaturday": "Sábado",
     "attributeShowGeofences": "Mostrar Geo-Zonas",
+    "deviceShowAttribute": "Mostrar atributo en la Barra Lateral",
```
tananaev, line 124: You should only add English. Translations are done with Transifex.

src/settings/PreferencesPage.jsx
```diff
@@ -258,6 +258,25 @@ const PreferencesPage = () => {
...
+                    label={t('deviceShowAttribute')}
+                    value={attributes.deviceShowAttribute || ''}
+                    onChange={(e) => {
+                      setAttributes({ ...attributes, deviceShowAttribute: e.target.value });
+                    }}
+                  >
+                    <MenuItem value="">{'\u00a0'}</MenuItem>
+                    {Object.keys(positionAttributes).map((option) => (
+                      <MenuItem key={option} value={option}>
+                        <Typography component="span" color="textPrimary">
```
tananaev, line 273: Why do we need `Typography` here?

src/settings/PreferencesPage.jsx
```diff
@@ -258,6 +258,25 @@ const PreferencesPage = () => {
                   titleGetter={(it) => t(it.name)}
                   label={t('deviceSecondaryInfo')}
                 />
+                <FormControl>
+                  <InputLabel>{t('deviceShowAttribute')}</InputLabel>
+                  <Select
+                    label={t('deviceShowAttribute')}
+                    value={attributes.deviceShowAttribute || ''}
+                    onChange={(e) => {
+                      setAttributes({ ...attributes, deviceShowAttribute: e.target.value });
```
tananaev, line 267: Is there an easy way to clear this attribute when empty value is selected?

src/main/DeviceRow.jsx
```diff
@@ -62,6 +63,7 @@ const DeviceRow = ({ data, index, style }) => {
 
   const devicePrimary = useAttributePreference('devicePrimary', 'name');
   const deviceSecondary = useAttributePreference('deviceSecondary', '');
+  const deviceAttributeToShow = useAttributePreference('deviceShowAttribute', '');
```
tananaev, line 66: You use different names in different places:
- `deviceShowAttribute`
- `deviceAttributeToShow`
- `deviceAttribute`

Let's unify this, probably to `deviceShowAttribute`.

src/settings/PreferencesPage.jsx
Masaneth: In the existing code of the page, attributes are normally not removed when they are set with empty values; they are kept with an empty string or a default value. To maintain consistency, I manage empty values in a similar way:

tananaev: OK, I think it's ok to keep as is. It's a bit different though because most of the other attributes are basically expected to have something. This one is basically optional.

Issue #1245, opened 2024-06-19
Show version number on the initial login page


jacobburrell:
**Is your feature request related to a problem? Please describe.**
It should be readily apparent what version of Traccar a server is running.

**Describe the solution you'd like**
Add a simple label at the bottom right-hand corner showing the version of traccar

**Describe alternatives you've considered**
Requiring the user to login or only show the information to admins, making the information private by default with the setting/option to display it.

**Additional context**
<img width="1680" alt="image" src="https://github.com/traccar/traccar-web/assets/61520543/99fae235-3b2c-40c6-88cb-3f652293d471">

tananaev:
Why do you need the version number?

dough29:
You should never tell bad people what version of software you are running

tananaev:
Our software is open source, so it's trivial to find out the version for someone who knows what they're doing, even if we don't include the number.

harshanasrimal:
@jacobburrell I hope this implementation works for you.

PR #1256, opened 2024-07-09
Show version on login screen

harshanasrimal:
Implemented the system to show version number on the initial authentication pages

Regarding the feature requested #1245


tananaev:
I'm personally against adding this. I don't see enough value in it.

harshanasrimal:
@tananaev I thought its nothing big deal cause there is an option to disable it from the config.

tananaev:
It adds another thing to maintain and worry about. Even if it's small, we need to carefully think about everything we add.

harshanasrimal:
Mm. It's up to you. I don't think that will be a big deal to maintain this. It's a simple UI change.

harshanasrimal:
@tananaev what do you think? do I need to withdraw my PR?

tananaev:
We can keep it open and see how much interest is there.

PR #1260, opened 2024-08-04
I think this way is easier to understand the trip start end points on the map

jcardus:
Captura de ecrã 2024-08-04, às 13 40 14


jcardus:
I also think we should change the color, wouldn't it be better to use green for start and red for end?

PR #1263, opened 2024-08-23
maximumFileSizeToCacheInBytes to 5Mb

jcardus:
The main javascript file is not being cached by the service worker be cause it passes the 4Mb limit.

The problem here is the server updates. If you update the server with a new build, the page won't open because the service worker will try to fetch and old javascript file which doesn't exist anymore. 

It can be stuck forever. For it to update you have to shutdown the browser (not just refresh the page or close the tab/window) so that all clients release the active service worker. 

If we cache the main javascript we won't have this problem, the page will open and the popup for the update will appear. 

Captura de ecrã 2024-08-23, às 23 33 18

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -22,7 +22,7 @@ export default defineConfig(() => ({
       includeAssets: ['favicon.ico', 'apple-touch-icon-180x180.png'],
       workbox: {
         navigateFallbackDenylist: [/^\/api/],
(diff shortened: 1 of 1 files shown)


tananaev:
I say we pick a higher number, like 10MB. Also let's make it proper 10MB (10 * 1024 * 1024).

PR #1264, opened 2024-08-24
Update pt_BR.json

jcardus:
"Definir tempo" doesn't make sense.

### DIFF
--- a/legacy/web/l10n/pt_BR.json
+++ b/legacy/web/l10n/pt_BR.json
@@ -247,7 +247,7 @@
     "linkGoogleMaps": "Google Maps",
     "linkAppleMaps": "Apple Maps",
     "linkStreetView": "Street View",
-    "positionFixTime": "Definir tempo",
+    "positionFixTime": "Hora",
     "positionDeviceTime": "Hora do dispositivo",
     "positionServerTime": "Hora do servidor",
     "positionValid": "Válido",


legacy/web/l10n/pt_BR.json
```diff
@@ -247,7 +247,7 @@
     "linkGoogleMaps": "Google Maps",
     "linkAppleMaps": "Apple Maps",
     "linkStreetView": "Street View",
-    "positionFixTime": "Definir tempo",
+    "positionFixTime": "Hora",
```
tananaev, line 250: This doesn't look like the right translation.

tananaev (commented):
Also translations have to do done through Transifex.

legacy/web/l10n/pt_BR.json
jcardus: I just put hour because I can't find an equivalent word to fix in this case. Maybe "Hora da posição" which means Position time.

I never heard of transifex, I'll check it out.

tananaev: There are 3 time parameters in Traccar. Each one is slightly different, so we need to make sure they are consistent and reflect what they actually mean.

jcardus: I tried to translate it to "Hora GPS" in transifex, don't know if anything happened.

tananaev: We sync it up every release.

PR #1273, opened 2024-09-23
Add navigation app link in status card

dough29:
Hello,

I often need to reach the position of my trackers but I prefer to use Waze.

Before, I had to open the link for Google Maps, copy latlon coordinates, open Waze manually and paste the coordinates.

With this query we add a "Waze" entry to the status card menu:

image

This will trigger the Waze app to open on the smartphone.

I have successfully tested this locally with an Android phone.

### DIFF
--- a/src/common/components/StatusCard.jsx
+++ b/src/common/components/StatusCard.jsx
@@ -126,6 +126,11 @@ const StatusCard = ({ deviceId, position, onClose, disableActions, desktopPaddin
   const positionAttributes = usePositionAttributes(t);
   const positionItems = useAttributePreference('positionItems', 'fixTime,address,speed,totalDistance');
 
+  const serverNavigationAppLink = useSelector((state) => state.session.server.attributes.navigationAppLink);
+  const serverNavigationAppTitle = useSelector((state) => state.session.server.attributes.navigationAppTitle);
+  const userNavigationAppLink = user.attributes.navigationAppLink;
+  const userNavigationAppTitle = user.attributes.navigationAppTitle;
+
   const [anchorEl, setAnchorEl] = useState(null);
 
   const [removing, setRemoving] = useState(false);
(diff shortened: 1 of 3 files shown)


tananaev:
I don't think it makes sense to hardcode something like this. Waze is not ubiquitous enough. Maybe a better option is a customizable menu option there where you can provide a custom URL with location placeholder.

btw you should only add English string.

dough29:
Thanks for your quick review.

Ok, I can spend a little more time making it customizable 👍

dough29:
Could you give me an idea how I could interpret a specific format in the attribute string so that I can replace it with latitude and longitude values ​​🤔

I should create custom user and server attributes like:
- state.session.server.attributes.navigationAppLink
- state.session.server.attributes.navigationAppTitle
- state.session.user.navigationAppLink
- state.session.user.navigationAppTitle

navigationAppLink should have a value like "https://www.waze.com/ul?ll=${position.latitude}%2C${position.longitude}" but I don't know how to handle the position.latitude and position.longitude in the string so that they are replaced with their value.

dough29:
Ok, navigationAppLink and navigationAppTitle should be in server attributes or only for user attributes?

tananaev:
They should be in `useCommonUserAttributes.js`. That way they will apply to both server and user.

src/common/components/StatusCard.jsx
```diff
@@ -126,6 +126,11 @@ const StatusCard = ({ deviceId, position, onClose, disableActions, desktopPaddin
   const positionAttributes = usePositionAttributes(t);
   const positionItems = useAttributePreference('positionItems', 'fixTime,address,speed,totalDistance');
 
+  const serverNavigationAppLink = useSelector((state) => state.session.server.attributes.navigationAppLink);
+  const serverNavigationAppTitle = useSelector((state) => state.session.server.attributes.navigationAppTitle);
+  const userNavigationAppLink = user.attributes.navigationAppLink;
+  const userNavigationAppTitle = user.attributes.navigationAppTitle;
```
tananaev, line 132: You should use `useAttributePreference` instead of doing this.

src/common/components/StatusCard.jsx
```diff
@@ -269,6 +274,8 @@ const StatusCard = ({ deviceId, position, onClose, disableActions, desktopPaddin
           <MenuItem component="a" target="_blank" href={`https://www.google.com/maps/search/?api=1&query=${position.latitude}%2C${position.longitude}`}>{t('linkGoogleMaps')}</MenuItem>
           <MenuItem component="a" target="_blank" href={`http://maps.apple.com/?ll=${position.latitude},${position.longitude}`}>{t('linkAppleMaps')}</MenuItem>
           <MenuItem component="a" target="_blank" href={`https://www.google.com/maps/@?api=1&map_action=pano&viewpoint=${position.latitude}%2C${position.longitude}&heading=${position.course}`}>{t('linkStreetView')}</MenuItem>
+          {serverNavigationAppLink && serverNavigationAppTitle && <MenuItem component="a" target="_blank" href={serverNavigationAppLink.replace('{latitude}', position.latitude).replace('{longitude}', position.longitude)}>{serverNavigationAppTitle}</MenuItem>}
+          {userNavigationAppLink && userNavigationAppTitle && <MenuItem component="a" target="_blank" href={userNavigationAppLink.replace('{latitude}', position.latitude).replace('{longitude}', position.longitude)}>{userNavigationAppTitle}</MenuItem>}
```
tananaev, line 278: I would just check one of these attributes instead of both. Maybe title.

src/common/components/StatusCard.jsx
dough29: Using `useAttributePreference`, if there is a server attribute set, we cannot display the one set in user attributes.

image

src/resources/l10n/en.json
```diff
@@ -609,5 +609,7 @@
     "categoryVan": "Van",
     "categoryScooter": "Scooter",
     "maintenanceStart": "Start",
-    "maintenancePeriod": "Period"
+    "maintenancePeriod": "Period",
+    "navigationAppLink": "Navigation app link",
+    "navigationAppTitle": "Navigation app title"
```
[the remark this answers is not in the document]
dough29: Fixec ✅

src/common/components/StatusCard.jsx
dough29: I think we should let users decide whether they want to add the navigation link attributes for the whole server, only for some users or both, any opinion?

dough29: Ok cool, I just tested this so if there are attributes defined on the server side, they will be overridden if the user defines them on their side 👍

src/common/components/StatusCard.jsx
dough29: Fixed ✅

src/common/components/StatusCard.jsx
dough29: Fixed ✅

tananaev:
Merged, thanks.

PR #1277, opened 2024-10-07
Fix Color normalization in Replay

Kalabint:
_Removed hardcoded 0 in minSpeed, its now fully dynamic._

Before:

image

After:

image

### DIFF
--- a/src/common/util/colors.js
+++ b/src/common/util/colors.js
@@ -23,10 +23,6 @@ const interpolateTurbo = (value) => {
 };
 
 const getSpeedColor = (speed, minSpeed, maxSpeed) => {
-  if (maxSpeed <= minSpeed) {
-    throw new Error('maxSpeed must be greater than minSpeed');
-  }
-
   const normalizedSpeed = (speed - minSpeed) / (maxSpeed - minSpeed);
 
   const [r, g, b] = interpolateTurbo(normalizedSpeed);
--- a/src/common/util/colors.js
+++ b/src/common/util/colors.js
@@ -22,9 +22,15 @@ const interpolateTurbo = (value) => {
   ];
 };
 
-const getSpeedColor = (speed, maxSpeed) => {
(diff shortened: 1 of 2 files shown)


src/common/util/colors.js
```diff
@@ -22,9 +22,15 @@ const interpolateTurbo = (value) => {
   ];
 };
 
-const getSpeedColor = (speed, maxSpeed) => {
-  const normalizedSpeed = Math.max(0, Math.min(1, speed / maxSpeed));
+const getSpeedColor = (speed, minSpeed, maxSpeed) => {
+  if (maxSpeed <= minSpeed) {
+    throw new Error('maxSpeed must be greater than minSpeed');
+  }
```
tananaev, line 28: Why do we need this check?

src/map/MapRoutePoints.js
```diff
@@ -57,7 +57,11 @@ const MapRoutePoints = ({ positions, onClick }) => {
   }, [onMarkerClick]);
 
   useEffect(() => {
+    if (positions.length === 0) return;
```
tananaev, line 60: And this one also.

src/common/util/colors.js
Kalabint: We don't. The vars get fed correctly in the step before, and user input does not occur.

I rushed too fast once i spotted the error, this for both of them.

Kalabint:
I just realised, that the getSpeedColor is also used in [MapRoutePath.js](https://github.com/Kalabint/traccar-web/blob/bugfix/fix-color-normalization/src/map/MapRoutePath.js), which is currently not adjusted to the new function call.

**Because of that the route line stays black in the trips menu.**
image


I already updated my [Kalabint:bugfix/fix-color-normalization](https://github.com/Kalabint/traccar-web/tree/bugfix/fix-color-normalization) branch to include the fix.

image


Should i open a new PR or what is the best way to go forward?

tananaev:
Yes, please open a new PR.

PR #1280, opened 2024-10-10
Better icon for trip finish

jcardus:
Users sometimes get confused when looking at the trip route on the map, trying to figure out the start and end point.

I think this icon helps.

before:
Screenshot 2024-10-10 at 16 02 42

after:
Screenshot 2024-10-10 at 16 02 47


tananaev:
If we do something like this, I think we need both start and finish icons. Also make sure they are consistent in terms of padding and style.

jcardus:
I see the icons are 24x24, what do we use for the padding size?

tananaev:
I believe we use material icons as is.

jcardus:
With 1px padding and 2 icons.

Oh I didn't kown this was material icons, I will try with those.

Screenshot 2024-10-10 at 16 57 24

jcardus:
with material icons:

Screenshot 2024-10-10 at 17 05 13

tananaev:
Looks better, but why only one has a color?

tananaev:
There has to be a symmetry. You can have one green and one neutral. You either make green (start) and red (stop), like a traffic light. Or you make them both neutral. Or something along those lines.

jcardus:
Green and Red:

Screenshot 2024-10-10 at 17 51 52

Issue #1272, opened 2024-09-22
Geofences in alphabetical order


Turbovix:
**Is your feature request related to a problem? Please describe.**
No, my request is not related to any problem. However, manipulating geofences is difficult when you have a lot of them, since they are not organized in alphabetical order.

**Describe the solution you'd like**
I find that if the geofences are sorted in alphabetical order it makes them easier to find.

![Captura de tela de 2024-09-20 00-44-48](https://github.com/user-attachments/assets/fd24c9cc-999b-4207-937b-4802526c8279)

Turbovix:
Dear all, when you make this improvement, please do so with the saved commands.
Thank you.

tananaev:
We're already doing server side sorting.

PR #1289, opened 2024-10-31
Add Sorting Functionality for Selection Parameters 

CpdnCristiano:
This change improves usability, especially when there are many devices or users, as finding specific entries becomes challenging without sorting options. With this functionality, Traccar users can quickly and easily locate devices or users in large lists.

### DIFF
--- a/src/common/components/LinkField.jsx
+++ b/src/common/components/LinkField.jsx
@@ -21,7 +21,9 @@ const LinkField = ({
     if (active) {
       const response = await fetch(endpointAll);
       if (response.ok) {
-        setItems(await response.json());
+        const data = await response.json();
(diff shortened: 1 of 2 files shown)


tananaev:
I think sorting it on the server side probably makes more sense.

CpdnCristiano:
I can do this. Should I classify by default, or should I create a filter, for example: `devices?sort=name`?

tananaev:
It's probably fine to sort by default. This should be done through the database query.

CpdnCristiano:
I will implement sorting on all endpoints, and issue #1272 requested sorting of groups. Then, I will make a PR.

tananaev:
I would recommend starting with one endpoint, so we can review the implementation before applying it to everything.

PR #1291, opened 2024-10-31
Textsize to 14 and use Bold.

jcardus:
I think this way the labels are more readable.
I also tried to change Open Sans and Arial Unicode to bold but it looks like it's not hosted on the cdn.

before:
Captura de ecrã 2024-10-31, às 13 54 34

after:
Captura de ecrã 2024-10-31, às 13 53 40

### DIFF
--- a/src/common/components/PositionValue.jsx
+++ b/src/common/components/PositionValue.jsx
@@ -93,7 +93,7 @@ const PositionValue = ({ position, property, attribute }) => {
     }
   };
 
-  if (value === undefined) {
+  if (value === undefined || value === null) {
     return '';
   }
 


tananaev:
What if you have 100 devices?

jcardus:
before:
Captura de ecrã 2024-10-31, às 16 23 56

after:
Captura de ecrã 2024-10-31, às 16 22 13

tananaev:
I like the before version.

I also see that you changed font to Noto Sans Bold. But we also use some other fonts. Which means it will be inconsistent.

jcardus:
I know, that's why I said: "I also tried to change Open Sans and Arial Unicode to bold but it looks like it's not hosted on the cdn"

tananaev:
Yeah, but then it doesn't make sense to change only one.

tananaev:
I recommend submitting a feature request and we see what people want. Again, I personally prefer smaller font.

PR #1293, opened 2024-11-03
Display navigation directions from MapBox

RafailStratiotis:
Display navigation directions from MapBox

But the route/navigation line doesn't work..
Can some one help for this?

### DIFF
--- a/src/map/MapPositions.js
+++ b/src/map/MapPositions.js
@@ -8,6 +8,9 @@ import { mapIconKey } from './core/preloadImages';
 import { useAttributePreference } from '../common/util/preferences';
 import { useCatchCallback } from '../reactHelper';
 import { findFonts } from './core/mapUtil';
+import MapboxDirections from '@mapbox/mapbox-gl-directions/dist/mapbox-gl-directions';
+import '@mapbox/mapbox-gl-directions/dist/mapbox-gl-directions.css';
+import mapboxgl from 'mapbox-gl';
 
 const MapPositions = ({ positions, onClick, showStatus, selectedPosition, titleField }) => {
   const id = useId();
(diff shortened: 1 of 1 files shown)


jinzo:
I'm not sure, but I would say that Mapbox probably does not like Maplibre GL JS fork of Mapbox GL JS that is used in Traccar. I would try with this: https://github.com/maplibre/maplibre-gl-directions As it supports "Mapbox Directions API" and OSRM - so I think it would be a good fit.

Not sure what @tananaev thinks about this feature, but I would be interested in it - if it would be based on maplibre-gl-directions, so I can use OSRM for it (or something else that is open or has configurable backends).

tananaev:
Definitely want to see a flexible solution. Ideally we support a large number of providers.

RafailStratiotis:
@jinzo I noticed the solution you thought of but it doesn't cover me, basically I want to see how long time it will take the vehicle to arrive at a specific destination but in the code i tried to generate it doesn't show the route/navigation line like in mapbox demo [https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-directions/](https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-directions/)

@tananaev can you help in some way?

tananaev:
What's the goal here? If you just want a custom implementation just for yourself, this is not the right place. If you want to contribute something, we have discuss it first and it has to be flexible enough.

RafailStratiotis:
@tananaev I'm here to contribute.. the solution proposed by @jinzo I think it does not help and does not cover my reasoning.
my rationale is : with left click a select some vehicle and right click on map select the destination it will end the car and see how many hours it takes for it to reach the destination.

I am offering an unfinished code that can be used by many maps besides mapbox, if you have something better to recommend it would be appreciated

jinzo:
@RafailStratiotis maplibre-gl-directions is probably an old(er) fork of mapbox-gl-directions (that also offers some nice controls), but I would say that there are high chance that the low level controls/code is there.
Points where I would start:
- https://maplibre.org/maplibre-gl-directions/#/examples/mapbox-directions-api-and-congestions
- https://github.com/maplibre/maplibre-gl-directions/blob/53e93ef28cd5726f19cf7c7681f24e929d3cbb17/src/directions/main.ts#L22
I did a quick check and both examples do the exactly same request, also getting the same data (namely `duration` field). You just need to figure out how to access that from the `MapLibreGlDirections` - currently I don't have more time to dig into this.

My point being, that it's probably a lot more feasible to make this work with maplibre-gl-directions than to get mapbox-gl-directions working with Traccar (without substantial fork).

PR #1294, opened 2024-11-04
Add speed color legend

jcardus:
Screenshot 2024-11-04 at 17 00 47


jcardus:
Screenshot 2024-11-04 at 16 50 07

tananaev:
I think this is overkill. I don't want to add more controls just for speed.

jcardus:
It goes out when we leave route replay

tananaev:
I understand, but still it takes a lot of space.

jcardus:
any suggestion?

tananaev:
I think more reasonable would be to maybe show some markers on the map to indicate speed. Maybe show it on hover only.

jcardus:
This way you get a quick idea of the speeding (where and how much). Maybe I can put it smaller.

jcardus:
Smaller version...


localhost_3000_(iPhone 12 Pro) (1)

tananaev:
Looks better. Can we make text the same style as the distance control? We can also probably round to 1 or even 0 digits. And maybe we don't need the middle point. Then we can reduce the width even more.

jcardus:
here it goes
localhost_3000_(iPhone 12 Pro) (1)

src/map/legend/MapSpeedLegend.js
```diff
@@ -0,0 +1,56 @@
...
+
+    const legend = document.createElement('div');
+
+    const colorBar = document.createElement('div');
+    colorBar.style.background = `linear-gradient(to right, ${gradientStops})`;
+    colorBar.style.height = '10px';
+
+    const speedLabel = document.createElement('span');
+    speedLabel.textContent = `${Math.round(speedFromKnots(this.minSpeed, this.speedUnit))} - ${
+      Math.round(speedFromKnots(this.maxSpeed, this.speedUnit))} ${speedUnitString(this.speedUnit, this.t)}`;
```
tananaev, line 47: Maybe split these calculations into separate variables for readability.

src/map/legend/MapSpeedLegend.js
```diff
@@ -0,0 +1,56 @@
+import { interpolateTurbo } from '../../common/util/colors';
+import { speedFromKnots, speedUnitString } from '../../common/util/converter';
+
+export class LegendControl {
```
tananaev, line 4: Probably better to call it `SpeedLegendControl`.

tananaev:
Looks good, thank you.

hypoclone:
Hello @jcardus and @tananaev, thank you guys for the great work you do. i don't know if it is ok to discuss it here or create another discussion on it, but i think it will be intuitive to have the replay track show stops and stops duration with an icon i.e

https://github.com/user-attachments/assets/b23b34df-78dd-484a-b26e-7fdd75c3a1f2

jcardus:
Maybe it's better to create a feature request, probably for the combined report which already has some of that information.

PR #1295, opened 2024-11-10
Optimize drawer width to reduce wasted space in reports and settings

jcardus:
before:
Captura de ecrã 2024-11-10, às 12 56 40

after:
Captura de ecrã 2024-11-10, às 12 56 56

before:
Captura de ecrã 2024-11-10, às 13 25 00

after:
Captura de ecrã 2024-11-10, às 13 24 49

before:
Captura de ecrã 2024-11-10, às 13 27 34

after:
Captura de ecrã 2024-11-10, às 13 27 18

Captura de ecrã 2024-11-10, às 13 26 45

### DIFF
--- a/src/common/components/MenuItem.jsx
+++ b/src/common/components/MenuItem.jsx
@@ -1,0 +1,22 @@
+import makeStyles from '@mui/styles/makeStyles';
+import { ListItemButton, ListItemIcon, ListItemText } from '@mui/material';
+import { Link } from 'react-router-dom';
+import React from 'react';
+
+const MenuItem = ({
+  title, link, icon, selected,
+}) => {
+  const classes = makeStyles({
+    menuItemText: {
+      whiteSpace: 'nowrap',
+    },
+  });
+  return (
+    <ListItemButton key={link} component={Link} to={link} selected={selected}>
+      <ListItemIcon>{icon}</ListItemIcon>
+      <ListItemText primary={title} className={classes.menuItemText} />
+    </ListItemButton>
+  );
+};
+
+export default MenuItem;
--- a/src/common/components/MenuItem.jsx
+++ b/src/common/components/MenuItem.jsx
@@ -3,14 +3,14 @@ import { ListItemButton, ListItemIcon, ListItemText } from '@mui/material';
 import { Link } from 'react-router-dom';
 import React from 'react';
 
-const MenuItem = ({
-  title, link, icon, selected,
-}) => {
-  const classes = makeStyles({
-    menuItemText: {
-      whiteSpace: 'nowrap',
-    },
-  });
+const useStyles = makeStyles(() => ({
+  menuItemText: {
+    whiteSpace: 'nowrap',
+  },
+}));
+
+const MenuItem = ({ title, link, icon, selected }) => {
+  const classes = useStyles();
   return (
     <ListItemButton key={link} component={Link} to={link} selected={selected}>
       <ListItemIcon>{icon}</ListItemIcon>
--- a/src/common/components/PageLayout.jsx
+++ b/src/common/components/PageLayout.jsx
@@ -29,7 +29,7 @@ const useStyles = makeStyles((theme) => ({
     flexDirection: 'column',
   },
   desktopDrawer: {
-    width: (props) => (props.miniVariant ? `calc(${theme.spacing(7)} + 1px)` : theme.dimensions.drawerWidthDesktop),
+    width: (props) => (props.miniVariant ? `calc(${theme.spacing(8)} + 1px)` : theme.dimensions.drawerWidthDesktop),
     transition: theme.transitions.create('width', {
       easing: theme.transitions.easing.sharp,
       duration: theme.transitions.duration.enteringScreen,
--- a/src/common/components/PageLayout.jsx
+++ b/src/common/components/PageLayout.jsx
@@ -29,7 +29,7 @@ const useStyles = makeStyles((theme) => ({
     flexDirection: 'column',
   },
   desktopDrawer: {
-    width: (props) => (props.miniVariant ? theme.dimensions.miniDrawerWidthDesktop : theme.dimensions.drawerWidthDesktop),
+    width: (props) => (props.miniVariant ? `calc(${theme.spacing(7)} + 1px)` : theme.dimensions.drawerWidthDesktop),
     transition: theme.transitions.create('width', {
       easing: theme.transitions.easing.sharp,
       duration: theme.transitions.duration.enteringScreen,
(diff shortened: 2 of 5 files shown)


jcardus:
before:
Captura de ecrã 2024-11-10, às 12 56 40

after:
Captura de ecrã 2024-11-10, às 12 56 56

before:
Captura de ecrã 2024-11-10, às 13 25 00

after:
Captura de ecrã 2024-11-10, às 13 24 49

before:
Captura de ecrã 2024-11-10, às 13 27 34

after:
Captura de ecrã 2024-11-10, às 13 27 18

Captura de ecrã 2024-11-10, às 13 26 45

tananaev:
I don't think it looks good visually.

src/common/components/PageLayout.jsx
```diff
@@ -27,7 +27,8 @@ const useStyles = makeStyles((theme) => ({
     flexDirection: 'column',
   },
   desktopDrawer: {
-    width: theme.dimensions.drawerWidthDesktop,
+    maxWidth: theme.dimensions.drawerWidthDesktop,
+    position: 'relative !important',
```
tananaev, line 31: btw why do we need this?

jcardus: otherwise the right content will use the full width and the wrapper will float on top of it.

jcardus: Emulator page also puts "position: relative".
For the !important, don't know why but it doesn't work without it.

jcardus:
Another option would be to add a collapse button to show only the icons. It would help a lot, especially when you start adding columns to reports.

tananaev:
Report columns are usually configurable.

jcardus:
Yes I known, what I mean is that we quickly get an horizontal scroll bar when we start adding columns to a report.

tananaev:
Would be great if we can make them manually adjustable, but current size by default.

tananaev:
I guess it works if it looks good.

jcardus:
ok, I'll work on that

jcardus:
https://github.com/user-attachments/assets/cc840c8c-304d-4f13-8feb-777139d23dbc

jcardus:
https://github.com/user-attachments/assets/9412ad6d-d349-4deb-8ed5-4e088edce429

src/reports/components/ReportsMenu.jsx
```diff
@@ -22,7 +22,7 @@ const MenuItem = ({
 }) => (
   <ListItemButton key={link} component={Link} to={link} selected={selected}>
     <ListItemIcon>{icon}</ListItemIcon>
-    <ListItemText primary={title} />
+    <ListItemText primary={title} sx={{ whiteSpace: 'nowrap' }} />
```
tananaev, line 25: What is this for? Also probably it should be set as a class.

src/common/components/PageLayout.jsx
```diff
@@ -66,14 +72,19 @@ const PageTitle = ({ breadcrumbs }) => {
...
+  const [miniVariant, setMiniVariant] = useState(false);
+  const classes = useStyles({ miniVariant });
   const theme = useTheme();
   const navigate = useNavigate();
 
   const desktop = useMediaQuery(theme.breakpoints.up('md'));
 
   const [openDrawer, setOpenDrawer] = useState(false);
 
+  const toggleDrawer = () => {
```
tananaev, line 84: I think this can be inlined.

src/common/components/PageLayout.jsx
```diff
@@ -82,10 +93,17 @@ const PageLayout = ({ menu, breadcrumbs, children }) => {
...
-            <ArrowBackIcon />
+          {!miniVariant && (
+          <>
+            <IconButton color="inherit" edge="start" sx={{ mr: 2 }} onClick={() => navigate('/')}>
+              <ArrowBackIcon />
+            </IconButton>
+            <PageTitle breadcrumbs={breadcrumbs} />
+          </>
+          )}
+          <IconButton color="inherit" edge="start" sx={{ ml: miniVariant ? -2 : 'auto' }} onClick={toggleDrawer}>
```
tananaev, line 104: I think this needs to be reformatted.

src/common/theme/dimensions.js
```diff
@@ -2,6 +2,7 @@ export default {
   sidebarWidth: '28%',
   sidebarWidthTablet: '52px',
   drawerWidthDesktop: '360px',
+  miniDrawerWidthDesktop: '56px',
```
tananaev, line 5: Is there some standard size we can use from MUI?

src/common/components/PageLayout.jsx
jcardus: which part?

jcardus: got it

src/common/components/PageLayout.jsx
```diff
@@ -27,7 +29,11 @@ const useStyles = makeStyles((theme) => ({
     flexDirection: 'column',
   },
   desktopDrawer: {
-    width: theme.dimensions.drawerWidthDesktop,
+    width: (props) => (props.miniVariant ? `calc(${theme.spacing(7)} + 1px)` : theme.dimensions.drawerWidthDesktop),
```
tananaev, line 32: `calc(${theme.spacing(7)} + 1px)` doesn't seem right. I know there is a version of a collapsed sidebar. Can we somehow use the size from there?

tananaev: Or maybe even use it as is instead of implementing our own thing.

https://mui.com/material-ui/react-drawer/#mini-variant-drawer

jcardus: That's exactly where that was taken from

jcardus: It's not a component, it's an example how to do the mini variant.

tananaev: Oh I didn't realize that.

src/common/components/PageLayout.jsx
tananaev: I think we should probably apply classes here as well for readability.

src/settings/components/SettingsMenu.jsx
```diff
@@ -17,20 +17,30 @@ import HelpIcon from '@mui/icons-material/Help';
...
 } from '../../common/util/permissions';
 import useFeatures from '../../common/util/useFeatures';
 
+const useStyles = makeStyles({
+  menuItemText: {
+    whiteSpace: 'nowrap',
+  },
+});
+
 const MenuItem = ({
```
tananaev, line 33: Maybe we can extract it as a common component to avoid duplication?

src/common/components/PageLayout.jsx
jcardus: I followed the same pattern as the upper IconButton

src/common/components/MenuItem.jsx
```diff
@@ -0,0 +1,22 @@
...
+import React from 'react';
+
+const MenuItem = ({
+  title, link, icon, selected,
+}) => {
+  const classes = makeStyles({
+    menuItemText: {
+      whiteSpace: 'nowrap',
+    },
+  });
```
tananaev, line 13: Can you please move this outside for consistency. Similar to how we have it in other places.

PR #1300, opened 2024-11-24
Add Tool for Converting Traccar Geofences to KML Format

ferryops:
### Summary
This PR introduces a tool to convert Traccar geofence data into .kml format, making it compatible with mapping and geofencing tools that support KML imports. The tool fetches geofence data from Traccar (e.g., through the API endpoint /api/geofences) and translates the polygon areas into KML-compatible coordinates.

### Purpose
This conversion tool was successfully tested to import Traccar geofences into the VSS (Vehicle Service System) platform by Howen. It enables a streamlined process for transferring geofences between Traccar and other platforms that support KML files, such as mapping and GPS monitoring applications.

### Implementation Details

- Coordinate Parsing: The tool includes a helper function that parses and reverses coordinate pairs from Traccar's POLYGON data format to the KML-required format.
- KML Structure: It generates KML-compliant XML, including style attributes like PolyStyle and LineStyle, to define colors for polygons.
- File Output: The generated KML data is saved to geofences.kml, ready for import into KML-compatible platforms.

### Example Usage
To run the conversion tool, execute the following command:

`node /tools/geofences-to-kml.js`


tananaev:
I think it's better to have a python tool. It also looks like the inputs and outputs are hardcoded.

ferryops:
> I think it's better to have a python tool. It also looks like the inputs and outputs are hardcoded.

Thank you for your advice, I will try rewriting this tool to Python and minimizing the hardcode

ferryops:
done, tools already rewrite from javascript to python

PR #1311, opened 2024-12-03
Sort device categories and attributes

jcardus:


jcardus:
Any comments on this?

tananaev:
Need some description.

jcardus:
Device categories on the device page are not being sorted. 

The filter for making sure name exists it's because there can be categories not yet translated in some languages.

tananaev:
I guess sorting kind of makes sense, but I don't think filtering does.

jcardus:
Yes, but then the sort will receive undefined values.
Should we put the category key (without 'category') in the name when the translation returns undefined?

tananaev:
Why are there undefined values?

jcardus:
I guess it can happen in the master branch before the Traccar bot has updated the translations.

src/settings/components/AddAttributeDialog.jsx
```diff
@@ -54,7 +54,7 @@ const AddAttributeDialog = ({ open, onResult, definitions }) => {
             }
             return filtered;
           }}
-          options={options}
+          options={options.sort((a, b) => a.name.localeCompare(b.name))}
```
tananaev, line 57: Wouldn't it make more sense to do it at the definition? Also probably makes sense to memo it.

PR #1314, opened 2024-12-20
Update ru.json

bc38:
RU localization of untranslated strings


tananaev:
For localization you should be using Transifex:

https://explore.transifex.com/traccar/traccar/

bc38:
> For localization you should be using Transifex:
> 
> https://explore.transifex.com/traccar/traccar/

Supplemented the translation

PR #1316, opened 2024-12-26
Update ChartReportPage.jsx

hypoclone:
This PR add pan to zoom and show dot only on mouse hover for the chart in the report page.

### DIFF
--- a/src/reports/ChartReportPage.jsx
+++ b/src/reports/ChartReportPage.jsx
@@ -4,7 +4,7 @@ import {
   FormControl, InputLabel, Select, MenuItem, useTheme,
 } from '@mui/material';
 import {
-  Brush, ReferenceArea, CartesianGrid, Line, LineChart, ResponsiveContainer, Tooltip, XAxis, YAxis,
+  Brush, CartesianGrid, Line, LineChart, ResponsiveContainer, Tooltip, XAxis, YAxis,
 } from 'recharts';
 import ReportFilter from './components/ReportFilter';
 import { formatTime } from '../common/util/formatter';
@@ -21,7 +21,7 @@ import useReportStyles from './common/useReportStyles';
 
 const ChartReportPage = () => {
   const [zoomDomain, setZoomDomain] = useState({ start: 'dataMin', end: 'dataMax' });
-  const handleBrushChange = (range) => { if (range) { setZoomDomain({ start: range.startIndex, end: range.endIndex }); }};
+  const handleBrushChange = (range) => { if (range) { setZoomDomain({ start: range.startIndex, end: range.endIndex }); } };
   
   const classes = useReportStyles();
   const theme = useTheme();
--- a/src/reports/ChartReportPage.jsx
+++ b/src/reports/ChartReportPage.jsx
@@ -20,7 +20,7 @@ import {
 import useReportStyles from './common/useReportStyles';
 
 const ChartReportPage = () => {
-  const [zoomDomain, setZoomDomain] = useState({ start: 'dataMin', end: 'dataMax' });
+  const [, setZoomDomain] = useState({ start: 'dataMin', end: 'dataMax' });
   const handleBrushChange = (range) => { if (range) { setZoomDomain({ start: range.startIndex, end: range.endIndex }); } };
   
   const classes = useReportStyles();
@@ -168,6 +168,7 @@ const ChartReportPage = () => {
                 dataKey={timeType}
                 height={30}
                 stroke={theme.palette.primary.main}
+                tickFormatter={() => ''}
                 onChange={(range) => handleBrushChange(range)}
               />
               <Line type="monotone" dataKey={type} stroke={theme.palette.primary.main} dot={false} activeDot={{ r: 6 }} />
--- a/src/reports/ChartReportPage.jsx
+++ b/src/reports/ChartReportPage.jsx
@@ -20,9 +20,6 @@ import {
 import useReportStyles from './common/useReportStyles';
 
 const ChartReportPage = () => {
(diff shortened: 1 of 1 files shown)


tananaev:
Can you provide a screen capture of what it looks like now.

hypoclone:
Yes @tananaev 

https://github.com/user-attachments/assets/cbf9af3e-6397-4d52-871e-7a36a694b23a

tananaev:
It doesn't look like the time is formatter in the `Brush`.

hypoclone:
Thank you, i have two suggestions, i could remove the time or leave it with a good format. see the two outcomes below

https://github.com/user-attachments/assets/8c6c0467-848a-46fc-afc8-743b9437ec9e


https://github.com/user-attachments/assets/c31189d6-43a0-4698-8379-5cc8b93668d6

Which do you suggest?

tananaev:
I guess the second option looks better for now. At least the text is not cut off.

src/reports/ChartReportPage.jsx
```diff
@@ -20,6 +20,9 @@ import {
 import useReportStyles from './common/useReportStyles';
 
 const ChartReportPage = () => {
+  const [, setZoomDomain] = useState({ start: 'dataMin', end: 'dataMax' });
+  const handleBrushChange = (range) => { if (range) { setZoomDomain({ start: range.startIndex, end: range.endIndex }); } };
```
tananaev, line 24: What is the point of this?

src/reports/ChartReportPage.jsx
```diff
@@ -161,7 +164,14 @@ const ChartReportPage = () => {
                 formatter={(value, key) => [value, positionAttributes[key]?.name || key]}
                 labelFormatter={(value) => formatTime(value, 'seconds')}
               />
-              <Line type="monotone" dataKey={type} stroke={theme.palette.primary.main} />
+              <Brush
+                dataKey={timeType}
+                height={30}
+                stroke={theme.palette.primary.main}
+                tickFormatter={() => ''}
```
tananaev, line 171: Is there a way to hide it completely instead of just rendering empty value?

src/reports/ChartReportPage.jsx
```diff
@@ -161,7 +164,14 @@ const ChartReportPage = () => {
...
+                onChange={(range) => handleBrushChange(range)}
+              />
+              <Line type="monotone" dataKey={type} stroke={theme.palette.primary.main} dot={false} activeDot={{ r: 6 }} />
```
tananaev, line 174: Can we move attributes on separate lines for better readability please.

src/reports/ChartReportPage.jsx
hypoclone: Hello @tananaev, good day to you, so sorry was a bit busy, that is how i manage the zoom state. 
`const [, setZoomDomain]` holds the zoom range and  
`const handleBrushChange` updates it based on user interaction.
The first value of `const [, setZoomDomain]` zoomDomain gives error on lint that it's declared and not used, i figure the function could work without it.

src/reports/ChartReportPage.jsx
hypoclone: I can't figure a way now, setting `tick={false}` doesn't work, another option ia to set the time and format with `tickFormatter={(value) => formatTime(value, 'time')}` Please do you have a suggestion?

src/reports/ChartReportPage.jsx
hypoclone: This has been done. Thank you

src/reports/ChartReportPage.jsx
tananaev: If you don't use the value anywhere, I don't understand how it works.

src/reports/ChartReportPage.jsx
tananaev: Let's leave it as empty for now.

src/reports/ChartReportPage.jsx
hypoclone: Yes Anton, it is not used anywhere in the code, it is only to hold and set the initial values for the `setZoomDomain` function, these initial values are provided by `dataMin` and `dataMax` , if you don't mind i could put it back but it will fail linting.

tananaev: Not sure I understand the explanation. Why set some values if they are not used anywhere?

hypoclone: The `useState` is used to keep track of the zoom level of the chart and change when user interact with the `brush.`
Initially, `zoomDomain` starts with `{ start: 'dataMin', end: 'dataMax' }`. This means the whole data set is visible when the chart is first loaded.
`handleBrushChange` is triggered when users interact with `brush`, `setZoomDomain` updates the `zoomDomain` state with the new start and end points, effectively zooming into the selected range.
`handleBrushChange(range)` returns the value stored in `zoomDomain` so it is actually it's values that are in use.
i'm sorry i really don't know how to explain better for you.

https://legacy.reactjs.org/docs/hooks-state.html
the `zoomDomain` could be any name i decided to use nothing because it was failing linting

hypoclone: Essentially updating the `zoomDomain` state in the `handleBrushChange` function, but nothing in the component is actually using the value of `zoomDomain`

tananaev: Let me ask in a different way. What if we completely remove these two lines and `onChange` parameter. What will change in the behavior?

hypoclone: Anton i removed the two const they don't seem to be necessary.

PR #1317, opened 2024-12-27
Better User experience when getting data from API

geelongmicrosoldering:
Hey there,

I noticed some of the code for this was already written, but was not yet incorporated into the Replay Page.

This prevents multiple API requests while the browser is waiting for a response from the server, and provides the user with some feedback about what's happening

### DIFF
--- a/src/other/ReplayPage.jsx
+++ b/src/other/ReplayPage.jsx
@@ -137,20 +137,22 @@ const ReplayPage = () => {
     setFrom(from);
     setTo(to);
     const query = new URLSearchParams({ deviceId, from, to });
-    const response = await fetch(`/api/positions?${query.toString()}`);
-    if (response.ok) {
-      setIndex(0);
-      const positions = await response.json();
-      setPositions(positions);
-      setLoading(false);
-      if (positions.length) {
-        setExpanded(false);
+    try {
+      const response = await fetch(`/api/positions?${query.toString()}`);
+      if (response.ok) {
+        setIndex(0);
+        const positions = await response.json();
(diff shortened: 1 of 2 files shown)


src/other/ReplayPage.jsx
```diff
@@ -140,12 +142,14 @@ const ReplayPage = () => {
       setIndex(0);
       const positions = await response.json();
       setPositions(positions);
+      setLoading(false);
```
tananaev, line 145: Let's call it once in a try-finally.

src/reports/components/ReportFilter.jsx
```diff
@@ -186,7 +186,7 @@ const ReportFilter = ({
             disabled={disabled}
             onClick={() => handleClick('json')}
           >
-            <Typography variant="button" noWrap>{t('reportShow')}</Typography>
+            <Typography variant="button" noWrap>{loading ? t('sharedLoading') : t('reportShow')}</Typography>
```
tananaev, line 189:
```suggestion
            <Typography variant="button" noWrap>{t(loading ? 'sharedLoading' : 'reportShow')}</Typography>
```

src/other/ReplayPage.jsx
tananaev: Check how we do it in other places.

geelongmicrosoldering: sorry i had a moment. was trying to work out what to do with the error in catch, and then realised i could just do try>finally instead

tananaev:
Thank you.

PR #1320, opened 2025-01-06
Fix for TypeError crash on null values for some attributes

sakisdog:
Added null checks in coolantTemp, power and battery attributes.
Removed unused parameter from formatPercentage


tananaev:
Merged, thank you.

PR #1337, opened 2025-03-11
Improve chart visualization for data gaps

codestareeng:
- Show solid line for continuous data segments
- Add dashed line to connect gaps in data
- Fix data scaling when gaps are present by filtering null values

This change improves readability of charts when tracking data has interruptions or missing values.

### DIFF
--- a/src/reports/ChartReportPage.jsx
+++ b/src/reports/ChartReportPage.jsx
@@ -173,14 +173,6 @@ const ChartReportPage = () => {
                 stroke={theme.palette.primary.main}
                 dot={false}
                 activeDot={{ r: 6 }}
-                connectNulls={false}
-              />
-              <Line
-                type="monotone"
-                dataKey={type}
-                stroke={theme.palette.primary.main}
-                strokeDasharray="3 3"
-                dot={false}
                 connectNulls={true}
               />
             </LineChart>
--- a/src/reports/ChartReportPage.jsx
+++ b/src/reports/ChartReportPage.jsx
@@ -176,7 +176,7 @@ const ChartReportPage = () => {
                 connectNulls={false}
               />
               <Line
-                type="linear"
+                type="monotone"
                 dataKey={type}
                 stroke={theme.palette.primary.main}
                 strokeDasharray="3 3"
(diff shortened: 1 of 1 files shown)


tananaev:
Please provide screenshots before and after.

codestareeng:
Before:
before
After:   
```
const values = items.map((it) => it[type]).filter((value) => value != null);
  const minValue = values.length ? Math.min(...values) : 0;
  const maxValue = values.length ? Math.max(...values) : 100;
  const valueRange = maxValue - minValue;
```
after1
After 2 lines:
```
              <Line
                type="monotone"
                dataKey={type}
                stroke={theme.palette.primary.main}
                strokeDasharray="3 3"
                dot={false}
                connectNulls={true}
              />
```
after2

tananaev:
What's the data example these screenshots are for? I don't really understand why points are not connected.

codestareeng:
Any data from the CAN bus. When the ignition is turned off, the CAN bus sleeps.

codestareeng:
Any data from the CAN bus. When the ignition is turned off, the CAN bus sleeps. The screenshots show fuel level data.

codestareeng:
Missclick, sorry

tananaev:
I think you misunderstood the question. I want to see the data that was used for that report.

codestareeng:
[data.json](https://github.com/user-attachments/files/19260200/data.json)

codestareeng:
odbFuelLevel

src/reports/ChartReportPage.jsx
```diff
@@ -173,6 +173,15 @@ const ChartReportPage = () => {
...
                 activeDot={{ r: 6 }}
+                connectNulls={false}
+              />
+              <Line
+                type="monotone"
+                dataKey={type}
+                stroke={theme.palette.primary.main}
+                strokeDasharray="3 3"
+                dot={false}
+                connectNulls={true}
```
tananaev, line 184: Doesn't this mean we draw the whole thing twice?

codestareeng: Yes, here we are drawing two lines. I tried to create additional types to draw the second line separately, but apparently, I need to add them to each missing object; otherwise, the line will not be drawn. I still lack experience in this.

codestareeng: Yeah, I think this is the best decision for now.

tananaev:
Looks like lint checks are failing.

Issue #1338, opened 2025-03-17
MapView recreating controls, losing state and jumping map on style change


paul-777:
**Describe the bug**
MapView.jsx removes and re-adds all it's children when the map style changes and mapReady flips. This reinitializes all the children causing their state to be lost - for MapDefaultCamera for example this causes it to jumpto the default map location again when you change map style - exactly where it jumps depends on whether there is a selected position. Children just need to be aware of the chane and that they need to re-add and map layers and sources.

It is triggered by having this line in MapView.jsx ln123
`{mapReady && children}`

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to the main map page
2. Select a tracker
3. Pan the map to a different location
4. Change the map style
5. Watch the map jump to the default location again rather than staying where it was

A potential fix is here
https://github.com/paul-777/traccar-web/tree/MapviewSubControlsLifecycle

paul-777:
Jumping is the most significant/obvious from a user perspective and one I find very irritating. But I actually found the issue when prototyping something similar to the POI layer and found that every time the map style changed the component re-fetched it's data. Looking at the code I think the PoiMap would do the same. IMO having the children get recreated when changing map style is rather obscure unexpected behaviour that can lead to some difficult to diagnose issues when developing components.
(Thanks for such a great product!)

tananaev:
I don't think it's obscure. It is a very clean way to handle map re-initialization. Otherwise you would have to think about in every single sub-component. Is it possible to have some kind of opt-in option for children?

tananaev:
Looks pretty good I think.

paul-777:
thanks, I'll squash the commits on the branch and create a PR

paul-777:
https://github.com/traccar/traccar-web/pull/1339

PR #1339, opened 2025-03-18
Allow subcontrols of MapView to tag themselves as handling map style …

paul-777:
…changes via a handlesMapReady property and for those components don't reinitialize them on style changes.

Discussed https://github.com/traccar/traccar-web/issues/1338#issuecomment-2733200066

### DIFF
--- a/src/map/core/MapView.jsx
+++ b/src/map/core/MapView.jsx
@@ -50,7 +50,6 @@ const initMap = async () => {
       });
     });
   }
-  updateReadyValue(true);
 };
 
 map.addControl(new maplibregl.NavigationControl());
@@ -65,6 +64,7 @@ const switcher = new SwitcherControl(
           setTimeout(waiting, 33);
         } else {
           initMap();
+          updateReadyValue(true);
         }
       };
       waiting();
@@ -120,7 +120,12 @@ const MapView = ({ children }) => {
 
   return (
     <div style={{ width: '100%', height: '100%' }} ref={containerEl}>
-      {mapReady && children}
+      {React.Children.map(children, (child) => {
+        if (React.isValidElement(child) && child.type.handlesMapReady) {
+          return React.cloneElement(child, { mapReady });
+        }
+        return mapReady ? child : null;
+      })}
     </div>
   );
 };
--- a/src/map/main/MapSelectedDevice.js
+++ b/src/map/main/MapSelectedDevice.js
@@ -1,12 +1,11 @@
 import { useEffect } from 'react';
-
 import { useSelector } from 'react-redux';
 import dimensions from '../../common/theme/dimensions';
 import { map } from '../core/MapView';
 import { usePrevious } from '../../reactHelper';
 import { useAttributePreference } from '../../common/util/preferences';
 
-const MapSelectedDevice = () => {
+const MapSelectedDevice = ({ mapReady }) => {
   const currentTime = useSelector((state) => state.devices.selectTime);
   const currentId = useSelector((state) => state.devices.selectedId);
   const previousTime = usePrevious(currentTime);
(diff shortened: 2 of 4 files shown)


src/map/core/MapView.jsx
```diff
@@ -50,7 +50,6 @@ const initMap = async () => {
       });
     });
   }
-  updateReadyValue(true);
```
tananaev, line 53: Why are we changing this?

paul-777: Can move it back if preferred - was something that took me a moment to work out and seems more logical to have the before and after events reflecting each other

tananaev: If it's only called in one place, it's probably fine to keep.

src/map/core/MapView.jsx
```diff
@@ -118,9 +118,24 @@ const MapView = ({ children }) => {
     };
   }, [containerEl]);
 
+  
+  const modifiedChildren = React.Children.toArray(children)
```
tananaev, line 122: Couldn't we combine these two loops?

src/map/core/MapView.jsx
```diff
@@ -120,7 +120,12 @@ const MapView = ({ children }) => {
 
   return (
     <div style={{ width: '100%', height: '100%' }} ref={containerEl}>
-      {mapReady && children}
+      {React.Children.toArray(children).map(child => {
```
tananaev, line 123:
```suggestion
      {React.Children.map(children, (child, _) => (
```

src/map/main/MapSelectedDevice.js
```diff
@@ -17,17 +16,28 @@ const MapSelectedDevice = () => {
...
       map.easeTo({
         center: [position.longitude, position.latitude],
         zoom: Math.max(map.getZoom(), selectZoom),
         offset: [0, -dimensions.popupMapOffset / 2],
       });
     }
-  });
+
+    prevPosition.current = position;
+  }, [currentId, previousId, currentTime, previousTime, mapFollow, position, selectZoom, mapReady]);
```
tananaev, line 36: What is going on in this file?

src/map/main/PoiMap.js
```diff
@@ -79,9 +80,11 @@ const PoiMap = () => {
...
-  }, [data]);
+  }, [data, mapReady]);
 
   return null;
 };
 
-export default PoiMap;
+PoiMap.handlesMapReady = true;
+
+export default PoiMap;
```
tananaev, line 90: You'll need to fix formatting. Line break is missing here.

src/map/main/MapSelectedDevice.js
```diff
@@ -17,17 +16,28 @@ const MapSelectedDevice = () => {
 
   const position = useSelector((state) => state.session.positions[currentId]);
 
+  const prevPosition = useRef(null);
```
tananaev, line 19: We should probably use `usePrevious` for this as well.

tananaev: It should also be called `previousPosition`.

src/map/core/MapView.jsx
```diff
@@ -118,9 +118,15 @@ const MapView = ({ children }) => {
     };
   }, [containerEl]);
 
+  
   return (
     <div style={{ width: '100%', height: '100%' }} ref={containerEl}>
-      {mapReady && children}
+      {React.Children.map(children, child => {
+        if (React.isValidElement(child) && child.type.handlesMapReady) {
+          return React.cloneElement(child, { mapReady });
```
tananaev, line 126: Would this cause element to re-render all the time?

paul-777: The only change I can see to the rendering in the profiler is the desired behaviour. When mapReady goes to false the "handlesMapReady" children render themselves rather than disappearing from the tree, then when mapReady goes to true all the components re-render - some for the first time and some due to mapReady change.

tananaev:
Looks like you need to fix some link checks.

PR #1340, opened 2025-03-18
Configurable width and opacity for routes and geofences

paul-777:


src/map/main/MapLiveRoutes.js
```diff
@@ -16,6 +16,15 @@ const MapLiveRoutes = () => {
 
   const history = useSelector((state) => state.session.history);
 
+  const reportWidth = useSelector((state) => {
+    return state.session.user.attributes['web.routeWidth'] || state.session.server.attributes['web.routeWidth'] || 2;
+    
+  });
+
+  const reportOpacity = useSelector((state) => {
```
tananaev, line 24: Why are you not using `usePreference` for all of these?

src/resources/l10n/en.json
```diff
@@ -134,7 +134,11 @@
     "attributeDeviceInactivityPeriod": "Device Inactivity Period",
     "attributeProcessingCopyAttributes": "Processing: Copy Attributes",
     "attributeColor": "Color",
+    "attributeWidth": "Width",
+    "attributeOpacity": "Opacity (0.0 - 1.0)",
     "attributeWebLiveRouteLength": "Web: Live Route Length",
+    "attributeWebRouteWidth": "Web: Route Width",
+    "attributeWebRouteOpacity": "Web: Route Opacity (0.0 - 1.0)",
```
tananaev, line 141: 1. Why do we need two sets of attributes?
2. Normally we don't include value range in the string like this.

paul-777: 1. A misunderstanding on my part on the use of resource strings - on is a geofence attribute on is the server/user attribute for routes but I see now they can both use the same resource string.
2. Opacity to me doesn't have a clear scale as it gets expressed in different ways ( 0.0-1.0, 0-100, 0-255 ) so it seemed helpful to express the range to avoid guessing/confusion as users aren't developers

src/map/main/MapLiveRoutes.js
paul-777: I'll change, hadn't found it. Assume `useAttributePreference` ?

tananaev: Yes, I think one of those.

src/resources/l10n/en.json
paul-777: Actually I think having these as separate strings is necessary as they are for separate attributes - one attached to geofences, the other to user/server for routes. I'm not sure there is a common string that would make sense in both cases.

tananaev: Why not? Something like "Map Line Opacity".

src/common/attributes/useCommonUserAttributes.js
```diff
@@ -125,6 +125,14 @@ export default (t) => useMemo(() => ({
     name: t('attributeWebLiveRouteLength'),
     type: 'number',
   },
+  'web.routeWidth': {
```
tananaev, line 128: I think we should unify the attribute key as well.

src/common/attributes/useCommonUserAttributes.js
```diff
@@ -125,6 +125,14 @@ export default (t) => useMemo(() => ({
     name: t('attributeWebLiveRouteLength'),
     type: 'number',
   },
+  'web.lineWidth': {
```
tananaev, line 128: I think it's better to call it `mapLineWidth`. Adding `web.` is an old way. Usually now we just have a camel-case name without dots.

src/map/main/MapLiveRoutes.js
```diff
@@ -16,6 +16,9 @@ const MapLiveRoutes = () => {
 
   const history = useSelector((state) => state.session.history);
 
+  const reportWidth = useAttributePreference('mapLineWidth', 2);
+  const reportOpacity = useAttributePreference('mapLineOpacity', 1);
```
tananaev, line 20: You should rename these variable as well to match. This one is definitely incorrect because it's not a report, but for consistency just call all variables `mapLine...`.

src/map/core/mapUtil.js
```diff
@@ -80,6 +80,8 @@ export const geofenceToFeature = (theme, item) => {
     properties: {
       name: item.name,
       color: item.attributes.color || theme.palette.geometry.main,
+      width: item.attributes.mapLineWidth || 2,
+      opacity: item.attributes.mapLineOpacity || 100,
```
tananaev, line 84: Sorry, just noticed another thing. Why is the opacity 100 here instead of 0-1?

PR #1342, opened 2025-03-20
Add color picker and slider control for attribute editing

paul-777:
Allow slider control for range attributes and color picker for color attrinbtes
image


tananaev:
I think this is too much extra code for features that are rarely used.

paul-777:
Ok, for my use case I'll be adding blocks of devices and geofences frequently and looking to color them uniquely within their blocks so this sort of think will really improve usability. I'll just maintain a fork.

PR #1343, opened 2025-03-24
Add missing engineTemp attribute

jcardus:
So we can show it on StatusCard

### DIFF
--- a/src/common/attributes/usePositionAttributes.js
+++ b/src/common/attributes/usePositionAttributes.js
@@ -108,6 +108,10 @@ export default (t) => useMemo(() => ({
     name: t('positionCoolantTemp'),
     type: 'number',
   },
+  engineTemp: {
+    name: t('positionEngineTemp'),
+    type: 'number',
+  },
   gps: {
     name: t('positionGps'),
     type: 'number',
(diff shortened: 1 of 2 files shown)


src/common/attributes/usePositionAttributes.js
```diff
@@ -108,6 +108,10 @@ export default (t) => useMemo(() => ({
     name: t('positionCoolantTemp'),
     type: 'number',
   },
+  engineTemp: {
+    name: t('positionEngineTemp'),
```
tananaev, line 112: I don't see this value in the strings. I also don't see it on the backend as a known attribute.

jcardus: https://github.com/traccar/traccar/blob/5e6c6119535f057e73d6b4fe3ba0e2617a7c537d/src/main/java/org/traccar/protocol/TeltonikaProtocolDecoder.java#L245

tananaev: We only add standard attributes.

What about the first part?

tananaev: It seems like my comment is not addressed.

jcardus: Should we change server side? What do you suggest for being able to see the engine temperature in the device status?

tananaev: If you want to add a new known attribute you should add it on the server side. The minimum requirement is that it is used by at least two different protocols. Then we can add the web app part.

PR #1347, opened 2025-04-01
Add position id as an attribute to route reports

Earlopain:
It can be useful in combination with `DELETE /api/positions/:id` from https://github.com/traccar/traccar/pull/5537

This may also make sense on other table elements (there are a few) but I personally have no usecase for doing that.

### DIFF
--- a/src/reports/RouteReportPage.jsx
+++ b/src/reports/RouteReportPage.jsx
@@ -140,7 +140,7 @@ const RouteReportPage = () => {
             </TableHead>
             <TableBody>
               {!loading ? items.slice(0, 4000).map((item) => (
-                <TableRow key={item.id}>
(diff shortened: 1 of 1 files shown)


Earlopain:
It just adds the position id as an html attribute to the table element on `/reports/route`

image

I want to write a simple userscript that offers me to delete wrong position information. I can use the UI to check where it is on the map, what speed it has etc. in relation to other position.

It's very hard to get data (like the position id) out of react. With this, I could simply read that attribute. I understand it's very specific and probably not broadly useful, so do with this as you wish.

tananaev:
I think it's better to add an option to delete right in the web app instead of this.

Earlopain:
Oh, sure, that would be even better. I can try and make a PR for that instead. Let me close this one for now

PR #1348, opened 2025-04-01
Add position deletion button to route report

Earlopain:
Use this to delete wrong data.

Followup to https://github.com/traccar/traccar-web/pull/1347. It uses the newly added endpoint to delete a position added in https://github.com/traccar/traccar/pull/5537

I placed the delete icon at the very end of the table. I feel like putting it next to the gps icon invites users to accidentally click on it (even if it has a confirmation prompt).

Again, very much out of my comfort zone with this (React). I largely followed how other parts of the codebase do it, like `ScheduledPage.jsx`.

[Screencast_20250401_173110.webm](https://github.com/user-attachments/assets/d7c47f8c-818a-4130-a1d7-3f6ea0497889)

### DIFF
--- a/src/reports/RouteReportPage.jsx
+++ b/src/reports/RouteReportPage.jsx
@@ -24,6 +24,8 @@ import MapCamera from '../map/MapCamera';
 import MapGeofence from '../map/MapGeofence';
 import scheduleReport from './common/scheduleReport';
 import MapScale from '../map/MapScale';
+import { useRestriction } from '../common/util/permissions';
+import CollectionActions from '../settings/components/CollectionActions';
 
 const RouteReportPage = () => {
   const navigate = useNavigate();
@@ -33,6 +35,7 @@ const RouteReportPage = () => {
   const positionAttributes = usePositionAttributes(t);
 
   const devices = useSelector((state) => state.devices.items);
+  const readonly = useRestriction('readonly');
 
   const [available, setAvailable] = useState([]);
   const [columns, setColumns] = useState(['fixTime', 'latitude', 'longitude', 'speed', 'address']);
@@ -136,6 +139,7 @@ const RouteReportPage = () => {
                 <TableCell className={classes.columnAction} />
                 <TableCell>{t('sharedDevice')}</TableCell>
                 {columns.map((key) => (<TableCell key={key}>{positionAttributes[key]?.name || key}</TableCell>))}
+                <TableCell className={classes.columnAction} />
               </TableRow>
             </TableHead>
             <TableBody>
@@ -162,6 +166,17 @@ const RouteReportPage = () => {
                       />
                     </TableCell>
                   ))}
+                  <TableCell className={classes.actionCellPadding}>
+                    <CollectionActions
+                      itemId={item.id}
+                      endpoint="positions"
+                      readonly={readonly}
+                      setTimestamp={() => {
+                        // NOTE: Gets called when an item was removed
+                        setItems(items.filter((position) => position.id !== item.id));
+                      }}
+                    />
+                  </TableCell>
                 </TableRow>
               )) : (<TableShimmer columns={columns.length + 2} startAction />)}
             </TableBody>
--- a/src/reports/common/useReportStyles.js
+++ b/src/reports/common/useReportStyles.js
@@ -46,4 +46,10 @@ export default makeStyles((theme) => ({
     flexGrow: 1,
     overflow: 'hidden',
   },
+  actionCellPadding: {
+    '&.MuiTableCell-body': {
+      paddingTop: 0,
+      paddingBottom: 0,
+    },
+  },
 }));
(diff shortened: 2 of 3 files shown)


src/reports/RouteReportPage.jsx
```diff
@@ -162,10 +170,26 @@ const RouteReportPage = () => {
                       />
                     </TableCell>
                   ))}
+                  {!readonly && (
+                    <TableCell style={{ paddingTop: '0px', paddingBottom: '0px' }}>
```
Earlopain, line 174: Buttons have a pretty large top/bottom maring that doesn't look very good. The first icon does ` padding="none"`, but that also removes left/right margin which I would want to keep for this one in order for it to not hug the right screen edge.

Let me know if there is a better method of doing this.

tananaev: Can we reuse `CollectionActions`?

Earlopain: No, I don't think so. It assumes that the model is editable, which isn't true for positions

tananaev: We can make edit optional?

Earlopain: Yeah, I guess I could try that. I will give it a go later today.

I haven't tried it out yet, it seems to move actions (just one in this case) into a dropdown menu, at least on mobile. The gps icon on the left definitely stay as it is at the moment, right?

Earlopain: I changed it over, let me know if it looks good to you


https://github.com/user-attachments/assets/4837e6e1-2fbb-4a6a-8f53-46dd72ba89ca

src/settings/components/CollectionActions.jsx
```diff
@@ -45,7 +45,7 @@ const CollectionActions = ({
     setMenuAnchorEl(null);
   };
 
-  const handleRemoveResult = (removed) => {
+  const defaultHandleRemoveResult = (removed) => {
     setRemoving(false);
```
tananaev, line 49: Your new method doesn't call `setRemoving`.

src/reports/RouteReportPage.jsx
```diff
@@ -162,6 +166,16 @@ const RouteReportPage = () => {
                       />
                     </TableCell>
                   ))}
+                  <TableCell style={{ paddingTop: '0px', paddingBottom: '0px' }}>
+                    <CollectionActions
+                      itemId={item.id}
+                      endpoint="positions"
+                      readonly={readonly}
+                      handleRemoveResult={() => {
```
tananaev, line 174: I think instead of adding `handleRemoveResult`, we can just reuse `setTimestamp`. It will only be called on deletion. A bit hacky and indirect, but doesn't require a completely new parameter.

src/settings/components/CollectionActions.jsx
```diff
@@ -65,7 +65,9 @@ const CollectionActions = ({
             ))}
             {!readonly && (
               <>
-                <MenuItem onClick={handleEdit}>{t('sharedEdit')}</MenuItem>
+                {editPath && (
+                  <MenuItem onClick={handleEdit}>{t('sharedEdit')}</MenuItem>
+                )}
```
tananaev, line 70:
```suggestion
                {editPath && <MenuItem onClick={handleEdit}>{t('sharedEdit')}</MenuItem>}
```

src/settings/components/CollectionActions.jsx
```diff
@@ -82,11 +84,13 @@ const CollectionActions = ({
           ))}
           {!readonly && (
             <>
-              <Tooltip title={t('sharedEdit')}>
-                <IconButton size="small" onClick={handleEdit}>
-                  <EditIcon fontSize="small" />
-                </IconButton>
-              </Tooltip>
+              {editPath && (
```
tananaev, line 87: I like that this reuses existing parameter.

src/settings/components/CollectionActions.jsx
Earlopain: Oh, you're right. Seems to work out alright though, I think it's supposed to hide the popup? Anyways, I'll apply your suggestion from below to fix this.

src/reports/RouteReportPage.jsx
```diff
@@ -162,6 +166,16 @@ const RouteReportPage = () => {
                       />
                     </TableCell>
                   ))}
+                  <TableCell style={{ paddingTop: '0px', paddingBottom: '0px' }}>
```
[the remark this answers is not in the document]
Earlopain: The default padding has more precedence, can I handle this without adding `!important`?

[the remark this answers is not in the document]
tananaev: not sure

src/reports/common/useReportStyles.js
```diff
@@ -46,4 +46,8 @@ export default makeStyles((theme) => ({
     flexGrow: 1,
     overflow: 'hidden',
   },
+  paddingIcon: {
```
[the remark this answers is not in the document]
Earlopain: Yes, that's better

tananaev:
Merged. Thanks for the contribution and for fixing all my comments.

Earlopain:
Great, thanks to you as well (:

PR #1349, opened 2025-04-02
Add an icon on the trip report to view the replay

Earlopain:
It uses query parameters to carry the state, so when you navigate back you still have your previous
timerange selected.

There's logic that when a query parameter `deviceId` is present, it navigates back to the root page. I've changed that to only apply on the root page itself, because I'd like to use a parameter named like that on a different page.

I also tried integrating the replay-style map directly into the trips page but that is above what I'm capable of.

[Screencast_20250402_211313.webm](https://github.com/user-attachments/assets/d5c716c6-f8da-4267-94ec-a041bc3e9d16)

### DIFF
--- a/src/other/ReplayPage.jsx
+++ b/src/other/ReplayPage.jsx
@@ -12,7 +12,7 @@ import PlayArrowIcon from '@mui/icons-material/PlayArrow';
 import PauseIcon from '@mui/icons-material/Pause';
 import FastForwardIcon from '@mui/icons-material/FastForward';
 import FastRewindIcon from '@mui/icons-material/FastRewind';
-import { useNavigate } from 'react-router-dom';
+import { useLocation, useNavigate } from 'react-router-dom';
 import { useSelector } from 'react-redux';
 import MapView from '../map/core/MapView';
 import MapRoutePath from '../map/MapRoutePath';
@@ -82,6 +82,8 @@ const ReplayPage = () => {
   const navigate = useNavigate();
   const timerRef = useRef();
 
+  const locationState = useLocation().state;
+
   const defaultDeviceId = useSelector((state) => state.devices.selectedId);
 
   const [positions, setPositions] = useState([]);
@@ -93,6 +95,7 @@ const ReplayPage = () => {
   const [expanded, setExpanded] = useState(true);
   const [playing, setPlaying] = useState(false);
   const [loading, setLoading] = useState(false);
+  const [initialState, setInitialState] = useState(locationState);
 
   const deviceName = useSelector((state) => {
     if (selectedDeviceId) {
@@ -133,6 +136,7 @@ const ReplayPage = () => {
 
   const handleSubmit = useCatch(async ({ deviceId, from, to }) => {
     setLoading(true);
+    setInitialState(null);
     setSelectedDeviceId(deviceId);
     setFrom(from);
     setTo(to);
@@ -219,7 +223,7 @@ const ReplayPage = () => {
               </div>
             </>
           ) : (
-            <ReportFilter handleSubmit={handleSubmit} fullScreen showOnly loading={loading} />
+            <ReportFilter handleSubmit={handleSubmit} fullScreen showOnly loading={loading} initialState={initialState} />
           )}
         </Paper>
       </div>
--- a/src/reports/TripReportPage.jsx
+++ b/src/reports/TripReportPage.jsx
@@ -5,10 +5,11 @@ import {
 } from '@mui/material';
 import GpsFixedIcon from '@mui/icons-material/GpsFixed';
 import LocationSearchingIcon from '@mui/icons-material/LocationSearching';
+import RouteIcon from '@mui/icons-material/Route';
 import {
   formatDistance, formatSpeed, formatVolume, formatTime, formatNumericHours,
 } from '../common/util/formatter';
-import ReportFilter from './components/ReportFilter';
+import ReportFilter, { replayReportLocationState } from './components/ReportFilter';
 import { useAttributePreference } from '../common/util/preferences';
 import { useTranslation } from '../common/components/LocalizationProvider';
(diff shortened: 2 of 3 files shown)


tananaev:
Not sure about this. It would be good to support query parameters for reports, but I don't want it to be a one-off thing.

Earlopain:
How about `location.state` instead? I first had it modify the actual ReportFilter parameters but it was very annoying when navigating back and my previous filter was just overwritten with something else.

I'm not sure how adding the same logic to the other reports would work. With the replay one, it's very easy to just hide the filters and use some other conditions. I guess it could simply load the query parameters? But then I'd _really_ want add something like `ignoreQueryParameters` to the filter on the replay page (because it's so annoying for it to change out under you when clicking a button)

Earlopain:
With `location.state` the benefit is that the user can't think that query parameters will also work for other reports. I really just want a way to move some information across navigate (without modifying the actual filter store data), and I didn't know about `navigate({state})` when I was starting on this.

Earlopain:
I've moved over to using state. Let me know what you think about that

src/reports/common/useReportStyles.js
```diff
@@ -52,4 +52,7 @@ export default makeStyles((theme) => ({
       paddingBottom: 0,
     },
   },
+  actionCellIcons: {
```
tananaev, line 55: What does this do?

src/reports/TripReportPage.jsx
```diff
@@ -189,16 +200,21 @@ const TripReportPage = () => {
...
-                  <TableCell className={classes.columnAction} padding="none">
-                    {selectedItem === item ? (
-                      <IconButton size="small" onClick={() => setSelectedItem(null)}>
-                        <GpsFixedIcon fontSize="small" />
-                      </IconButton>
-                    ) : (
-                      <IconButton size="small" onClick={() => setSelectedItem(item)}>
-                        <LocationSearchingIcon fontSize="small" />
+                  <TableCell className={classes.columnAction}>
+                    <div className={classes.actionCellIcons}>
```
tananaev, line 204: Why do you need a div?

src/other/ReplayPage.jsx
```diff
@@ -161,6 +163,14 @@ const ReplayPage = () => {
     window.location.assign(`/api/positions/kml?${query.toString()}`);
   };
 
+  useEffect(() => {
+    const { from, to, deviceId } = state;
+    if (from && to && deviceId) {
+      setShowFilter(false);
+      handleSubmit({ from, to, deviceId });
```
tananaev, line 170: I don't think this is right. It basically ends up in inconsistent state.

src/reports/common/useReportStyles.js
Earlopain: It was supposed to display the two icons inline instead of below each other. But the better solution is to just add a new table row for this so this becomes unnnessesary. It's gone now

src/reports/TripReportPage.jsx
Earlopain: I wanted to add the class I previously added and it looks like with react `className` you can only add a single class nicely, with more you have to use interpolation or something like that?

Anyways, it's gone now, since I got rid of the need for extra styling

src/other/ReplayPage.jsx
Earlopain: Hm, does it? `handleSubmit` seems to persist everything it needs for it to make sense https://github.com/traccar/traccar-web/blob/fcf357ce628986d7b137a3e3d6de41f014241ca7/src/other/ReplayPage.jsx#L136-L138 The download button for example behaves correctly.

Or do you mean something else?

tananaev: But not in the filter.

Earlopain: The filter is not rendered in this case. Having it update in the filter would throw away your previously selected timerange on the trips report when you navigate back which feels very bad

tananaev: That is the inconsistency I'm talking about. You're basically just hiding it.

Earlopain: OK, I moved the state handling into the report component itself.

It simply prefers the state filter but reverts back to the store when the user closes the replay controls. So no more hiding the filter, the user can change it if they decide to close the replay.

It also should make it possible to load other reports on page load with relatively few changes in the future, if there is a usecase for that.

src/reports/components/ReportFilter.jsx
```diff
@@ -90,6 +114,14 @@ const ReportFilter = ({
     }
   };
 
+  useEffect(() => {
+    if (locationState.showOnPageLoad) {
+      // Don't treat re-renders as first page load.
+      locationState.showOnPageLoad = false;
```
Earlopain, line 120: Not so sure about this, it feels like there has to be a better way to accomplish the same thing.

Without this, the replay filter just recloses immediatly when the users wants to adjust the filter.

tananaev: Yes, this all doesn't look clean enough.

Earlopain: I moved this into the parent component. I think this is better?

Earlopain: I've moved the entire state into the main page component and pass it down into the filter report now. I don't think having the filter contain the logic for that was the best idea.

I hope this addresses your concern

Earlopain:
Hey, do you want me to change anything here? I think it's encapsulated fine now, with the location state being part of the main page component instead of what I had before.

tananaev:
As I said earlier, my concern is multiple ways of doing the same thing here. I don't like that we have a state and parameters. We should do one way or another. I like the parameters option, but then we need to migrate everything to it.

Earlopain:
I realize I don't know enough about react to put this into an acceptable state. Oh well

PR #1357, opened 2025-04-05
Adds map overlay switch on map

TheYakuzo:
Introduces a button to toggle the overlay's visibility. The overlay preference persists across sessions.

Memoize MapOverlay to avoid unnecessary re-renders

Move activeOverlay management from MapOverlay.js to MainMap.jsx to avoid code duplication on MapOverlay.js and MapOverlayButton.js [ new component ]

The button is not rendered when no map overlay is indicated (hence the displacement of the activeOverlay logic).

Screenshot 2025-04-05 at 18 34 07

Screenshot 2025-04-05 at 18 34 17


src/map/overlay/MapOverlayButton.js
```diff
@@ -0,0 +1,53 @@
+import { useEffect, useMemo } from 'react';
+import { map } from '../core/MapView';
+import './overlay.css';
+
+const statusClass = (status) => `maplibregl-ctrl-icon maplibre-ctrl-overlay maplibre-ctrl-overlay-${status}`;
+
+class OverlayControl {
```
tananaev, line 7: Is this basically a copy of the alarm button? I think we need to figure out how to reuse a button instead of copying.

TheYakuzo: Yes, it's true, I'll see tomorrow to redo these two buttons.
Do you usually put them in src/common/components?

tananaev: Map components should be under map folder.

tananaev:
Can we split into 2 PRs:
1. Refactor button to be generic
2. Add overlay control button

Also I think the icon shouldn't have a color.

TheYakuzo:
Done : https://github.com/traccar/traccar-web/pull/1358 ( refactoring )
This current PR remains the same, there should be no conflicts.

TheYakuzo:
color is now black for enabled and gray for disabled

tananaev:
We already implemented something similar.

PR #1359, opened 2025-04-07
Decouple display locale from date locale

Earlopain:
I have my browser display content in english by default, so the locale for dates also becomes en. However, regardless of that setting my default locale for dates is de, which among other things causes weeks to start on monday (not sunday).

It aligns it to which locale the browser renders date input fields (am/pm 12h vs 24h, DDMMYYYY vs MMDDYYYY, etc.). These are already in the correct locale (display locale != date locale)

Browsers don't offer to configure display and date locales separately. Rather such settings are delegated to the operating system. On Windows that would be "Time and Language / Language and Region / Regional Format).

I believe this affects the reports only (primarily "This Week"/"Previous Week", other places explicitly overwrite the locale to en.

`formatTime` already uses the correct locale because it works through `toLocaleDateString` under the hood. It's a build-in browser method that chooses the appropriate locale automatically.

### DIFF
--- a/src/common/components/LocalizationProvider.jsx
+++ b/src/common/components/LocalizationProvider.jsx
@@ -144,6 +144,19 @@ const getDefaultLanguage = () => {
   return 'en';
 };
 
+const getDefaultDateLocale = () => {
+  let locale;
+  if (typeof Intl !== 'undefined' && typeof Intl.DateTimeFormat !== 'undefined') {
+    locale = Intl.DateTimeFormat().resolvedOptions().locale;
+  } else {
+    locale = getDefaultLanguage();
+  }
+  if (locale.length > 2) {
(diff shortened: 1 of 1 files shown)


src/common/components/LocalizationProvider.jsx
```diff
@@ -144,6 +144,19 @@ const getDefaultLanguage = () => {
   return 'en';
 };
 
+const getDefaultDateLocale = () => {
```
tananaev, line 147: It doesn't make sense to me that we're completely ignoring selected language.

Earlopain: This sucks actually. I just tested it on chrome and it behaves differently from firefox. Firefox uses the host os language to render input elements but chrome uses the browser language setting. I guess I will just choose the a language with the expected week format

PR #1361, opened 2025-04-11
Add position fuel used

jcardus:


src/common/attributes/usePositionAttributes.js
```diff
@@ -224,6 +224,10 @@ export default (t) => useMemo(() => ({
     name: t('positionFuelConsumption'),
     type: 'number',
   },
+  fuelUsed: {
```
tananaev, line 227: let's move this above `fuelConsumption` to match the order we have in `Position.java`

PR #1363, opened 2025-04-12
Update ChartReportPage.jsx

hypoclone:
Added multiple select to chart

https://github.com/user-attachments/assets/bf6c0e31-b9eb-4faf-8869-d5cd02a96d5e

### DIFF
--- a/src/reports/ChartReportPage.jsx
+++ b/src/reports/ChartReportPage.jsx
@@ -119,7 +119,6 @@ const ChartReportPage = () => {
               value={selectedTypes}
               onChange={(e) => setSelectedTypes(e.target.value)}
               multiple
-              renderValue={(selected) => selected.join(', ')}
               disabled={!items.length}
             >
               {types.map((key) => (
--- a/src/reports/ChartReportPage.jsx
+++ b/src/reports/ChartReportPage.jsx
@@ -98,14 +98,15 @@ const ChartReportPage = () => {
     }
   });
 
-  const chartColor = () => {
-    const letters = '4E5A76B829DF3C10';
-    let color = '#';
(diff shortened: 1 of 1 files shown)


src/reports/ChartReportPage.jsx
```diff
@@ -98,6 +98,15 @@ const ChartReportPage = () => {
...
   });
 
+  const chartColor = () => {
+    const letters = '4E5A76B829DF3C10';
+    let color = '#';
+    for (let i = 0; i < 6; i += 1) {
+      color += letters[Math.floor(Math.random() * 16)];
+    }
+    return color;
+  };
```
tananaev, line 108: I think we should use some predefined palette. Ideally from the theme.

src/reports/ChartReportPage.jsx
```diff
@@ -106,8 +115,10 @@ const ChartReportPage = () => {
             <InputLabel>{t('reportChartType')}</InputLabel>
             <Select
               label={t('reportChartType')}
-              value={type}
-              onChange={(e) => setType(e.target.value)}
+              value={selectedTypes}
+              onChange={(e) => setSelectedTypes(e.target.value)}
+              multiple
+              renderValue={(selected) => selected.join(', ')}
```
tananaev, line 121: Isn't this the default?

hypoclone: the default was for a single selection, this is for multiple selection

tananaev: Not sure I understand. I'm talking about line 121 here.

hypoclone: Oh! you are right, thank you

tananaev:
Merged, thank you.

PR #1364, opened 2025-04-14
Update localization files (es)

miguelcarrascoq:


tananaev:
Localization should be done through Transifex:

https://transifex.com/traccar/

PR #1372, opened 2025-04-28
added login field in user view and users grid

cadore:


tananaev:
You're missing a description on why it would be useful for most people.

tananaev:
I believe we only use it for LDAP integration, which is relatively rare scenario.

cadore:
ok, dear
I'll close the request, maybe put the login field in attributes.

PR #1381, opened 2025-05-27
feat: position convert

sgpublic:
This PR is based on traccar/traccar#5566.

This PR adds support for the following map platforms:

+ Tencent Maps

PS: Baidu Maps support is not included in this PR because its tile cutting algorithm is different from the norm and requires further investigation.

This PR completes the display of different coordinates for different map types. Specifically: when using Amap, display GCJ-02 coordinates; when using OpenFreeMap, display original coordinates.

This PR fixes an i18n translation error:

+ `mapOpenFreeMap` in `zh.json`: should be `开放自由地图`


tananaev:
We have implemented GCJ-02 support in Traccar:

https://github.com/traccar/traccar-web/commit/f71884a29993772f49135d118f29d4f5a9c48564

PR #1403, opened 2025-06-15
Update DeviceRow.jsx to highlight selected device

hypoclone:
This commit add highlight to the selected device in devices list.


src/main/DeviceRow.jsx
```diff
@@ -84,6 +88,8 @@ const DeviceRow = ({ data, index, style }) => {
         key={item.id}
         onClick={() => dispatch(devicesActions.selectId(item.id))}
         disabled={!admin && item.disabled}
+        selected={selectedDeviceId === item.id}
+        className={selectedDeviceId === item.id ? classes.selected : undefined}
```
tananaev, line 92: Can we use `null` instead of `undefined`?

tananaev:
Looks pretty nice. Thanks!

PR #1431, opened 2025-07-06
Use datetime-local input, support clear button, show expiration time with minutes

Sinaloense:
Problem explanation:

- My local time zone is GMT-7.  
- When selecting an expiration date: `07/08/2025 `
- The database stores it as: `2025-07-08 00:00:00`
- However, when displaying expirationTime using `formatTime(item.expirationTime, 'date')`, it appears as `07/07/2025` (one day earlier).  
- This happens because the date is not stored in UTC, and converting to local time shifts the date.

Solution:
~~Store the date always in **UTC** to avoid timezone-related mismatches.~~
Change input date to datetime-local.

- Change the input to datetime-local to support hours and minutes.
  PC:
  image

  Android:
  image

Bonus:
- Add support for the Clear button, which now properly set the field to `01/01/2099 12:00 AM` and value to `null`.  
  image

- Show the expiration date with hours and minutes in the table.  
  image

### DIFF
--- a/src/common/util/formatter.js
+++ b/src/common/util/formatter.js
@@ -2,6 +2,7 @@ import dayjs from 'dayjs';
 import duration from 'dayjs/plugin/duration';
 import relativeTime from 'dayjs/plugin/relativeTime';
 import localizedFormat from 'dayjs/plugin/localizedFormat';
+import utc from 'dayjs/plugin/utc';
 
 import {
   altitudeFromMeters,
@@ -18,6 +19,7 @@ import { prefixString } from './stringUtils';
 dayjs.extend(duration);
 dayjs.extend(relativeTime);
 dayjs.extend(localizedFormat);
+dayjs.extend(utc);
 
 export const formatBoolean = (value, t) => (value ? t('sharedYes') : t('sharedNo'));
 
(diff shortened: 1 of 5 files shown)


tananaev:
I don't think storing in UTC is the right approach. I think what we should do is send full ISO time with the timezone, like we do in other places, like reports.

Sinaloense:
I realized I incorrectly named this as "UTC", when in fact it's being sent as an ISO.
The database stores it like: `2025-07-13 07:00:00`

```
export const formatLocalTimeToUTC = (value) => {
  if (value) {
    return dayjs(value).utc().toISOString();
  }

  return '';
}
```

But I now see this can be simplified by using dayjs().format()
So instead of:
`setItem({ ...item, expirationTime: formatLocalTimeToUTC(e.target.value) });`
I can just do:
`setItem({ ...item, expirationTime: dayjs(e.target.value).format() });`

I’ve made the changes, please review it when you have time.

Sinaloense:
I have now reverted to using `toISOString()` for `expirationTime`:
`setItem({ ...item, expirationTime: new Date(e.target.value).toISOString() });`

The issue was caused by using an input of type `"date"`.
Switching to type `"datetime-local"` resolves the problem.

PR #1471, opened 2025-08-09
Adding dynamic icons and address

neuvill:
This is a contribution that allows the user to:

**1. Choose dynamic (moving, parking and idling) in icons list,
2. Adding the current address as asecond text under the device names.**

**1.  Dynamic icons:**

To be able to use this feature, a computed attribute must be created as follows:

_Description:_ Dynamic status,
_Attribut:_ Status
_Expression:_
`motion ? 'moving' : (ignition ? 'idling' : 'parcked')`
_Type:_ string.

The best is to have the status updated dynamically in position table by adding adding a column in tc_positions that follows the same logic as the computed attribute

**2.  Address:**
The address can be diplayed in real time under the device, by choosing Address as Device Details in preferences, the configuration file must be updated accordingly


neuvill:
OK will do the necessary

PR #1474, opened 2025-08-14
Adding dynamic icons and Address

neuvill:
This is a contribution that allows users to choose "Dynamic icons" as an icon and "Address" as a Device details (secondary text):

### **1- Dynamic icons:**

In the device now you can choose Dynamic icon and this allows you to view the stat of the device (moving, idling and parked) directly in the device list or in the map and it changes dynamically depending on the state, 

to allow this to work,we need to add a _calculated attribute_ as follows:

**Description:** Dynamic status
**Attribut:** dynamicStatus
**Expression:**
`motion ? 'moving' : (ignition ? 'idling' : 'parked')`
**Type:** String

### **2 - Address:**

Now the Address can be choose as device details (secondary text), just make sure that Traccar is connected to an _unlimited reverse geocoding server_ and the conf file is correctly configured to always display the address.

**PS** 
I also added the total number of devices in the search bar.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6490,19 +6490,6 @@
         "url": "https://opencollective.com/eslint"
       }
     },
-    "node_modules/eslint/node_modules/@eslint/js": {
-      "version": "9.32.0",
-      "resolved": "https://registry.npmjs.org/@eslint/js/-/js-9.32.0.tgz",
-      "integrity": "sha512-BBpRFZK3eX6uMLKz8WxFOBIFFcGFJ/g8XuwjTHCqHROSIsopI+ddn/d5Cfh36+7+e5edVS8dbSHnBNhrLEX0zg==",
-      "dev": true,
-      "license": "MIT",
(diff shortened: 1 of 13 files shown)


tananaev:
Wait, I thought dynamic icons mean you can upload your own icon. I'm not sure I understand the value of this change.

neuvill:
The idea is to have the stat of the device as you can see in the screenshot hereunder:
Screenshot 2025-08-14 114136
Screenshot 2025-08-14 114055

Allowing the users to add their own icons is a little bit tricky, it must be converted before storing it

tananaev:
I don't think simply replacing existing icons make sense. Maybe the motion status can be on the right side of the list, next to ignition and other icons.

tananaev:
Rotating icon on the map also doesn't make sense because we already have a direction indicator.

tananaev:
In general this feature needs much more consideration and discussion before implementing.

PR #1483, opened 2025-08-23
Search option for multiple selection field

harshanasrimal:
When managing a large number of devices, selecting the correct one from a long dropdown list is difficult and time-consuming.

### **Improvements**
- Added a search ability to the multiple enabled selection field.
- Made the device dropdown in reports searchable, allowing users to quickly filter and find devices.

### **Benefits**
- Improves usability when handling many devices.
- Reduces time spent scrolling through long dropdowns.
- Provides a smoother and more efficient reporting workflow.

### DIFF
--- a/src/common/components/SelectField.jsx
+++ b/src/common/components/SelectField.jsx
@@ -34,8 +34,7 @@ const SelectField = ({
     if (!multiple || value.length === 0) {
       return t('sharedSearch');
     }
-    const selectedTitles = items.filter(item => value.includes(keyGetter(item))).map(item => titleGetter(item)).join(', ');
-    return selectedTitles;
+    return items.filter(item => value.includes(keyGetter(item))).map(item => titleGetter(item)).join(', ');
   };
 
   useEffect(() => setItems(data), [data]);
--- a/src/common/components/SelectField.jsx
+++ b/src/common/components/SelectField.jsx
@@ -1,9 +1,10 @@
 import { useEffect, useState } from 'react';
 import {
-  FormControl, InputLabel, MenuItem, Select, Autocomplete, TextField,
+  FormControl, MenuItem, Autocomplete, TextField,
 } from '@mui/material';
 import { useEffectAsync } from '../../reactHelper';
 import fetchOrThrow from '../util/fetchOrThrow';
+import { useTranslation } from './LocalizationProvider';
 
 const SelectField = ({
   label,
@@ -17,15 +18,24 @@ const SelectField = ({
   data,
   keyGetter = (item) => item.id,
   titleGetter = (item) => item.name,
-  searchable = false, // New prop to enable search for multiple
 }) => {
   const [items, setItems] = useState();
+
+  const t = useTranslation();
 
   const getOptionLabel = (option) => {
     if (typeof option !== 'object') {
       option = items.find((obj) => keyGetter(obj) === option);
     }
     return option ? titleGetter(option) : emptyTitle;
+  };
+
+  const getPlaceholder = () => {
+    if (!multiple || value.length === 0) {
+      return t('sharedSearch');
+    }
+    const selectedTitles = items.filter(item => value.includes(keyGetter(item))).map(item => titleGetter(item)).join(', ');
+    return selectedTitles;
   };
 
   useEffect(() => setItems(data), [data]);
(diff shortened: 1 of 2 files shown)


tananaev:
Looks better. Is it ready for review?

src/common/components/SelectField.jsx
```diff
@@ -17,6 +17,7 @@ const SelectField = ({
   data,
   keyGetter = (item) => item.id,
   titleGetter = (item) => item.name,
+  searchable = false, // New prop to enable search for multiple
```
tananaev, line 20: Is there a reason we need this configurable?

src/common/components/SelectField.jsx
```diff
@@ -40,20 +41,58 @@ const SelectField = ({
     return (
       <FormControl fullWidth={fullWidth}>
         {multiple ? (
-          <>
-            <InputLabel>{label}</InputLabel>
-            <Select
-              label={label}
+          searchable ? (
+            // Use Autocomplete for multiple with search
+            <Autocomplete
```
tananaev, line 46: I think it makes sense to use the same `Autocomplete` for everything now.

src/common/components/SelectField.jsx
```diff
@@ -40,20 +41,58 @@ const SelectField = ({
...
+              getOptionLabel={(option) => titleGetter(option)}
+              renderOption={(props, option) => (
+                <MenuItem {...props} key={keyGetter(option)} value={keyGetter(option)}>
+                  {titleGetter(option)}
+                </MenuItem>
+              )}
+              isOptionEqualToValue={(option, value) => keyGetter(option) === keyGetter(value)}
+              value={items.filter(item => value.includes(keyGetter(item)))}
+              onChange={(_, newValue) => {
+                const selectedIds = newValue.map(item => keyGetter(item));
```
tananaev, line 59: inline this

src/common/components/SelectField.jsx
```diff
@@ -40,20 +41,58 @@ const SelectField = ({
...
+                onChange({ target: { value: selectedIds } });
+              }}
+              renderInput={(params) => (
+                <TextField {...params} label={label} placeholder={`Search... (${value.length} selected)`} />
```
tananaev, line 63: It's a bit odd to have both "search..." and "x selected". Maybe we show search when empty only. And selected if something selected. Also don't forget about localization. For selected maybe we can just show the comma-separated list.

src/common/components/SelectField.jsx
```diff
@@ -40,20 +41,58 @@ const SelectField = ({
...
+              )}
+              slotProps={{
+                chip: {
+                  style: { display: 'none' }
```
tananaev, line 67: Is there a better way to hide chips?

tananaev: Do we actually still need it?

src/common/components/SelectField.jsx
```diff
@@ -40,20 +41,58 @@ const SelectField = ({
...
+                }
+              }}
+              filterOptions={(options, { inputValue }) =>
+                options.filter(option =>
+                  titleGetter(option).toLowerCase().includes(inputValue.toLowerCase())
+                )
+              }
+              noOptionsText="No results found"
+              disableCloseOnSelect
```
tananaev, line 76: What are these for?

src/common/components/SelectField.jsx
harshanasrimal: then every field will be searchable

tananaev: And the problem with that is?

src/common/components/SelectField.jsx
harshanasrimal: - `noOptionsText` is the message display when no result found. It should localize. my bad. 
- `disableCloseOnSelect` is use to keep the dropdown open even user select an item. If not, the user will have to open the dropdown one by one

harshanasrimal: my bad. we can drop it

src/common/components/SelectField.jsx
tananaev: How is it harming anything?

src/common/components/SelectField.jsx
```diff
@@ -39,34 +50,32 @@ const SelectField = ({
...
+              if (multiple) {
+                const selectedIds = value.map(item => keyGetter(item));
+                onChange({ target: { value: selectedIds } });
+              } else {
+                onChange({ target: { value: value ? keyGetter(value) : emptyValue } });
+              }
+            }}
+            renderInput={(params) => <TextField {...params} label={label} placeholder={getPlaceholder()}/>}
+            disableCloseOnSelect={multiple}
+            slotProps={{
```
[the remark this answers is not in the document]
tananaev: Actually can we render selected items using `renderValue`? Instead of using placeholder for that?

src/common/components/SelectField.jsx
```diff
@@ -20,13 +21,23 @@ const SelectField = ({
...
       option = items.find((obj) => keyGetter(obj) === option);
     }
     return option ? titleGetter(option) : emptyTitle;
   };
 
+  const getPlaceholder = () => {
+    if (!multiple || value.length === 0) {
+      return t('sharedSearch');
+    }
+    const selectedTitles = items.filter(item => value.includes(keyGetter(item))).map(item => titleGetter(item)).join(', ');
```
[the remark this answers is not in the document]
harshanasrimal: solved

src/common/components/SelectField.jsx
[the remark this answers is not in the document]
tananaev: Why does it cause issues exactly?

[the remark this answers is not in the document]
harshanasrimal: Screenshot 2025-08-26 at 22 07 06

Its the normal behaviour and you see the value is rendered within the input element (as value parameter of the input)

And when we use `renderValue` it render the value separately instead rendering it as value of the input element.

Screenshot 2025-08-26 at 22 14 18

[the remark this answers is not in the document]
harshanasrimal: will have to implement the element manually. I will try when I get a free time

tananaev:
I believe we already have this implemented on master.

PR #1491, opened 2025-08-28
update mapbox-gl to 3.14.0 and dependency in package-lock.json

SmartGpsBD:
use latest dependency


tananaev:
It seems like you're very misinformed on what we use this dependency for and why we keep it at this version.

I'm going to close this PR. You should do a bit more research.

PR #1497, opened 2025-08-31
Closing status card on map.

RafailStratiotis:
Pre Select Device on map for Reports


RafailStratiotis:
yes i'm sorry this is for closing status card on map..
i will send the code for "Pre Select Device on map for Reports"

PR #1498, opened 2025-08-31
AuditPage - Display Names from Users

RafailStratiotis:
AuditPage - Display Names from Users

### DIFF
--- a/src/reports/AuditPage.jsx
+++ b/src/reports/AuditPage.jsx
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useState, useEffect } from 'react';
 import {
   Table, TableRow, TableCell, TableHead, TableBody,
 } from '@mui/material';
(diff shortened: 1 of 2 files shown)


src/reports/AuditPage.jsx
```diff
@@ -28,10 +28,21 @@ const AuditPage = () => {
   const { classes } = useReportStyles();
   const t = useTranslation();
 
-  const [columns, setColumns] = usePersistedState('auditColumns', ['actionTime', 'userId', 'actionType', 'objectType']);
+  const [columns, setColumns] = usePersistedState('auditColumns', ['actionTime', 'userId', 'actionType', 'objectType', 'objectId', 'address']);
```
tananaev, line 31: Let's not change this

src/reports/AuditPage.jsx
```diff
@@ -28,10 +28,21 @@ const AuditPage = () => {
...
   const [items, setItems] = useState([]);
   const [loading, setLoading] = useState(false);
 
+  const [users, setUsers] = useState({});
+  useEffect(() => {
+    fetchOrThrow('/api/users')
+      .then(res => res.json())
+      .then(data => {
```
tananaev, line 39: This needs to be done with await

src/reports/AuditPage.jsx
```diff
@@ -61,7 +72,11 @@ const AuditPage = () => {
             <TableRow key={item.id}>
               {columns.map((key) => (
                 <TableCell key={key}>
-                  {key === 'actionTime' ? formatTime(item[key], 'minutes') : item[key]}
+                  {key === 'actionTime'
+                    ? formatTime(item[key], 'minutes')
+                    : key === 'userId'
+                      ? users[item[key]] || item[key]
+                      : item[key]}
```
tananaev, line 79: This should be extracted into a function, similar to what we do for other report formatting

PR #1502, opened 2025-09-02
Device license expiry in popup

hypoclone:
This PR added Device license expiry option in popup for status card, also the font color will be red if the license has expired.
[Expiry date.webm](https://github.com/user-attachments/assets/98459357-4ccb-47c7-bf6d-4aaf6f61257c)


tananaev:
This change doesn't really make sense as is. If we want to add device attributes to the popup, it should be done in a generic way, so you can add any attribute. We also should use existing names, not adding new strings just for the popup.

tananaev:
I mean if we add device attributes and properties, we should add all of them, not just one.

tananaev:
I'm going to close this for now.

PR #1574, opened 2025-10-25
Override Geofence name as address if device is within a geofence

jcardus:


jcardus:
@tananaev can you check?

jcardus:
I don't see where

tananaev:
Actually never mind. But I think it doesn't make sense calling `GeofencesValue` from `AddressValue`. Maybe we can have some higher level component that will call one or the other. It also definitely needs to be configurable.

PR #1575, opened 2025-10-27
Fix visibility of the close button in status card

lovery:
The close button is not visible when there is an image for the device and the theme of the phone is dark, but it is there and you can click on it if you know where exactly it is:
IMG_20251027_140623

I've added a line with the device name and the close button like it is when there is no image.

This behavior is happening on Pixel 7 and Pixel 8 with Chrome and Firefox.

### DIFF
--- a/src/common/components/StatusCard.jsx
+++ b/src/common/components/StatusCard.jsx
@@ -49,7 +49,8 @@ const useStyles = makeStyles()((theme, { desktopPadding }) => ({
     alignItems: 'flex-start',
   },
   mediaButton: {
-    color: theme.palette.primary.contrastText,
+    color: theme.palette.text.primary,
+    backgroundColor: theme.palette.background.default + '88',
     mixBlendMode: 'difference',
   },
   header: {


tananaev:
We have this style on the icon:
```
  mediaButton: {
    color: theme.palette.primary.contrastText,
    mixBlendMode: 'difference',
  },
```
Which should make it visible, but I guess it doesn't work for some reason. So that part needs fixing.

lovery:
Here is a few screenshots of how the fix looks like:
Screenshot from 2025-10-28 13-28-28Screenshot from 2025-10-28 13-27-01

tananaev:
I already pointed it out earlier. `mix-blend-mode: difference;` should inverse the color, so it should be always visible, no matter what the background is.

tananaev:
Why is it unreasonable?

tananaev:
This is already fixed on master with `mix-blend-mode`, so your claim was completely false.

Screenshot 2026-06-22 at 7 27 13 PM

PR #1580, opened 2025-10-28
Nb-translation

asbjornstokka:
Suggested improvements to Norwegian bokmål


jinzo:
I'm not sure if there is a two way sync, but the usual way to translate the web interface is thorough Transifex at [https://app.transifex.com/traccar/traccar/dashboard/](https://app.transifex.com/traccar/traccar/dashboard/) and that then gets synced to the code before release.

tananaev:
@jinzo is correct. You have to use Transifex for localization.

PR #1751, opened 2026-04-14
Added translations for more position attributes

nexx512:
I added a few new translations for attributes that had none for my teltonika device.


tananaev:
That's not how it works. We only add translations for common attributes declared in the Position class on the server side. Those are usually the ones that are commonly reused between different protocols.

PR #1753, opened 2026-04-15
Use description in StatusCard and PositionPage for computed attributes

nexx512:
As the description of computed attributes is used when linking them to devices, it might also be good to use these texts in the positions page and status card.

### DIFF
--- a/src/common/attributes/useComputedAttributes.js
+++ b/src/common/attributes/useComputedAttributes.js
@@ -1,0 +1,23 @@
+import { useEffect, useState } from 'react';
+import fetchOrThrow from '../util/fetchOrThrow';
+
+export default () => {
+  const [descriptions, setDescriptions] = useState({});
+
+  useEffect(() => {
+    fetchOrThrow('/api/attributes/computed')
+      .then((response) => response.json())
+      .then((items) => {
+        const map = {};
+        items.forEach((item) => {
+          if (item.description && item.attribute) {
+            map[item.attribute] = item.description;
+          }
+        });
+        setDescriptions(map);
+      })
+      .catch(() => {});
+  }, []);
+
+  return descriptions;
+};
--- a/src/common/attributes/useComputedAttributes.js
+++ b/src/common/attributes/useComputedAttributes.js
@@ -1,23 +1,16 @@
-import { useEffect, useState } from 'react';
-import fetchOrThrow from '../util/fetchOrThrow';
+import { useMemo } from 'react';
+import { useSelector } from 'react-redux';
 
 export default () => {
-  const [descriptions, setDescriptions] = useState({});
+  const items = useSelector((state) => state.computedAttributes.items);
 
-  useEffect(() => {
-    fetchOrThrow('/api/attributes/computed')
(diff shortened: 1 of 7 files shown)


src/common/attributes/useComputedAttributes.js
```diff
@@ -0,0 +1,23 @@
+import { useEffect, useState } from 'react';
+import fetchOrThrow from '../util/fetchOrThrow';
+
+export default () => {
+  const [descriptions, setDescriptions] = useState({});
+
+  useEffect(() => {
+    fetchOrThrow('/api/attributes/computed')
```
tananaev: It doesn't make sense to load it every time you open the card.

src/common/components/StatusCard.jsx
```diff
@@ -216,7 +218,7 @@ const StatusCard = ({ deviceId, position, onClose, disableActions, desktopPaddin
                         .map((key) => (
                           <StatusRow
                             key={key}
-                            name={positionAttributes[key]?.name || key}
+                            name={computedAttributes[key] || positionAttributes[key]?.name || key}
```
tananaev, line 221: This also doesn't verify the linking between device and the attribute.

src/common/attributes/useComputedAttributes.js
```diff
@@ -1,23 +1,16 @@
...
-      })
-      .catch(() => {});
-  }, []);
-
-  return descriptions;
+  return useMemo(() => {
+    const map = {};
+    Object.values(items).forEach((item) => {
+      if (item.description && item.attribute) {
+        map[item.attribute] = item.description;
```
tananaev, line 11: Still not sure this is the right thing to do in general. A description is not the same thing as a label. I reviewed some of the descriptions we have on the common server and most of them are too long.

nexx512: I thought about it too, but as they are used in the dialog for linking it to devices, I thought it might be more consistent that way. But maybe I'm wrong.

nexx512: Maybe it should be the other way round then. Using the label when linking computed attributed to devices?

tananaev: Label might be useful when linking, but I don't think it's the same. Linking just shows one string per line, which is not that bad if it's long. What you're doing is adding it to a table, which is definitely not designed for a long string.

nexx512: Yes, that's correct. I was just thinking that it's weird if on one page you have a description as identifier and on another you see the label. But apparently it doesn't seem to be a problem for most people. Probably just leave the way it is.

tananaev: When linking description is important because you could have completely different expressions to calculate the same attribute for different devices. The output should be the attribute label because that's what the expression ultimately spits out.

nexx512: Thank you for that insight. I totally missed that point. So this PR was a turn in the wrong direction. I'll close it.

PR #1796, opened 2026-05-14
Fix ruler map style loading

AussieTraccar:
**Purpose:** Fix map style loading error when toggling map layers, introduced in 15331f2


tananaev:
Fixed in a slightly different way in https://github.com/traccar/traccar-web/commit/fb4df5213c32a9e29f86ed4732e263dfdae49aed.

PR #1801, opened 2026-05-14
Improve ruler button toggle

AussieTraccar:
**Purpose:** The map ruler button is difficult to identify when it is active or inactive, as the colour difference between the two states is very similar. 

**Solution:** This PR implements a distinct colour change between inactive state; black on white (default icon colour) and active state; white on black (reversed icon colour). This PR also supports eliminating touch screen hover stickiness for mobile devices.

### DIFF
--- a/src/map/control/MapRuler.jsx
+++ b/src/map/control/MapRuler.jsx
@@ -174,7 +174,7 @@ const MapRuler = ({ positions, onActiveChange }) => {
         map.removeSource('ruler');
       }
     };
-  }, [theme, t, distanceUnit, classes.button]);
+  }, [theme, t, distanceUnit, classes.button, classes.active]);
 
   return null;
 };
--- a/src/map/control/MapRuler.jsx
+++ b/src/map/control/MapRuler.jsx
@@ -12,14 +12,21 @@ import { useTranslation } from '../../common/components/LocalizationProvider';
 
 const useStyles = makeStyles()(() => ({
   button: {
-    '&&': {
-      display: 'flex',
-      alignItems: 'center',
-      justifyContent: 'center',
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'center',
+    backgroundColor: '#e6e6e6',
+    color: '#333',
+    WebkitTapHighlightColor: 'transparent',
+  },
+  active: {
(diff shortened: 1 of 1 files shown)


tananaev:
Please provide screenshots before and after.

AussieTraccar:
eventListener_mouse_exit

Thinking about it, the eventListener is not the best way to go about it, as ruler button is slightly dimmed with mouse exit.

Will revisit without eventListeners.

AussieTraccar:
Screenshots of before and after.

**Ruler button (Before):** Idle -> mouse enter -> active

ruler_toggle-before

# 

**Ruler button (After):** Idle -> mouse enter -> active

ruler_toggle-after

tananaev:
I don't know if this design makes sense. Current one is the standard MapLibre uses for all controls. For example drawing controls.

AussieTraccar:
I would have thought it falls under the Maplibre toggle interactions, which was why I initially did it with eventListeners, but it didn't quite maintain your established theme.

From a user experience perspective, it provides visual clarity with the action. The other map controls also provide some visual clarity with pop-out fields dialogs etc. Seemingly trivial interactions like this, can contribute to the overall user experience.

tananaev:
Not sure I understand the point you're trying to make. There's already a selected state.

AussieTraccar:
Maybe I am missing something, but what is the selected state difference?

All of the map controls, with the exception of the ruler button, provide feedback to the user when the button has been selected (e.g. switcher -> map selection pop-up and geocoder -> search pop-up). The ruler button provides no feedback to the user when it is selected. This PR provides that feedback, for when the button has been selected.

I hope that makes a bit more sense?

tananaev:
| Deselected | Selected |
|-|-|
| Screenshot 2026-05-16 at 6 50 34 AM | Screenshot 2026-05-16 at 6 50 41 AM |

AussieTraccar:
Strange, as I only see the dimmed icon on mouse enter (desktop browser) and not dimmed with active like my screenshots above. On the mobile app (Android) it is almost indistinguishable between active and inactive. I am just proposing to make it more pronounced.

tananaev:
I think you need to investigate why standard selected state is not working for you.

PR #1813, opened 2026-05-26
Prevent real-time position updates from overwriting accumulator edits

timbze:
Previously, editing the hours or total distance accumulators had two main issues:
1. Incoming real-time position updates would overwrite the user's input mid-edit.
2. Typing a specific distance (e.g., 42 miles) would immediately change to a floating-point approximation (e.g., 42.00000000000001) due to continuous two-way unit conversion between the input field and the underlying state.

This commit fixes only issue 1 (after discussing things) by:
- No longer updating inputs when device position is updated

### DIFF
--- a/src/settings/AccumulatorsPage.jsx
+++ b/src/settings/AccumulatorsPage.jsx
@@ -33,14 +33,14 @@ const AccumulatorsPage = () => {
   const [item, setItem] = useState();
 
   useEffect(() => {
-    if (position) {
+    if (position && !item) {
       setItem({
         deviceId: parseInt(deviceId, 10),
         hours: position.attributes.hours || 0,
         totalDistance: position.attributes.totalDistance || 0,
       });
     }
-  }, [deviceId, position]);
+  }, [deviceId, position, item]);
 
   const handleSave = useCatch(async () => {
     await fetchOrThrow(`/api/devices/${deviceId}/accumulators`, {
--- a/src/settings/AccumulatorsPage.jsx
+++ b/src/settings/AccumulatorsPage.jsx
@@ -33,20 +33,25 @@ const AccumulatorsPage = () => {
   const [item, setItem] = useState();
 
   useEffect(() => {
-    if (position) {
+    if (position && !item) {
       setItem({
         deviceId: parseInt(deviceId, 10),
-        hours: position.attributes.hours || 0,
-        totalDistance: position.attributes.totalDistance || 0,
+        hours: (position.attributes.hours || 0) / 3_600_000,
+        totalDistance: distanceFromMeters(position.attributes.totalDistance || 0, distanceUnit),
       });
     }
-  }, [deviceId, position]);
+  }, [deviceId, position, item, distanceUnit]);
 
   const handleSave = useCatch(async () => {
+    const updateItem = {
+      ...item,
+      hours: item.hours * 3_600_000,
+      totalDistance: distanceToMeters(item.totalDistance, distanceUnit),
+    };
     await fetchOrThrow(`/api/devices/${deviceId}/accumulators`, {
       method: 'PUT',
       headers: { 'Content-Type': 'application/json' },
-      body: JSON.stringify(item),
+      body: JSON.stringify(updateItem),
     });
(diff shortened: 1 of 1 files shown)


tananaev:
This looks overcomplicated. Maybe we can just pin position?

timbze:
> This looks overcomplicated. Maybe we can just pin position?

You mean that there are no live updates right? That could work too. I wasn't sure if that was part of the intended features. I can simplify.

tananaev:
Yep, so it doesn't update after you open the screen.

timbze:
Simplified now. Numbers can have max 6 decimals with how I did it now. Let me know if you think the floating point convert issue should be solved differently.

src/settings/AccumulatorsPage.jsx
```diff
@@ -30,17 +30,11 @@ const AccumulatorsPage = () => {
...
-        deviceId: parseInt(deviceId, 10),
-        hours: position.attributes.hours || 0,
-        totalDistance: position.attributes.totalDistance || 0,
-      });
-    }
-  }, [deviceId, position]);
+  const [item, setItem] = useState(() => ({
+    deviceId: parseInt(deviceId, 10),
+    hours: position?.attributes.hours || 0,
+    totalDistance: position?.attributes.totalDistance || 0,
```
tananaev: We probably don't even need to fallback to zeroes here.

src/settings/AccumulatorsPage.jsx
```diff
@@ -62,15 +56,15 @@ const AccumulatorsPage = () => {
             <AccordionDetails className={classes.details}>
               <TextField
                 type="number"
-                value={item.hours / 3600000}
+                value={Number((item.hours / 3600000).toFixed(6))}
```
tananaev: I'm not a big fan of hardcoding formatting here.

timbze:
I've changed code again. Let me know what you think. I think `|| 0` is required, I manually created a device and added a position with emulator, but it did not have hours from it and then got js error because hours had no value. It might be just emulator thing, not sure. But I thought probably it's good to leave the safety as it was before.

timbze:
I had to put back `useEffect` because when accumulator page is reloaded, then on load there is no position and it just shows 0.

tananaev:
Can you explain why you had to change everything? It again looks too complicated and messy.

timbze:
The diff looks messier, but the code looks cleaner IMO.

Now I only do the convert once when setItem is called, and I only convert again if a change is saved. Before, the convert happened every time text got entered.

I had to keep the `useEffect` because when page is reloaded while accumulators page is open then position does not have a value initially. So I added `&& !item` to only set item once when position gets a value.

tananaev:
I don't think this is the right approach. This goes against every other place in the app.

tananaev:
What was wrong with your second option? I think it looked cleanest.

timbze:
I assume you mean because `item` values are now in different unit, that's understandable. I changed it now to only fix the issue of updates overwriting the input. Is this the correct fix?

The float rounding issues will happen in more places when user's setting is miles, would the proper fix be to round in `converter.js` to something like 10 digits in the `distanceFromMeters` function?

timbze:
> What was wrong with your second option? I think it looked cleanest.

I had removed useEffect which I realized causes a bug when page is refreshed. Also I am not sure how you want formatting since you didn't like how I did it.

timbze:
I had used the formatting to address the bug with float values. But possibly that should rather be handled in `converter.js`

tananaev:
I say we separate two things. Formatting and fixing updates are different unrelated issues. What was the problem with page refreshes?

timbze:
When I remove `useEffect` and I refresh the page `settings/accumulators/1`, the input values only show 0. It seems like position does not immediately have a value on page load.

When opening the accumulators page from the total distance -> settings icon, then it works.

I think this may have to do with socket connection to the server, which is not ready quickly enough on a page refresh.

tananaev:
OK, let's merge what you have currently.

tananaev:
Thanks!

PR #1816, opened 2026-05-27
Add PMTiles protocol support for custom maps

yuiseki:
## Summary

Registers the `pmtiles://` protocol with MapLibre so a custom map URL can point to a style.json whose vector sources are backed by a [PMTiles](https://docs.protomaps.com/pmtiles/) archive.

This makes it possible to use serverless, single-file vector basemaps hosted on commodity storage (S3, R2, etc.) without running a tile server.

This is also useful for offline or self-hosted deployments where users want to keep a single-file vector basemap next to Traccar, without running a separate tile server.

The existing custom map handling already passes a non-template style URL straight to `map.setStyle`, so registering the protocol is all that is needed.

## Impact

This change is purely additive and introduces no breaking changes. Every existing map keeps working exactly as before, whether it is a raster tile source, a vector style URL, or one of the provider maps (Google, Mapbox, MapTiler, and so on). The only new behavior is that a style referencing `pmtiles://` URLs now resolves instead of failing with an unknown protocol error, so this simply unlocks PMTiles support for anyone who wants it without affecting anyone who does not.

## Changes

- Add the `pmtiles` dependency.
- Register the `pmtiles` protocol next to the existing `google` protocol in `MapView.jsx`.

## Testing

Verified against a local traccar server by setting the custom map URL to a style.json whose sources use `pmtiles://.../planet.pmtiles` (OpenMapTiles schema). The map renders correctly and fetches tiles via HTTP Range requests (206 Partial Content) with no errors. Also confirmed with the Protomaps basemap schema.

### DIFF
--- a/src/map/core/MapView.jsx
+++ b/src/map/core/MapView.jsx
@@ -18,8 +18,7 @@ element.style.boxSizing = 'initial';
 
 maplibregl.addProtocol('google', googleProtocol);
 
-const pmtilesProtocol = new PMTilesProtocol();
-maplibregl.addProtocol('pmtiles', pmtilesProtocol.tile);
+maplibregl.addProtocol('pmtiles', new PMTilesProtocol().tile);
 
 export const map = new maplibregl.Map({
   container: element,
--- a/src/map/core/MapView.jsx
+++ b/src/map/core/MapView.jsx
@@ -1,6 +1,7 @@
 import 'maplibre-gl/dist/maplibre-gl.css';
 import maplibregl from 'maplibre-gl';
 import { googleProtocol } from 'maplibre-google-maps';
+import { Protocol as PMTilesProtocol } from 'pmtiles';
 import { useRef, useLayoutEffect, useEffect, useState, useMemo } from 'react';
 import { useTheme } from '@mui/material';
 import MapSwitcher from '../control/MapSwitcher';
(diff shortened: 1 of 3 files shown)


tananaev:
How big is the library?

yuiseki:
@tananaev Thank you for comment!

The library is tiny in the bundle. `pmtiles@4.4.1` only pulls in `fflate` (a small inflate lib) as a dependency. In the production build it tree-shakes down to about **+18.6 kB raw / +7.0 kB gzipped**, added to the lazy-loaded `MapView` chunk, so there is no impact on the initial load.

(For reference, the on-disk install footprint is ~1.2 MB including source maps and both CJS/ESM builds, but none of that ships to users.)

src/map/core/MapView.jsx
```diff
@@ -17,6 +18,9 @@ element.style.boxSizing = 'initial';
 
 maplibregl.addProtocol('google', googleProtocol);
 
+const pmtilesProtocol = new PMTilesProtocol();
+maplibregl.addProtocol('pmtiles', pmtilesProtocol.tile);
```
tananaev:
```suggestion
maplibregl.addProtocol('pmtiles', new PMTilesProtocol().tile);
```

tananaev:
Can you also provide some map examples for testing.

yuiseki:
Sure! Here are two example styles you can paste into the custom map URL to test PMTiles. Both are OpenMapTiles-schema styles backed by a single `planet.pmtiles` file, served with CORS and HTTP Range support, so they work out of the box:

- https://yuiseki.dev/static/styles/osm-bright.json
- https://yuiseki.dev/static/styles/osm-fiord.json

tananaev:
Merged, thanks.

PR #1819, opened 2026-05-28
Fix input reformatting due to float value tail noise

### LINKED ISSUE #1817: Accumulators `total distance` miles input reformats while entering number due to conversion and float non-precision
Enter 42 and it immediately changes to 42.00000000000001
Image

One way this can be solved, but I don't know if this is the right way.
```js
// in converter.js change this
export const distanceFromMeters = (value, unit) => value * distanceConverter(unit);

// to
export const distanceFromMeters = (value, unit) => Number((value * distanceConverter(unit)).toFixed(13));
```

timbze:
Fixes #1817

I think this same thing can be a problem for other conversions from liters and knots so I have addressed those too.

### DIFF
--- a/src/common/util/converter.js
+++ b/src/common/util/converter.js
@@ -1,3 +1,5 @@
+const PRECISION_FIX_DECIMAL = 13;
+
 const speedConverter = (unit) => {
   switch (unit) {
     case 'kmh':
@@ -22,7 +24,8 @@ export const speedUnitString = (unit, t) => {
   }
 };
 
-export const speedFromKnots = (value, unit) => value * speedConverter(unit);
+export const speedFromKnots = (value, unit) =>
+  Number((value * speedConverter(unit)).toFixed(PRECISION_FIX_DECIMAL));
 
 export const speedToKnots = (value, unit) => value / speedConverter(unit);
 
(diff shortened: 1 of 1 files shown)


src/common/util/converter.js
```diff
@@ -22,7 +24,8 @@ export const speedUnitString = (unit, t) => {
   }
 };
 
-export const speedFromKnots = (value, unit) => value * speedConverter(unit);
+export const speedFromKnots = (value, unit) =>
+  Number((value * speedConverter(unit)).toFixed(PRECISION_FIX_DECIMAL));
```
tananaev, line 28: Why do we need to wrap in a number?

src/common/util/converter.js
```diff
@@ -1,3 +1,5 @@
+const PRECISION_FIX_DECIMAL = 13;
```
tananaev, line 1: Why 13?

src/common/util/converter.js
tananaev: Also `toFixed` returns a string. I don't think that's what we expect for these functions.

src/common/util/converter.js
timbze: 14 didn't fix the issue, 13 did. Searching online I found 14 and 15 being used, so I'm not sure how to know what is the right number.

From usability perspective it could probably be quite a bit less.

src/common/util/converter.js
timbze: This is what I find on how this problem is usually handled when I search online or ask AI. I guess rounding math could probably work too.

Wrapping in number is because `toFixed` returns a string and then Number turns it back into string before returning.

tananaev:
Fixed it in slightly another way in https://github.com/traccar/traccar-web/commit/d2ca6095218d27dd9e10ca17d2b7479e2b0a703a. Thanks.

timbze:
> Fixed it in slightly another way in [d2ca609](https://github.com/traccar/traccar-web/commit/d2ca6095218d27dd9e10ca17d2b7479e2b0a703a). Thanks.

Thank you

PR #1824, opened 2026-06-02
[FEAT]: Change expiration input type to datetime-local

BOKG:
The share API already supports an exact expiration timestamp, but the share page only exposes a date input. This change switches the input to datetime-local so short-lived share links can be created from the UI as well.

No API or schema changes are required.


tananaev:
Do you have a screenshot before and after?

BOKG:
Hi! @tananaev.  I hope you are very well.

Before:

After:
Captura desde 2026-06-02 18-53-43

PR #1833, opened 2026-06-08
Fix log subscription WebSocket state handling

Sxntoz:
This change improves WebSocket handling for the Logs page by sending the current log subscription state when the connection opens and preventing messages from being sent when the socket is not open.

### DIFF
--- a/src/SocketController.jsx
+++ b/src/SocketController.jsx
@@ -33,6 +33,9 @@ const SocketController = () => {
   const authenticated = useSelector((state) => Boolean(state.session.user));
   const includeLogs = useSelector((state) => state.session.includeLogs);
 
+  const includeLogsRef = useRef(includeLogs);
+  includeLogsRef.current = includeLogs;
+
   const socketRef = useRef();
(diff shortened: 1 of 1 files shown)


tananaev:
Can you please provide more details on what's broken and how this fixes it.

Sxntoz:
The issue happens when the Logs page toggles log subscription while the WebSocket is not fully open. In that case the client can try to send the logs subscription message while the socket is still connecting, closing, or already closed.

This change fixes it by checking that the socket is in the OPEN state before sending updates. It also sends the current logs subscription state in the onopen handler, so the server receives the correct state after the connection is established or reconnected.

src/SocketController.jsx
```diff
@@ -91,6 +94,7 @@ const SocketController = () => {
 
     socket.onopen = () => {
       dispatch(sessionActions.updateSocket(true));
+      socket.send(JSON.stringify({ logs: includeLogsRef.current }));
```
tananaev, line 97: I suggest we only send this if "includeLogs = true".

PR #1843, opened 2026-06-10
fix(map): gracefully handle missing devices in position features

syed-ghufran-hassan:
* Add safeguards when creating map feature properties to handle cases where a device is not available in the Redux store.
* Prevent runtime errors caused by temporary device/position synchronization issues.
* Skip invalid features (or use safe fallbacks) instead of accessing undefined device properties.
* Improve map stability during device removal, refreshes, and data-loading edge cases.

### DIFF
--- a/src/map/MapPositions.js
+++ b/src/map/MapPositions.js
@@ -36,33 +36,37 @@ const MapPositions = ({
   disabledRef.current = disabled;
 
   const createFeature = useCallback(
-    (devices, position, selectedPositionId) => {
-      const device = devices[position.deviceId];
-      let showDirection;
-      switch (directionType) {
-        case 'none':
-          showDirection = false;
-          break;
-        case 'all':
-          showDirection = position.course > 0;
-          break;
-        default:
-          showDirection = selectedPositionId === position.id && position.course > 0;
-          break;
-      }
-      return {
-        id: position.id,
-        deviceId: position.deviceId,
-        name: device.name,
-        fixTime: formatTime(position.fixTime, 'seconds'),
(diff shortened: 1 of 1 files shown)


tananaev:
What exactly is broken? What are the steps to reproduce?

syed-ghufran-hassan (commented):
### What was broken?

The map assumed that every position's `deviceId` had a corresponding device entry in the Redux store. During device refreshes, removals, or temporary synchronization delays, positions could be received before the associated device data was available.

In these cases, map feature creation attempted to access properties of an undefined device object, resulting in runtime errors and preventing position markers from rendering correctly.

### Steps to reproduce

1. Open the map page with active device positions.
2. Trigger a device list refresh, device removal, or any operation that temporarily clears/reloads the Redux device store.
3. Ensure position updates are received while the device store is incomplete or still loading.
4. Observe that position feature generation encounters missing device references.

### Actual behavior

The map may throw runtime errors when feature properties are created for positions whose devices are not yet present in the Redux store. This can lead to broken marker rendering or map instability.

### Expected behavior

The map should handle missing device references gracefully by using safe fallback values or skipping invalid features until device data becomes available. Position updates should not cause runtime errors during refresh, removal, or synchronization edge cases.

### Fix

Added safeguards during feature creation to handle undefined devices safely. Feature properties now use fallback values (or skip invalid entries) instead of accessing properties on an undefined device object, improving map stability during data synchronization transitions.

tananaev:
> Trigger a device list refresh, device removal, or any operation that temporarily clears/reloads the Redux device store.

This is a complete nonsense. Have you actually tried it?

PR #1870, opened 2026-06-28
fixed #5925: added dynamic periodinterval

harrowmykel:
This is a fix for https://github.com/traccar/traccar/issues/5925 <-- read this to see what it does.

It rewrites #1858 to use the same issue id


src/resources/l10n/en.json
```diff
@@ -591,6 +591,9 @@
     "reportStops": "Stops",
     "reportSummary": "Summary",
     "reportDaily": "Daily Summary",
+    "reportWeekly": "Weekly Summary",
+    "reportMonthly": "Monthly Summary",
+    "reportYearly": "Yearly Summary",
```
tananaev: I feel like we should probably generify all these, if we're added this complexity.

harrowmykel:
That makes sense.

tananaev:
Please reply properly to comments. Email doesn't work very well on GitHub.

src/resources/l10n/en.json
harrowmykel: That makes sense. Great idea

harrowmykel: added changes

src/resources/l10n/de.json
```diff
@@ -590,6 +590,10 @@
     "reportTrips": "Fahrten",
     "reportStops": "Stopps",
     "reportSummary": "Zusammenfassung",
+    "periodIntervalDaily": "Täglich",
+    "periodIntervalWeekly": "Wöchentlich",
+    "periodIntervalMonthly": "Monatlich",
+    "periodIntervalYearly": "Jährlich",
```
tananaev: You should only include English.

harrowmykel: removed it

PR #1892, opened 2026-07-10
Add Matrix notificator support

MaBurn:
# Add Matrix notificator support (UI)

## Background

We are a non-profit association (Verein) and long-time Traccar users. For data-protection reasons we don't want to route notification data through WhatsApp or Telegram, and we'd like to avoid the recurring cost of SMS. Our association already uses [Matrix](https://matrix.org) as its messenger and prefers it as an open, self-hostable and secure protocol, so we added a Matrix notificator to Traccar (with the help of Claude) and would like to contribute it back.

This pull request is the web UI companion to the server-side change in `traccar/traccar`: .

## What this adds

- A new `matrixUser` user attribute (`useUserAttributes.js`), so a user can store their Matrix address (e.g. `@jane:example.org`) for personal, per-user notifications.
- On the user page, the Matrix address field is shown only when the server actually has the `matrix` notificator enabled — it's gated on `/api/notifications/notificators`, so nothing appears for deployments that don't use Matrix.
- Localization keys `attributeMatrixUser` and `notificatorMatrix`, with translations added to the language files.

## Notes

The Matrix address field is optional. Users without it simply receive notifications in the server-configured default Matrix room; users who set it can receive personal notifications in a direct room. This mirrors how the server-side notificator resolves rooms.

Happy to adjust naming, gating, or translations to match project conventions.


tananaev:
You should not include translations here. Translations are done through Transifex.

MaBurn:
You're right — reverted the translation files. Kept only en.json.

PR #1911, opened 2026-07-23
l10n: fix German translation for mapDefault

brickeydev:


tananaev:
Please use Transifex for localization.

PR #1921, opened 2026-08-04
Show channels 1 and 2 simultaneously in live video

emmanuelh-dev:
## Summary
- Show both camera channels (1 and 2) at once in the live video view instead of a single channel with a manual selector
- Auto-start streams on load and auto-reconnect on stream errors
- Stack channels vertically on mobile screens

image
image

## Test plan
- [x] Open the live video view for a device with two configured channels and confirm both streams start automatically
- [x] Verify a dropped stream reconnects automatically
- [x] Verify layout on a mobile-width viewport stacks channels vertically

### DIFF
--- a/src/other/StreamPage.jsx
+++ b/src/other/StreamPage.jsx
@@ -1,21 +1,46 @@
 import { useEffect, useRef, useState } from 'react';
 import { useSelector } from 'react-redux';
-import { Typography, IconButton, Toolbar, Paper, TextField } from '@mui/material';
+import { Typography, IconButton, Toolbar, Paper } from '@mui/material';
 import { makeStyles } from 'tss-react/mui';
 import { useNavigate, useSearchParams } from 'react-router-dom';
-import { default as Hls, Events } from 'hls.js/light';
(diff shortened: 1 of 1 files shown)


src/other/StreamPage.jsx
```diff
@@ -1,56 +1,180 @@
...
+const RETRY_DELAY = 5000;
+
 const useStyles = makeStyles()((theme) => ({
   root: {
     height: '100%',
     display: 'flex',
     flexDirection: 'column',
   },
-  video: {
+  title: {
```
tananaev, line 25: Why there's so much CSS? I prefer absolute minimal.

emmanuelh-dev: Done

tananaev:
Can we actually split this PR into smaller ones. I see one clear one is a refactor to `ChannelPlayer`. And then maybe the rest as a separate PR.

emmanuelh-dev:
Sure. I will split it into two separate PRs: first the ChannelPlayer refactor, then the multi-channel controls.

PR #1922, opened 2026-08-05
Extract ChannelPlayer and fix HLS player concurrency issues

emmanuelh-dev:
## Summary
Split out of #1921 per @tananaev's request — this is just the `ChannelPlayer` refactor with no behavior change to the existing single-channel view. The multi-channel add/remove controls will follow in a separate PR built on top of this one.

- Extracts the HLS setup into its own `ChannelPlayer` component
- Fixes concurrency issues found in review:
  - clears the pending retry timeout before scheduling a new one, so repeated fatal errors can't stack overlapping retries
  - keeps the `<video>` element mounted while playing instead of unmounting it on error, so the `Hls` instance never outlives the DOM node
  - retries via `hls.startLoad()` / `hls.recoverMediaError()` instead of tearing down and recreating the whole `Hls` instance
  - guards the `MANIFEST_PARSED` and retry callbacks with a `destroyed` flag so they don't touch the video ref after teardown

## Test plan
- [ ] Start a single-channel stream and confirm it still behaves the same as before (manual channel index, start/stop toggle)
- [ ] Simulate a dropped stream and confirm it recovers automatically without stacking retries

### DIFF
--- a/src/other/StreamPage.jsx
+++ b/src/other/StreamPage.jsx
@@ -3,7 +3,7 @@ import { useSelector } from 'react-redux';
 import { Typography, IconButton, Toolbar, Paper, TextField } from '@mui/material';
 import { makeStyles } from 'tss-react/mui';
 import { useNavigate, useSearchParams } from 'react-router-dom';
-import { default as Hls, Events } from 'hls.js/light';
+import { default as Hls, Events, ErrorTypes } from 'hls.js/light';
 import PlayArrowIcon from '@mui/icons-material/PlayArrow';
 import StopIcon from '@mui/icons-material/Stop';
 import { useTranslation } from '../common/components/LocalizationProvider';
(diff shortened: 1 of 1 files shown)


src/other/StreamPage.jsx
```diff
@@ -3,14 +3,16 @@ import { useSelector } from 'react-redux';
...
-import { default as Hls, Events } from 'hls.js/light';
+import { default as Hls, Events, ErrorTypes } from 'hls.js/light';
 import PlayArrowIcon from '@mui/icons-material/PlayArrow';
 import StopIcon from '@mui/icons-material/Stop';
 import { useTranslation } from '../common/components/LocalizationProvider';
 import { useCatchCallback } from '../reactHelper';
 import BackIcon from '../common/components/BackIcon';
 import fetchOrThrow from '../common/util/fetchOrThrow';
 
+const RETRY_DELAY = 5000;
```
tananaev: Inline given used in a single place.

src/other/StreamPage.jsx
```diff
@@ -35,13 +37,63 @@ const useStyles = makeStyles()((theme) => ({
...
+const ChannelPlayer = ({ classes, deviceId, channel, playing, setError, sendCommand }) => {
+  const videoRef = useRef(null);
+
+  useEffect(() => {
+    if (!playing) {
+      return undefined;
+    }
+
+    let destroyed = false;
+    let retryTimeout;
```
tananaev: Why do we need both?

src/other/StreamPage.jsx
```diff
@@ -35,13 +37,63 @@ const useStyles = makeStyles()((theme) => ({
   },
 }));
 
+const ChannelPlayer = ({ classes, deviceId, channel, playing, setError, sendCommand }) => {
+  const videoRef = useRef(null);
+
+  useEffect(() => {
+    if (!playing) {
+      return undefined;
+    }
```
tananaev: What is this guarding against?

tananaev (commented):
Can you also attach a video of this working. And video of error recovery.

emmanuelh-dev:
Now i dont have the error of "Conexion error" but, here is working, the camera is in my car and was installed yesterday and i notice this, so, i want to contribute with this part 

https://github.com/user-attachments/assets/aaf52c93-661d-4c2b-b7b5-c4b23b5e145c


When I open the app from my phone and tap "watch live video," the stream I was already viewing drops. I'm not sure if it's because of how Traccar handles the realtime connection it might be requesting a new session from the camera. I haven't checked whether it reuses the session that's already being requested or creates a new one, but that's the behavior I've noticed.

Captura de pantalla 2026-08-04 a la(s) 8 38 03 p m

src/other/StreamPage.jsx
emmanuelh-dev: inlined it in the latest push — it was only used once.

src/other/StreamPage.jsx
emmanuelh-dev: right, destroyed was redundant — clearTimeout on cleanup already prevents the retry callback from firing after unmount, and hls.destroy() unbinds its listeners so MANIFEST_PARSED/ERROR can't fire either. Removed it in the latest push.

src/other/StreamPage.jsx
emmanuelh-dev: This was leftover from an earlier version — ChannelPlayer is only ever mounted by the parent when playing is true, so the prop can never be false here. Removed the prop and the guard in the latest push.

src/other/StreamPage.jsx
```diff
@@ -35,13 +35,50 @@ const useStyles = makeStyles()((theme) => ({
...
+        return;
+      }
+      setError(true);
+      clearTimeout(retryTimeout);
+      retryTimeout = setTimeout(() => {
+        setError(false);
+        if (data.type === ErrorTypes.MEDIA_ERROR) {
+          hls.recoverMediaError();
+        } else {
+          hls.startLoad();
```
tananaev, line 62: Why do we want to unconditionally retry? I think it would be better to have a refresh button or something.

PR #1957, opened 2026-09-01
Add setSpeedLimit command attributes

onethings:
### Summary

Adds the `setSpeedLimit` command to the web UI so the speed limit and the overspeed duration can be configured from the device and saved-command dialogs.

- `useCommandAttributes.js`: define `setSpeedLimit` fields — speed limit (`data`) and overspeed duration in seconds (`duration`).

The `duration` value is only used by the JT808 encoder (JT/T 808-2019 parameter 0x56, written only when provided). Other protocols that support `setSpeedLimit` (e.g. Gator, Khd, HuaSheng) only read the speed value and ignore `duration`.

Reuses existing translation strings — `commandSetSpeedLimit`, `attributeSpeedLimit`, `reportDuration` — no new localization keys are needed.

### DIFF
--- a/src/common/attributes/useCommandAttributes.js
+++ b/src/common/attributes/useCommandAttributes.js
@@ -148,6 +148,11 @@ export default (t) =>
           name: t('attributeSpeedLimit'),
           type: 'number',
         },
+        {
+          key: 'duration',
+          name: t('reportDuration'),
+          type: 'number',
+        },
       ],
       modePowerSaving: [
         {
--- a/src/common/attributes/useCommandAttributes.js
+++ b/src/common/attributes/useCommandAttributes.js
@@ -148,12 +148,6 @@ export default (t) =>
           name: t('attributeSpeedLimit'),
           type: 'number',
         },
-        {
-          key: 'duration',
-          name: t('reportDuration'),
-          type: 'number',
-          protocol: 'jt808',
-        },
       ],
       modePowerSaving: [
         {
--- a/src/common/attributes/useCommandAttributes.js
+++ b/src/common/attributes/useCommandAttributes.js
@@ -142,6 +142,19 @@ export default (t) =>
           type: 'string',
         },
       ],
+      setSpeedLimit: [
+        {
+          key: 'data',
+          name: t('attributeSpeedLimit'),
+          type: 'number',
+        },
+        {
+          key: 'duration',
+          name: t('reportDuration'),
+          type: 'number',
+          protocol: 'jt808',
+        },
+      ],
       modePowerSaving: [
         {
           key: 'enable',
--- a/src/settings/components/BaseCommandView.jsx
+++ b/src/settings/components/BaseCommandView.jsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 import { useSelector } from 'react-redux';
 import { Autocomplete, Checkbox, FormControlLabel, TextField } from '@mui/material';
 import { useTranslation } from '../../common/components/LocalizationProvider';
@@ -20,6 +20,8 @@ const BaseCommandView = ({
   const limitCommands = useRestriction('limitCommands');
 
   const textEnabled = useSelector((state) => state.session.server.textEnabled);
+  const device = useSelector((state) => state.devices.items[deviceId]);
+  const protocol = device?.protocol;
 
   const availableAttributes = useCommandAttributes(t);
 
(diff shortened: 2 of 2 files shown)


tananaev:
Sorry, but I don't think we can just add parameters like this. It has to be something useful for other protocols.

onethings:
Thanks for the feedback! You make a valid point that it has to be useful for other protocols. 

I looked into other protocols like **Khd**, **HuaSheng**, and **Gator**. It turns out that both Khd and HuaSheng actually require a time duration for speeding alarms (while Gator is the only one that doesn't). 

Therefore, implementing the time duration parameter for **JT808, Khd, and HuaSheng** will be very useful and beneficial for consistency. Here is the protocol reference supporting this:

---

### 1. Khd (Protocol 5058)
[KHDtrack Standard Protocol V2.pdf](https://www.traccar.org/protocol/5058-khd/KHDtrack%20Standard%20Protocol%20V2.pdf)

#### 📌 Command Overview (`))<0xDB><\r>`)
| Item | Description |
| :--- | :--- |
| **Description** | Configure the over-speed and trigger-time limit values. |
| **Example Packet** | `29 29 DB 00 09 0A 9F 95 38 Byte1 Byte2 Byte3 Xor 0D` |
| **Device Reply** | Command `0x85`, data field `V8=0xDB` indicating ACK. |
| **Supported Device** | **T360-108** device only. |

#### ⚙️ Byte Parameter Details
| Field | Description & Unit | Example & Value Explanation |
| :--- | :--- | :--- |
| **Byte1** | Over-speed limit (km/h) | `0x64` (`0x64` = 100) means 100 km/h limit.<br>`0x00` means cancel over-speed alarm checking. |
| **Byte2 Byte3** | Over-speed trigger-time limit (sec) | `0x000A` (`0x000A` = 10) means the over-speed limit trigger-time is 10 sec. |

Screenshot 2026-09-02 at 09 17 57

---

### 2. HuaSheng (Protocol 5111)
[OBD Device Protocol_V1.1.4_160526.docx](https://www.traccar.org/protocol/5111-huasheng/OBD%20Device%20Protocol_V1.1.4_160526.docx)

#### 📋 Variable Parameter Section
| Type | Length | Value / Parameter | Explanation |
| :--- | :--- | :--- | :--- |
| `0X0001` | BYTE | Protection status | `0x00` = not protected<br>`0x01` = protected |
| `0X0002` | SHORT | Report interval | Unit: second |
| `0X0003` | SHORT | Sleep Wake up interval | Unit: Min |
| `0X0004` | BYTE | Over speed limit | Unit: km |
| `0X0005` | SHORT | Over speed delay interval | Unit: second |

HuaSheng

---

### 3. Gator (Protocol 5052)
[Gator vehicle tracker communication protocol.doc](https://www.traccar.org/protocol/5052-gator/Gator%20vehicle%20tracker%20communication%20protocol.doc) | [S208 S228 protocolENGLISH-st.doc](https://www.traccar.org/protocol/5052-gator/S208%20S228%20protocolENGLISH-st.doc)

*As a comparison, Gator does not need time duration at speeding setting.*

Gator

---
Let me know if this implementation looks good to you!

src/common/attributes/useCommandAttributes.js
```diff
@@ -142,6 +142,19 @@ export default (t) =>
...
+        {
+          key: 'data',
+          name: t('attributeSpeedLimit'),
+          type: 'number',
+        },
+        {
+          key: 'duration',
+          name: t('reportDuration'),
+          type: 'number',
+          protocol: 'jt808',
```
tananaev: I think we need to drop this. You keep expanding scope for every PR.

tananaev:
Good analysis. I think we can go ahead, assuming you will handle backend as well. I added a comment.

src/common/attributes/useCommandAttributes.js
onethings: You're right — I'll drop the `duration` field to keep the scope focused.

On the backend (traccar PR #5998) I'll update the `setSpeedLimit` encoder so it only
writes the overspeed-duration parameter (0x0056) when `duration` is actually provided —
otherwise the device keeps its current setting, so nothing breaks without the field.

I'll update both PRs accordingly.

tananaev: I think you misunderstood the comment. I am talking about protocol field specifically. Duration makes sense.

onethings: My apologies — I misread your comment. You're right, duration makes sense, so I've kept it and only removed the protocol: 'jt808' property from the entry (the UI doesn't use that property anyway). Fix is pushed to jt808-set-speed-limit.

PR #1961, opened 2026-09-06
fix: fix security issue in useMapStyles.js

anupamme:
## Summary
Fix critical severity security issue in `src/map/core/useMapStyles.js`.

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | V-001 |
| **Severity** | CRITICAL |
| **Scanner** | multi_agent_ai |
| **Rule** | `V-001` |
| **File** | `src/map/core/useMapStyles.js:13` |
| **Assessment** | Likely exploitable |

**Description**: Hardcoded fallback API keys for LocationIQ and Ordnance Survey map services are embedded in client-side JavaScript. These keys are visible to any user accessing the web application and can be extracted using browser developer tools.

## Evidence

**Exploitation scenario**: Open browser DevTools, navigate to Sources tab, locate useMapStyles.js, extract hardcoded keys.

**Scanner confirmation**: multi_agent_ai rule `V-001` flagged this pattern.

**Production code**: This file is in the production codebase, not test-only code.

## Threat Model Context

This is a web application - XSS and injection vulnerabilities can affect end users.

## Changes
- `src/map/core/useMapStyles.js`

## Behavior Preservation
The change is scoped to 1 file on the vulnerable path.

---
[redacted]

### DIFF
--- a/src/map/core/useMapStyles.js
+++ b/src/map/core/useMapStyles.js
@@ -35,14 +35,12 @@ export default () => {
 
   const googleKey = useAttributePreference('googleKey');
   const mapTilerKey = useAttributePreference('mapTilerKey');
-  const locationIqKey =
-    useAttributePreference('locationIqKey') || 'pk.0f147952a41c555a5b70614039fd148b';
+  const locationIqKey = useAttributePreference('locationIqKey');
   const bingMapsKey = useAttributePreference('bingMapsKey');
   const tomTomKey = useAttributePreference('tomTomKey');
   const hereKey = useAttributePreference('hereKey');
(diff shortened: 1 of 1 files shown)


tananaev:
Complete nonsense. Those are client keys, which are intended to be public.

anupamme:
The keys are indeed intended as client-side credentials, so I agree this isn’t a vulnerability merely because they’re present in the bundle. The scanner over-classified the finding based on the presence of API-key-like strings. I’ll close this as a false positive rather than treating the keys as secrets. A separate hardening improvement could move the defaults to deployment configuration, but that would be a configuration/credential-management change rather than a security vulnerability fix.

